PR #1, opened 2023-07-24
Added `maxBytes` that validates a string in byte length

genki:
I think this is handy for the most of databases that limit their column/field length in byte.


library/src/validations/maxBytes/maxBytes.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @param error The error message.
+ *
+ * @returns A validation function.
+ */
+export function maxBytes<TInput extends string>(
+  requirement: number,
+  error?: string
+) {
+  return (input: TInput, info: ValidateInfo) => {
+    const bytes = encodeURI(input).replace(/%../g, 'x').length;
```
fabian-hiller, line 17: Is it true that `encodeURI(input).replace(/%../g, 'x').length` works reliably to calculate the bytes? An alternative could be e.g. `new TextEncoder().encode(input).length`. Furthermore, to reduce the bundle size and since there is only one use, I would not use a variable and add the calculation directly to the if statement.

fabian-hiller:
Thanks for this PR, the idea is great. Feel free to add `minBytes` and `bytes` as well. Otherwise I will do that once we have this PR merged.

library/src/validations/maxBytes/maxBytes.ts
genki: This is an old idiom for calculating the byte length of string in the cross-browser way.
I agree with you `new TextEncoder().encode(input).length` is cleaner. But some of old browser doesn't recognize it.

fabian-hiller:
I would add `minBytes` and `bytes` anyway for completeness. Due to the modular design we don't hurt anyone with this. Also I tend to use `TextEncoder` for calculation. If someone wants to support Internet Explorer, he can implement `maxBytes` independently with `encodeURI`.

genki:
@fabian-hiller Yep. I will add `minBytes`, `bytes` and have used the `TextEncoder` to this PR :)

genki:
Forgot to export of `minBytes` and `bytes`. I have added them.

fabian-hiller:
Thank you for your commitment. 💙 I will check everything later and then merge it.

PR #12, opened 2023-07-26
fix: types

### LINKED ISSUE #7: Issues importing with Node16/NodeNext moduleResolution
"Are the types wrong?" tool reports that `Node16`/`NodeNext` are having issues importing the package
https://arethetypeswrong.github.io/?p=valibot%400.2.0

`Bundler` is reported as using fallback condition, this is not ideal and should be fixed as well.

image

userquin:
This PR includes:
- add package manager to root package.json
- add `unbuild` dev dependency in library package.json
- build valibot using unbuild instead Vite: Vite still there, if you want still use it (you can remove the dependency)
- exports all submodules
- added `dts` file per submodule for node10: included in `dist` package entry
- tsc will be only used to copy source files to dist folder once built
- udpated playground script to generate also stubs

If you want to only export the index module:
- remove all library `d.ts` files from library folder, remove also `d.cts` ones, don't remove `index.d.cts`
- remove all exports from library package.json (keep only `"."`)
- remove `"resolveJsonModule": true`  from library tsconfig file: only used to read the import entry from the package exports
- remove all code from `build.config.ts` file, keep only the import and the default export using `defineBuildConfig` (remove also the type from the import)
- replace `entries` with `entries: ['src/index'],` in `build.config.ts` file
- remove `*.d.ts` from library package.json `dist` option, don't remove `*.d.cts`

You can build and pack valibot: `pnpm build && pnpm pack` and upload generated tgz file to https://arethetypeswrong.github.io/

closes #7

### DIFF
--- a/library/build.config.ts
+++ b/library/build.config.ts
@@ -10,7 +10,7 @@ Object.keys(exportsList).forEach((key) => {
     return;
   }
 
-  const importValue = exportsList[key]['import'];
+  const importValue = exportsList[key]['import']['default'];
   if (key === '.') {
     entries.push({
       input: 'src/index',
@@ -38,5 +38,6 @@ export default defineBuildConfig({
       respectExternal: true,
     },
   },
+  failOnWarn: false,
   externals: ['vitest'],
 });
--- a/library/build.config.ts
+++ b/library/build.config.ts
@@ -1,0 +1,42 @@
+import { type BuildEntry, defineBuildConfig } from 'unbuild';
+
+import packageJSON from './package.json';
+
+const entries: BuildEntry[] = [];
+const exportsList = packageJSON['exports'];
+const match = '.';
+Object.keys(exportsList).forEach((key) => {
+  if (key.slice(0, match.length) !== match) {
+    return;
+  }
+
+  const importValue = exportsList[key]['import'];
+  if (key === '.') {
+    entries.push({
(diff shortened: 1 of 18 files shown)


fabian-hiller:
Thanks for your research and technical implementation! What is the advantage of the submodules and the `.d.ts` and `.d.cts` files in the `library` directory? I have not fully understood this yet.

userquin:
About submodules, here none. About d.cts and d.ts, you need to add both to allow use valibot in CJS and ESM, check typescriptlang link, TypeScript requires both.

fabian-hiller:
Do these files have to be in the `library` directory or can they also be in `dist`?

userquin:
Must be on root, ts resolution will lookup them for node10, dist folder will be regenerated on each build.

Check the resolutions entries in https://arethetypeswrong.github.io/ wirh the tgz file

fabian-hiller:
Would we save all the hassle if we generated a single `.mjs`, `.cjs`, `.d.ts` and `.d.cts` from the many files in the build step? I don't want to complicate the repository unnecessarily. Currently I am a bit averse to the many files in the `library` directory.

userquin:
Yes, we only need the index.d.cts and the unbuild.config.ts module, we can remove all the submodules, check first comment about removing submodules.

I'll cleanup the PR tmr.

Issue #11, opened 2023-07-26
Add built in type guards


JoakimDeak:
Usually when using zod a common pattern I see is defining a type guard around the `safeParse` that would look something like:
```ts
type Data = z.infer<typeof dataSchema>
const isData = (input: unknown): input is Data => {
  return dataSchema.safeParse(input).success
}
```
So I would suggest adding something like an `is` function directly in the library to fill this function so that users don't have to create these wrapper functions.

For an example let's say you have a React component that listens to messages and if it's a specific type of message then it sets some local state. Without a built in type guard it would look something like this:
```tsx
const messageDataSchema = object({
  eventId: literal('exampleEventId'),
  id: string(),
})

type MessageData = Output<typeof messageDataSchema>

const isMessageData = (input: unknown): input is MessageData => {
  return safeParse(messageDataSchema, input).success
}

const Component = () => {
  const [id, setId] = useState('')
  useEffect(() => {
    const handler = ({ data }: MessageEvent['data']) => {
      if (isMessageData(data)) {
        setId(data.id)
      }
    }
    window.addEventListener('message', handler)
    return () => {
      window.removeEventListener('message', handler)
    }
  }, [])
}
```
So we create the schema, infer that to a type, use the type to create a type guard, and then we can use that type guard in our message handler. Alternatively if we have built in type guards we can do:
```tsx
const messageDataSchema = object({
  eventId: literal('exampleEventId'),
  id: string(),
})

const Component = () => {
  const [id, setId] = useState('')
  useEffect(() => {
    const handler = ({ data }: MessageEvent['data']) => {
      if (is(messageDataSchema, data)) {
        setId(data.id)
      }
    }
    window.addEventListener('message', handler)
    return () => {
      window.removeEventListener('message', handler)
    }
  }, [])
}
```
Now we create the schema and use that directly with the type guard which results in less code for the user to write. Of course we could use `safeParse` directly and use its return value and be type safe like this:
```tsx
const messageDataSchema = object({
  eventId: literal('exampleEventId'),
  id: string(),
})

const Component = () => {
  const [id, setId] = useState('')
  useEffect(() => {
    const handler = ({ data }: MessageEvent['data']) => {
      const res = safeParse(messageDataSchema, data)
      if (res.success) {
        setId(res.data.id)
      }
    }
    window.addEventListener('message', handler)
    return () => {
      window.removeEventListener('message', handler)
    }
  }, [])
}
```
[...]

PR #13, opened 2023-07-26
GH-11: Add generic `is` type guard

JoakimDeak:


fabian-hiller:
Thank you for creating this PR. I'll check your code later.

jasikpark (approved):
looks great to me 💜 thx for building this, using it now while i wait for the merge

fabian-hiller:
We are still considering whether there are alternative names to `is` in #11. If there is no better alternative, I think we stick with `is` and then I check the implementation.

library/src/methods/is/is.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @param input The input to be tested
+ *
+ * @returns A type predicate for the given schema
+ */
+export function is<TSchema extends BaseSchema>(
+  schema: TSchema,
+  input: unknown
+): input is Output<TSchema> {
+  return safeParse(schema, input).success;
```
fabian-hiller, line 17: `safeParse` has more functionality and thus code than `is` technically required. Therefore we could change this line to the following code and remove the import above.

```ts
try {
  schema.parse(input);
  return true;
} catch (error) {
  return false;
}
```

library/src/methods/is/is.ts
```diff
@@ -0,0 +1,18 @@
+import { BaseSchema, Output } from '../../types';
+import { safeParse } from '../safeParse';
+
+/**
+ * Type guard for a given schema
+ *
+ * @param schema The scheme to be used
+ *
```
fabian-hiller, line 8: Can you remove this empty line?

fabian-hiller:
If we add `is`, we should also add `isAsync`

library/src/methods/is/is.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @param input The input to be tested
+ *
+ * @returns A type predicate for the given schema
+ */
+export function is<TSchema extends BaseSchema>(
+  schema: TSchema,
+  input: unknown
+): input is Output<TSchema> {
```
fabian-hiller, line 16: Using `Output` here is an error, because transformations have no effect. Therefore `Input` must be used.

fabian-hiller:
I'll revise your code in a moment and then merge it.

fabian-hiller:
> If we add `is`, we should also add `isAsync`

This does not seem to be technically possible at the moment: https://github.com/microsoft/TypeScript/issues/37681

PR #15, opened 2023-07-26
Add passthrough schema and related tests + fix unwrap

dmorosinotto:
Add passthrough and passthroughAsync function inspided by [Zod](https://zod.dev/?id=passthrough) to parse an ObjectSchema keeping extra props.

Hope I get the Types right ;-)

### DIFF
--- a/library/src/methods/unwrap/unwrap.ts
+++ b/library/src/methods/unwrap/unwrap.ts
@@ -11,6 +11,8 @@ import type {
   NullishSchemaAsync,
   OptionalSchema,
   OptionalSchemaAsync,
+  PassthroughSchema,
+  PassthroughSchemaAsync,
 } from '../../schemas';
 
 /**
@@ -24,6 +26,8 @@ export function unwrap<
   TSchema extends
     | OptionalSchema<any>
     | OptionalSchemaAsync<any>
+    | PassthroughSchema<any>
+    | PassthroughSchemaAsync<any>
     | NullableSchema<any>
     | NullableSchemaAsync<any>
     | NullishSchema<any>
--- a/library/src/schemas/index.ts
+++ b/library/src/schemas/index.ts
@@ -19,6 +19,7 @@ export * from './nullType';
 export * from './number';
 export * from './object';
 export * from './optional';
+export * from './passthrough';
 export * from './record';
 export * from './recursive';
 export * from './set';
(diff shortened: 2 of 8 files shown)


fabian-hiller:
Thanks a lot for this PR! I have looked at the code and think that `passthrough` should be a method and not a schema, since it does not represent a type. What is your opinion on this?

dmorosinotto:
Hi @fabian-hiller  onestly even I think that `passthrough` should be a **method** and initially put in in that directory, but when I looked at your code to take inspiration on how to implement it I initially copied from `partial` and `merge` but it doesn't work to use the `object()` constructor to implement it - Then I look at how you implement `optional` and others that are in **schemas**, and it works to me to code `passthrough` in the same way overriding the `parse` and using it as a _"wrapper"_ so I implement is in this way, but I don't have any problem to move in the other directory if you prefer.
Sincerely I haven't realized what are the main main difference/choise that guide you to put in one or the other, I only look and "similar" code and put in the same place (I'm lazy)

PS: some more discussion point:
1. if you prefer other name please let me know I just use the same name of Zod... but open to alternative refactor
2. do you think that it's correct to implement it as a _"wrapper"_  and my modification to `unwrap` or do you prefer other way to implement it?
3. about this I was even thinking to implement it changing the `object()` implementation adding another overload with an optional `strict: boolean` param  before error - but to do it I need to change the implementation of `getErrorAndPipe` handling 3 params boolean, string, Pipe[] to discriminate...  I don't know if you'll approve/prefer this kind of implementation, maybe the result is more clear because the real natura of `passthrough` is just an ObjectSchema that keep the extra params, but don't know if you like it... 
4. Or maybe on the same "idea" i can create other 2 function `objectWithExtraProps` that return ObjectSchema but simply use my parse override to keep props... 

This are all the ideas I came up when implementing it, what's you opinions?

fabian-hiller:
Currently, I am leaning towards one method. However, I don't have the time to think it through in more detail at the moment. Maybe it makes sense to move this topic to an issue for now and as soon as I get around to it, I can share my thoughts there. Regarding the name, I'm happy to hear about alternatives. I find `passthrough` ok but not perfect.

fabian-hiller:
If you create an issue, I would be interested in use cases where `passthrough` is important and makes sense.

fabian-hiller:
I added `passthrough` and `strip` in v0.14.0. Thank you for your contribution!

Issue #36, opened 2023-08-01
feature request: i18n


thundermiracle:
Thanks for your great library.
I think multilingual support is important for this library.
For example, our company chose dayjs instead of date-fns because i18n is far more easy to apply Japanese.
Would you please consider support i18n?

Add an option `i18n` to `parse` function:
```ts
parse(input: unknown, info?: ParseInfo, i18n?: Record<string, string>)
```

And get message from i18n in each function:
```ts
throw new ValiError([
  {
    validation: 'ends_with',
    origin: 'value',
    message: error || i18n?.["end_with"] || 'Invalid end',
    input,
    ...info,
  },
]);
```

So user can only define the i18n for the functions they're using to reduce bundle size. I'm not sure this implementation is OK for this library.

thundermiracle:
@fabian-hiller 
Thanks for your response. I can help implement this simple feature if it's needed. Just notify me.

fabian-hiller:
I do. Thank you for the offer.

fabian-hiller:
@zkulbeda yes, that was my first thought too. Thank you for contribution!

samuelstroschein:
@fabian-hiller We settled on a Record that uses BCP-47 language tags, similar to what @thundermiracle suggested,  see https://github.com/inlang/inlang/blob/08ab2f846a1c47455fcb038b67b1f1b7725c5fdf/source-code/core/language-tag/src/schema.ts#L18-L22.  

Using a record is the simplest API possible and should be sufficient. I am unsure whether you need to force correct LanguageTag usage like we do, given that validation schemas are defined and used by the same consumer. If valibot consumers are supposed to share validation schemas, then enforcement of BCP-47 language tags as keys of the record would make sense.

fabian-hiller:
Thanks for your feedback @samuelstroschein! I'll wait another week or two and then do a first implementation.

ivanhofer:
I really like the approach `zod` uses. It is really flexible and powerful. You can define a global `errorMap` (which is a function and not a record) and also pass it to each `.parse` call if you want to override it.

fabian-hiller:
@ivanhofer thank you for your feedback! Can you add a code example here to the comments?

thundermiracle:
Zod's error message is flexible, it can combine parameters with error messages for details. For example, error message of `includes` is not just `Invalid value`, but is `Invalid input: must include "${issue.validation.includes}"`.

See:
[...]

PR #41, opened 2023-08-02
feature: i18next

brunoibarbosa:
Implemented the i18next library to translate or customize messages without the need to modify every time you use. Online translations are not perfect and may differ from what is commonly used, so I think it would be nice if the community help to translate. I added the translation to Brazilian Portuguese.

To change the language just run the code below in the index of the project which library will be used:
```
import { i18next } from 'valibot';

i18next.init({ lng: 'pt-BR })
```

Or to customize the schemas/validations messages, use your own json file:
```
import { i18next } from 'valibot';
import * as customTranslate from './translate.json';

i18next.addResourceBundle('en', 'validation', customTranslate );
```

I hope this is helpful in some way.


fabian-hiller:
Thank you very much for creating this PR. I suspect that your implementation has a negative impact on bundle size, and a small bundle size is one of the big advantages of Valibot. Furthermore, I would like to avoid adding dependencies. I recommend you to take a look at this issue: #36

PR #42, opened 2023-08-02
Fix ISO validation hours to allow `00`

sillvva:
The existing ISO timestamp validations only allowed `01`-`09` for hours, but not `00`. This fixes that.

### DIFF
--- a/library/src/validations/isoDateTime/isoDateTime.test.ts
+++ b/library/src/validations/isoDateTime/isoDateTime.test.ts
@@ -8,7 +8,7 @@ describe('isoDateTime', () => {
     const validate = isoDateTime();
     const value1 = '2023-07-11T19:34';
     expect(validate(value1, info)).toBe(value1);
-    const value2 = '0000-01-01T01:00';
+    const value2 = '0000-01-01T00:00';
     expect(validate(value2, info)).toBe(value2);
     const value3 = '9999-12-31T23:59';
     expect(validate(value3, info)).toBe(value3);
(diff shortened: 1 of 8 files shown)


fabian-hiller:
Are you sure that ISO times start at `00:00`? In my tests I added `00:00` as an invalid value and `01:00` as the first valid value. Can you research this and share your findings here?

sillvva:
In my testing yesterday, I actually ran into this issue. I have a date field that was showing as invalid. When converted to ISO, the hours were 00. I plugged it into [regex101.com](https://www.regex101.com) and was able to fix it with this change.

https://en.wikipedia.org/wiki/ISO_8601

Hours range from 00 to 24.

As of ISO 8601-1:2019/Amd 1:2022, [midnight](https://en.wikipedia.org/wiki/Midnight) may be referred to as "00:00:00", corresponding to the instant at the beginning of a calendar day; or "24:00:00", corresponding to the instant at the end of a calendar day.

fabian-hiller:
Thank you very much for your research. I hope to find time tonight to check your code and merge this PR.

fabian-hiller:
Sorry, I haven't gotten around to it yet. I'll try to check the changes tonight.

fabian-hiller:
If I understand it correctly, `24:00:00` has been removed from the standard. But presumably it is still accepted as input. Since it complicates the regex to include `24` with the hours, I won't do that for now. I'll check your code in a moment and update the testes. Thanks again for your contribution!

PR #43, opened 2023-08-02
docs: add typeschema to ecosystem

decs:
Following up on https://github.com/decs/typeschema/issues/11, this pull request lists TypeSchema on the ecosystem page


fabian-hiller:
Thank you! Since it is a universal adapter that converts Valibot to a neutral scheme, doesn't it then fit perfectly into the cathegory `Valibot to X`?

decs:
makes sense. I'll update the commit

Issue #19, opened 2023-07-27
Any plan to improve performance?


southorange0929:
I write a simple benchmark with real case. 
- zod
- yup
- joi
- valibot

```ts
import Benchmark from "benchmark";

import { z } from "zod";
import * as yup from "yup";
import Joi from "joi";
import * as valibot from "valibot";

const SUITE_NAME = "";
const suite = new Benchmark.Suite("Validate Benchmark");

const value = {
  latitude: 10,
  longitude: 10,
  scale: 10,
  name: "",
  address: "",
  powerLevel: "low",
};

const stringSchema = z.object({
  latitude: z.number(),
  longitude: z.number(),
  scale: z.number().min(5).max(18),
  name: z.string().optional(),
  address: z.string().optional(),
  powerLevel: z.enum(["low", "medium", "high"]),
});

const yupSchema = yup.object({
  latitude: yup.number(),
  longitude: yup.number(),
  scale: yup.number().min(5).max(18),
  name: yup.string().optional(),
  address: yup.string().optional(),
  powerLevel: yup.mixed().oneOf(["low", "medium", "high"]),
});

const joiSchema = Joi.object({
  latitude: Joi.number(),
  longitude: Joi.number(),
  scale: Joi.number().min(5).max(18),
  name: Joi.string().optional(),
  address: Joi.string().optional(),
  powerLevel: Joi.string().valid("low", "medium", "high"),
});

const valibotSchema = valibot.object({
  latitude: valibot.number(),
  longitude: valibot.number(),
  scale: valibot.number([valibot.minRange(5), valibot.maxRange(18)]),
  name: valibot.optional(valibot.string()),
  address: valibot.optional(valibot.string()),
  powerLevel: valibot.enumType(["low", "medium", "high"]),
});

suite
  .add("zod validate", () => {
    stringSchema.safeParse(value);
  })
  .add("yup validate", () => {
    yupSchema.validate(value);
  })
  .add("joi validate", () => {
    joiSchema.validate(value);
  })
  .add("valibot validate", () => {
    valibot.safeParse(valibotSchema, value);
  })
  .on("cycle", (e: Benchmark.Event) => {
    console.log(`${SUITE_NAME}: ${e.target}`);
  })
  .on("complete", function () {
    console.log("Fastest is " + this.filter("fastest").map("name"));
    process.exit(0);
  })
  .run();

```

[...]

PR #46, opened 2023-08-03
Reduce spread operators to improve runtime performance

naruaway:
I found that removing the object spread operator greatly improve the runtime performance of `parse`. Without this PR (`main`: 506149a2de9755f91a033f340a42917b8457c82e), I confirmed that in many benchmarks, Valibot is much slower than Zod but after this PR, Valibot is mostly faster than Zod.

I created this PR so that people can run benchmarks and we can discuss the actual implementation details to move forward.

## Benchmarks
I used [this benchmark](https://github.com/fabian-hiller/valibot/issues/19#issue-1824216605) to confirm this PR makes a big impact at least for my environment (MacBook Air M2). I would appreciate if more people can confirm this. Note that I removed yup and Joi since they seemed to affect the benchmark results. We should make sure the benchmark is actually correct but I think **it's safe to say that Valibot can be actually faster than Zod with small tweaks like this PR**

- Before this PR (`main`: 506149a2de9755f91a033f340a42917b8457c82e)
```
: zod validate x 2,258,349 ops/sec ±1.14% (98 runs sampled)
: valibot validate x 713,707 ops/sec ±0.48% (93 runs sampled)
Fastest is zod validate
```

- After this PR (97ebba69c1cda0edabe9a4086054da3d1939edbf)
```
: zod validate x 2,289,850 ops/sec ±0.72% (101 runs sampled)
: valibot validate x 2,762,403 ops/sec ±0.07% (98 runs sampled)
Fastest is valibot validate
```

Note that I also tested this in https://github.com/moltar/typescript-runtime-type-benchmarks/pull/1112

## Backstory 
Recently I got amazed how Valibot is designed ❤️  ([Twitter thread in Japanese](https://twitter.com/naruaway/status/1687000165406904320)).

The implementation is really clean, modular, and simple. Since I am a huge fun of Zod and its bundle size is big, I was trying to invent special compiler for Zod to reduce the bundle size and it turned out that it was futile since Zod's implementation is hard to untangle and simplify.
So I just decided to use Valibot for my personal projects and I also want to support wider adoption for more serious production usage as well.

However, if we think of Valibot as "Zod replacement", runtime performance of `parse` is definitely a concern. For example, if we implement some validation logic in server side, we might suffer from DoS attack more often than when using Zod if the validation performance of Valibot is slower. Actually I expected Valibot is faster since the implementation is simpler than Zod. But it turned out that it's much slower than Zod in several benchmarks ([one example](https://github.com/fabian-hiller/valibot/issues/19#issue-1824216605)). I also tested several patterns locally and Zod was always much faster. After I played with Valibot codebase, I just found that we use the object spread operator in the hotpath of `parse` and removing that greatly improves the performance in my environment.

## Misc
I also tested some other things like caching object keys in `object.ts` and it looks like it's also improving the performance in some benchmarks. But I want to make sure we have good suite of benchmarks before discussing further on this. Reducing the object spread operator seems to have the biggest impact.

### DIFF
--- a/library/CHANGELOG.md
+++ b/library/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the library will be documented in this file.
 
+## vX.X.X (Month DD, YYYY)
+
+- Add `getPipeInfo` util and refactor code for performance reasons (pull request #46)
+
 ## v0.12.0 (August 11, 2023)
 
 - Change input type of `mimeType` validation to `Blob`
--- a/library/src/schemas/any/any.ts
+++ b/library/src/schemas/any/any.ts
@@ -1,5 +1,5 @@
 import type { BaseSchema, Pipe } from '../../types.ts';
-import { executePipe } from '../../utils/index.ts';
+import { executePipe, getPipeInfo } from '../../utils/index.ts';
 
 /**
  * Any schema type.
@@ -36,7 +36,7 @@ export function any(pipe: Pipe<any> = []): AnySchema {
      * @returns The parsed output.
      */
     parse(input, info) {
-      return executePipe(input, pipe, { ...info, reason: 'any' });
+      return executePipe(input, pipe, getPipeInfo(info, 'any'));
     },
   };
 }
--- a/library/src/schemas/any/anyAsync.ts
+++ b/library/src/schemas/any/anyAsync.ts
@@ -1,5 +1,5 @@
 import type { BaseSchemaAsync, PipeAsync } from '../../types.ts';
-import { executePipeAsync } from '../../utils/index.ts';
+import { executePipeAsync, getPipeInfo } from '../../utils/index.ts';
 
 /**
  * Any schema type.
@@ -36,7 +36,7 @@ export function anyAsync(pipe: PipeAsync<any> = []): AnySchemaAsync {
      * @returns The parsed output.
      */
     async parse(input, info) {
-      return executePipeAsync(input, pipe, { ...info, reason: 'any' });
+      return executePipeAsync(input, pipe, getPipeInfo(info, 'any'));
     },
   };
 }
--- a/library/src/schemas/array/array.ts
+++ b/library/src/schemas/array/array.ts
@@ -4,6 +4,7 @@ import {
   executePipe,
   getCurrentPath,
   getErrorAndPipe,
+  getPipeInfo,
 } from '../../utils/index.ts';
 
 /**
@@ -137,10 +138,11 @@ export function array<TArrayItem extends BaseSchema>(
       }
 
       // Execute pipe and return output
-      return executePipe(output as Output<TArrayItem>[], pipe, {
-        ...info,
-        reason: 'array',
-      });
+      return executePipe(
+        output as Output<TArrayItem>[],
+        pipe,
+        getPipeInfo(info, 'array')
+      );
     },
   };
 }
--- a/library/src/schemas/array/arrayAsync.ts
+++ b/library/src/schemas/array/arrayAsync.ts
@@ -10,6 +10,7 @@ import {
   executePipeAsync,
   getCurrentPath,
   getErrorAndPipe,
+  getPipeInfo,
 } from '../../utils/index.ts';
 
 /**
@@ -133,10 +134,11 @@ export function arrayAsync<TArrayItem extends BaseSchema | BaseSchemaAsync>(
       }
 
       // Execute pipe and return output
-      return executePipeAsync(output as Output<TArrayItem>[], pipe, {
-        ...info,
-        reason: 'array',
-      });
+      return executePipeAsync(
+        output as Output<TArrayItem>[],
+        pipe,
+        getPipeInfo(info, 'array')
+      );
     },
   };
 }
--- a/library/src/schemas/bigint/bigint.ts
+++ b/library/src/schemas/bigint/bigint.ts
@@ -1,6 +1,10 @@
 import { ValiError } from '../../error/index.ts';
 import type { BaseSchema, Pipe } from '../../types.ts';
-import { executePipe, getErrorAndPipe } from '../../utils/index.ts';
+import {
+  executePipe,
+  getErrorAndPipe,
+  getPipeInfo,
+} from '../../utils/index.ts';
 
 /**
  * Bigint schema type.
@@ -71,7 +75,7 @@ export function bigint(
       }
 
       // Execute pipe and return output
-      return executePipe(input, pipe, { ...info, reason: 'bigint' });
+      return executePipe(input, pipe, getPipeInfo(info, 'bigint'));
     },
   };
 }
--- a/library/src/schemas/bigint/bigintAsync.ts
+++ b/library/src/schemas/bigint/bigintAsync.ts
@@ -1,6 +1,10 @@
 import { ValiError } from '../../error/index.ts';
 import type { BaseSchemaAsync, PipeAsync } from '../../types.ts';
-import { executePipeAsync, getErrorAndPipe } from '../../utils/index.ts';
+import {
+  executePipeAsync,
+  getErrorAndPipe,
+  getPipeInfo,
+} from '../../utils/index.ts';
 
 /**
  * Bigint schema async type.
@@ -77,7 +81,7 @@ export function bigintAsync(
       }
 
       // Execute pipe and return output
-      return executePipeAsync(input, pipe, { ...info, reason: 'bigint' });
+      return executePipeAsync(input, pipe, getPipeInfo(info, 'bigint'));
     },
   };
 }
--- a/library/src/schemas/blob/blob.ts
+++ b/library/src/schemas/blob/blob.ts
@@ -1,6 +1,10 @@
 import { ValiError } from '../../error/index.ts';
 import type { BaseSchema, Pipe } from '../../types.ts';
-import { executePipe, getErrorAndPipe } from '../../utils/index.ts';
+import {
+  executePipe,
+  getErrorAndPipe,
+  getPipeInfo,
+} from '../../utils/index.ts';
 
 /**
  * Blob schema type.
@@ -71,7 +75,7 @@ export function blob(
       }
 
       // Execute pipe and return output
-      return executePipe(input, pipe, { ...info, reason: 'blob' });
+      return executePipe(input, pipe, getPipeInfo(info, 'blob'));
     },
   };
 }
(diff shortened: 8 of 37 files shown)


fabian-hiller:
Thank you for contributing to the success of Valibot! 🤖

naruaway:
@fabian-hiller thanks for taking a look! Good luck for your bachelor thesis  ❤️ 
Looking forward to see your updates on September!

Just curious, does it mean you will never merge any PRs like this until September? Maybe you are expecting big refactoring

fabian-hiller:
No, I think that Valibot already has a good foundation. I don't expect any big changes at the moment. If I make better progress with my bachelor thesis than expected, I might merge this PR much earlier.

naruaway:
@fabian-hiller 
> No, I think that Valibot already has a good foundation.
Good to hear because my impression was the same! Clean and simple, small codebase ❤️ 

Let me actually try to build extensive benchmark suite so that we can make sure Valibot is always fast in any environment. I'll create another issue to track benchmarking in general later

naruaway:
Now I created tracking issue for performance improvement in general https://github.com/fabian-hiller/valibot/issues/73 so we can use this PR only for the specific change

fabian-hiller:
I have now done this investigation myself. You are right. The impact on performance is extreme. In my investigation, I noticed that the following makes no difference to performance. Therefore, I prefer the variant that requires less code.

```ts
// With ternary operator
export const mergeInfoWithReason = (info: ParseInfo | undefined, reason: IssueReason) => {
  return info ? {
    origin: info.origin,
    path: info.path,
    abortEarly: info.abortEarly,
    abortPipeEarly: info.abortPipeEarly,
    reason,
  } : { reason };
}

// Without ternary operator
export const mergeInfoWithReason = (info: ParseInfo | undefined, reason: IssueReason) => {
  return {
    origin: info?.origin,
    path: info?.path,
    abortEarly: info?.abortEarly,
    abortPipeEarly: info?.abortPipeEarly,
    reason,
  };
}
```

I appreciate your research and would leave it up to you to implement this change for the entire library. Otherwise, I can implement it and mention you in the release notes. Before we go into the implementation, I would clarify a few details about the final code with you.

naruaway:
@fabian-hiller thanks for following up on this!
Now I am finishing up an improving of https://github.com/naruaway/valibot-benchmarks to be able to leave an evidence in reproducible way since having reliable benchmarks is critical for any performance improvements. I will restart working on this tomorrow evening  👍 

Yeah I would aim for minimal code to get the same level of perfs across major runtimes (combinations of (macOS (ARM), Linux (x86)) x (Bun(JavaScriptCore) and Node.js/Deno (V8)))

fabian-hiller:
> I appreciate your research and would leave it up to you to implement this change for the entire library. Otherwise, I can implement it and mention you in the release notes. Before we go into the implementation, I would clarify a few details about the final code with you.

Do you want to make the changes, or should I do it? If so, in what time frame can you do it?

I'm reviewing the performance today and tomorrow as part of my thesis. I want to implement any improvements over the weekend to be able to focus on other areas of the paper next week.

naruaway:
@fabian-hiller Ok, I was planning to finish up benchmark suite thing first but I anyway we are on the same page that this is significant perf booster, let me do the change now (I have time now before sleeping) 👍 I might follow up precise perf results / small tweaks later then

fabian-hiller:
Great! Can you rename `mergeInfoWithReason` to `getPipeInfo` and use `function` instead of an arrow function?

naruaway:
@fabian-hiller I just realized that we might change `executePipe` itself to accept `ParseInfo` and `IssueReason` separately but just to get quick perf boost without further churn, let me do the simplest change (`getPipeInfo` function everywhere) first and potentially file another PR

fabian-hiller:
You're right. The logic can also be in the pipe function.

naruaway:
@fabian-hiller, anyway I pushed `getPipeInfo` change and I believe I did not miss any other places:
```
# On main branch
$ rg -l '...info, reason'|wc -l
23

# On this PR
$ rg -l '...info, reason'|wc -l
0
```
Could you review and merge if it looks good? (of course feel free to tweak or point out anything 👍 )

Note that this is not enough to "beat" Zod in benchmarks for valid data most of the time. We also need to eliminate the other spread operator like [this](https://github.com/fabian-hiller/valibot/blob/ce49b0b4809a62035767c4b7697fabddd871392c/library/src/schemas/object/object.ts#L114-L117). And this cannot be put in executePipe. So let me create similar util for now. I'll add the change in this PR if you do not merge this PR yet. If you merge this PR, I'll file another PR

fabian-hiller:
I'll check it in 30 minutes and give you feedback. Feel free to create further PRs for the other spread operators. I'll continue later with my test as well.

naruaway:
Ah I missed  multi line patterns:
```
...info,
reason: 'tuple',
```
Thanks for the follow up and refactor 👍

fabian-hiller:
Thanks again for your contribution! 🤝

naruaway:
FYI I ran a benchmark suite.
## vs. Previous Valibot
I ran [naruaway/valibot-benchmarks](https://github.com/naruaway/valibot-benchmarks) using Node.js v20.5.0 and Bun v0.7.2 on my MacBook Air M2 to compare this PR (609308f177c5d150f7e5bc38ead40669a22d50ae) against the main branch (ce49b0b4809a62035767c4b7697fabddd871392c).

- Significant speed up (more than 100%) is observed for Node.js. In Bun, we saw up to 16% speed up. Different engines behave differenly.
- the most notable case is **182% faster for Node.js** :tada: for [`optional_nullable` data set](https://github.com/naruaway/valibot-benchmarks/tree/ac99da10597c9b4c98e20a4d1a5df74aba663896/test_data/optional_nullable)
- This change does not affect perfs for invalid data much, which is slow anyway

<details>
<summary>Valibot in this PR compared against Valibot `main`</summary>
<pre>
Running benchmarks using nodejs
Showing the result when running runtime environment (e.g. Node.js) 100 times for each:
case: "deep", data: "valid"
 48% faster
  baseline: 90220 ops/s, +- 2% stddev
  target: 133560 ops/s, +- 2% stddev
case: "many_features", data: "invalid"
 1% faster
  baseline: 16301 ops/s, +- 1% stddev
  target: 16482 ops/s, +- 2% stddev
case: "many_features", data: "valid"
 22% faster
  baseline: 63010 ops/s, +- 2% stddev
  target: 76972 ops/s, +- 1% stddev
case: "optional_nullable", data: "valid"
 182% faster
  baseline: 72746 ops/s, +- 2% stddev
  target: 205313 ops/s, +- 2% stddev
case: "wide", data: "invalid"
 0% faster
  baseline: 3863 ops/s, +- 2% stddev
  target: 3878 ops/s, +- 2% stddev
case: "wide", data: "valid"
 146% faster
  baseline: 53734 ops/s, +- 2% stddev
  target: 132262 ops/s, +- 2% stddev
Running benchmarks using bun
Showing the result when running runtime environment (e.g. Node.js) 100 times for each:
case: "deep", data: "valid"
 12% faster
  baseline: 210049 ops/s, +- 2% stddev
  target: 234258 ops/s, +- 2% stddev
case: "many_features", data: "invalid"
 1% slower
  baseline: 81017 ops/s, +- 2% stddev
  target: 80411 ops/s, +- 2% stddev
case: "many_features", data: "valid"
 7% faster
  baseline: 254261 ops/s, +- 2% stddev
  target: 272429 ops/s, +- 2% stddev
case: "optional_nullable", data: "valid"
 16% faster
  baseline: 174454 ops/s, +- 5% stddev
  target: 201589 ops/s, +- 5% stddev
case: "wide", data: "invalid"
 0% slower
  baseline: 20467 ops/s, +- 3% stddev
  target: 20431 ops/s, +- 3% stddev
case: "wide", data: "valid"
 16% faster
  baseline: 130907 ops/s, +- 3% stddev
  target: 151470 ops/s, +- 2% stddev
out of 12 cases, 10 cases (83%) were faster
<pre>
</details>


## vs. Zod

I ran a similar benchmark as the above but to compare this PR against Zod 

- Still Valibot is slower than Zod for most of the cases, even for valid data
- But for some schema and engine (Bun), now it's faster
- Note that when the data is invalid, Valibot is consistently slower than Zod for any engine / schemas

<details>
<summary>Valibot in this PR compared against Zod v3.21.4</summary>
<pre>
Running benchmarks using nodejs
Showing the result when running runtime environment (e.g. Node.js) 100 times for each:
case: "deep", data: "valid"
 22% slower
  baseline: 171942 ops/s, +- 2% stddev
  target: 133599 ops/s, +- 1% stddev
case: "many_features", data: "invalid"
 50% slower
  baseline: 33022 ops/s, +- 1% stddev
  target: 16586 ops/s, +- 1% stddev
case: "many_features", data: "valid"
 15% slower
  baseline: 91853 ops/s, +- 3% stddev
  target: 77764 ops/s, +- 2% stddev
case: "optional_nullable", data: "valid"
 2% faster
  baseline: 205524 ops/s, +- 3% stddev
  target: 209799 ops/s, +- 1% stddev
case: "wide", data: "invalid"
 83% slower
  baseline: 22717 ops/s, +- 2% stddev
  target: 3898 ops/s, +- 1% stddev
case: "wide", data: "valid"
 0% slower
  baseline: 133707 ops/s, +- 4% stddev
  target: 133632 ops/s, +- 2% stddev
Running benchmarks using bun
Showing the result when running runtime environment (e.g. Node.js) 100 times for each:
case: "deep", data: "valid"
 59% faster
  baseline: 148668 ops/s, +- 3% stddev
  target: 235739 ops/s, +- 3% stddev
case: "many_features", data: "invalid"
 60% slower
  baseline: 205429 ops/s, +- 0% stddev
  target: 82011 ops/s, +- 0% stddev
case: "many_features", data: "valid"
 43% faster
  baseline: 191894 ops/s, +- 1% stddev
  target: 274992 ops/s, +- 1% stddev
case: "optional_nullable", data: "valid"
 18% slower
  baseline: 248983 ops/s, +- 1% stddev
  target: 204771 ops/s, +- 3% stddev
case: "wide", data: "invalid"
 72% slower
  baseline: 75206 ops/s, +- 1% stddev
  target: 20904 ops/s, +- 1% stddev
case: "wide", data: "valid"
 20% slower
  baseline: 192068 ops/s, +- 2% stddev
  target: 153557 ops/s, +- 1% stddev
out of 12 cases, 3 cases (25%) were faster
<pre>
</details>

naruaway:
I'll work on https://github.com/fabian-hiller/valibot/pull/92 tomorrow evening in Japan

 @fabian-hiller , if you are in a hurry, feel free to apply my intended changes in your way without waiting for me 👍 Together with https://github.com/fabian-hiller/valibot/pull/47, I just now reconfirmed that Valibot gets faster than Zod for both https://github.com/naruaway/valibot-benchmarks and https://github.com/moltar/typescript-runtime-type-benchmarks **excluding "cases with invalid data"**

PR #47, opened 2023-08-03
perf(object): cache object keys

naruaway:
I confirmed that this improves parse performance since checking object keys is relatively expensive operation (probably similar reason as https://github.com/fabian-hiller/valibot/pull/46).

I'll add benchmark results later but the code change itself might be already acceptable. I think the only drawback of this change is slightly slower schema creation.

Note that Zod is [caching object keys here](https://github.com/colinhacks/zod/blob/78a409012a4dc34a455f5c4a7e028ca47c921e1b/src/types.ts#L2238-L2243) but the logic will run on the initial `parse` call. We can do the same thing but it will increase the code size. So far I personally prefer not having such a complicated logic.

### DIFF
--- a/library/src/schemas/object/object.ts
+++ b/library/src/schemas/object/object.ts
@@ -59,6 +59,9 @@ export function object<TObjectShape extends ObjectShape>(
   // Get error and pipe argument
   const { error, pipe } = getErrorAndPipe(arg2, arg3);
 
+  // Cache object keys for better .parse() performance
+  const objectKeys = Object.keys(object);
+
   // Create and return object schema
   return {
     /**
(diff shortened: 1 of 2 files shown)


fabian-hiller:
Thank you for this insight. Feel free to share your benchmarks here in the comments. Since I'm currently writing my bachelor thesis, it may take a while until I merge changes that are currently not absolutely necessary.

naruaway:
Just realized that @fabian-hiller already implemented the cache in https://github.com/fabian-hiller/valibot/commit/dbb3df06866387b1ca85df64a3bcab9151c6a31f with lazy initialization 👍 
Closing the PR now

PR #49, opened 2023-08-04
docs: add Hono to ecosystem page

ZerNico:
Hono now includes support for Valibot (https://github.com/honojs/middleware/pull/102) so we should add it to the list.

Thanks for your work!


fabian-hiller:
Thank you! It is an honor!

PR #62, opened 2023-08-05
feature: integer validation

gmaxlev:
I noticed that the project lacks a validation function for integer values, while other validation libraries typically support this feature. Therefore, I have implemented a new validation function to check for integer values.


fabian-hiller:
Thank you! I will check and merge your code later!

PR #67, opened 2023-08-06
security: Mitigate Prototype Pollution Vulnerability

NiclasHaderer:
This pull request addresses a critical issue in the current record schema that makes it susceptible to prototype pollution. The vulnerability arises due to the lack of key validation, allowing certain keys to be exploited, leading to potential security risks.

To mitigate this vulnerability, I have implemented a key blacklisting mechanism. By doing so, specific keys prone to prototype pollution are disallowed within the record schema.

### DIFF
--- a/library/src/schemas/record/record.ts
+++ b/library/src/schemas/record/record.ts
@@ -8,6 +8,8 @@ import {
 import { type StringSchema, string } from '../string/index.ts';
 import type { RecordOutput, RecordInput } from './types.ts';
 
+const BLOCKED_KEYS = new Set(['__proto__', 'prototype', 'constructor']);
+
 /**
  * Record key type.
  */
@@ -163,6 +165,13 @@ export function record<
       // Parse each key and value by schema
       // Note: `Object.entries(...)` converts each key to a string
       Object.entries(input).forEach(([inputKey, inputValue]) => {
+
+        // Check if key is blocked
+        if (BLOCKED_KEYS.has(inputKey)) {
+            return;
+        }
+
+
         // Get current path
         const path = getCurrentPath(info, {
           schema: 'record',
--- a/library/src/schemas/record/record.test.ts
+++ b/library/src/schemas/record/record.test.ts
@@ -6,6 +6,7 @@ import { number } from '../number/index.ts';
 import { string } from '../string/index.ts';
 import { union } from '../union/index.ts';
 import { record } from './record.ts';
+import {any} from "../any";
 
 describe('record', () => {
   test('should pass only objects', () => {
@@ -87,4 +88,14 @@ describe('record', () => {
     expect(output3).toEqual(transformInput());
     expect(output4).toEqual(transformInput());
   });
+
+
+  test("should not be vulnerable to prototype pollution", () => {
+    const schema = record(string(), any());
+    const input = JSON.parse('{"__proto__":{"polluted":"yes"}}');
+    expect(input.__proto__.polluted).toBe("yes");
+    expect(({} as any).polluted).toBeUndefined();
+    const parsed = parse(schema, input);
+    expect((parsed as any).polluted).toBeUndefined();
+  })
 });


fabian-hiller:
Thank you!

library/src/schemas/record/record.ts
```diff
@@ -163,6 +165,13 @@ export function record<
       // Parse each key and value by schema
       // Note: `Object.entries(...)` converts each key to a string
       Object.entries(input).forEach(([inputKey, inputValue]) => {
+
+        // Check if key is blocked
+        if (BLOCKED_KEYS.has(inputKey)) {
```
BastiDood, line 170: Hello there! A little bit late to the party here, but I'd like to note that although the `Set#has` method is $O(1)$, that constant coefficient is often significantly expensive due to the underlying hashing algorithms.

With that said, I believe we can speed this up by forgoing the hashed `Set` altogether and just perform a linear search on the blocked keys themselves. This is particularly efficient since there are only three keys to check here. The CPU is thus likely to put these strings closer to its cache. This is opposed to using a `Set` which requires a lot of intermediate computations for the hash.

In summary, we should use an array for the `BLOCKED_KEYS` instead of a `Set` because the $O(n)$ array search (for three elements) is vastly superior over the expensive $O(1)$ lookup of the `Set`.

> **DISCLAIMER:** I do not have real-world benchmarks of `valibot` to back these claims up, but I know that this is generally true for modern CPUs.

NiclasHaderer: Hi, interesting thought. I tested a bit and it looks like both about the same speed. If anything the set is a tiny bit faster, but just a fraction.
```javascript
const BLOCKED_KEYS_SET = new Set(['__proto__', 'prototype', 'constructor']);
const BLOCKED_KEYS_ARRAY = ['__proto__', 'prototype', 'constructor'];

// Generate a random string of random length between 3 and 10 characters
function getRandomString() {
  const characters = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789';
  const length = Math.floor(Math.random() * 8) + 3;
  let result = '';
  for (let i = 0; i < length; i++) {
    const randomIndex = Math.floor(Math.random() * characters.length);
    result += characters.charAt(randomIndex);
  }
  return result;
}

function measureTime(callback) {
  const startTime = process.hrtime.bigint();
  callback();
  const endTime = process.hrtime.bigint();
  return Number(endTime - startTime) / 1e6; // Convert nanoseconds to milliseconds
}

function testSetLookup() {
  const numberOfHits = 10000;
  const numberOfRandomStrings = 100 * numberOfHits;
  for (let i = 0; i < numberOfHits; i++) {
    BLOCKED_KEYS_SET.has('constructor');
    BLOCKED_KEYS_SET.has('__proto__');
    BLOCKED_KEYS_SET.has('nonexistent');
  }
  for (let i = 0; i < numberOfRandomStrings; i++) {
    BLOCKED_KEYS_SET.has(getRandomString());
  }
}

function testArrayLookup() {
  const numberOfHits = 10000;
  const numberOfRandomStrings = 100 * numberOfHits;
  for (let i = 0; i < numberOfHits; i++) {
    BLOCKED_KEYS_ARRAY.includes('constructor');
    BLOCKED_KEYS_ARRAY.includes('__proto__');
    BLOCKED_KEYS_ARRAY.includes('nonexistent');
  }
  for (let i = 0; i < numberOfRandomStrings; i++) {
    BLOCKED_KEYS_ARRAY.includes(getRandomString());
  }
}

// Warm up just-in-time compiler
for (let i = 0; i < 20; i++) {
  testArrayLookup();
}

// Perform benchmark for Array-based lookup
const arrayLookupTime = measureTime(testArrayLookup);
console.log(`Array lookup time: ${arrayLookupTime.toFixed(3)} ms`);

for (let i = 0; i < 20; i++) {
  testSetLookup();
}

// Perform benchmark for Set-based lookup
const setLookupTime = measureTime(testSetLookup);
console.log(`Set lookup time: ${setLookupTime.toFixed(3)} ms`);

const fasterBy = (setLookupTime / arrayLookupTime).toFixed(2);
console.log(`Array lookup is ${fasterBy} times faster than Set lookup.`);
```

BastiDood: That's fascinating! The V8 engine is probably doing some fancy optimization for that. I ran your script on my machine just to verify your results. Indeed, my first run did replicate what you found. But then I closed all of my applications and windows to make sure that the Node.js process has fewer competitors. These are my results (in milliseconds) over 10 runs _after my initial cold start_:

Run|Array|Set
:--|--:|--:
1 | 110.556 | 110.646
2 | 110.171 | 110.613
3 | 109.749 | 109.963
4 | 110.417 | 110.644
5 | 110.252 | 110.881
6 | 110.612 | 117.636
7 | 112.153 | 110.141
8 | 110.070 | 109.997
9 | 110.926 | 110.340
10 | 110.578 | 114.066

Observe that in most of the runs, the array consistently (but ever so marginally!) outperforms the set.

> Here are the relevant hardware specifications of my machine: **Intel® Core™ i7-10750 CPU @ 2.60GHz (12 CPUs)** with **8192 MB of RAM** on **Windows 10 Build 19045.3271**.

fabian-hiller: In summary, your recommendation remains to replace set with an array?

BastiDood: Yes, that is the goal. I have submitted #68 for this reason.

PR #71, opened 2023-08-06
fix(ci): enable CI on pull requests

BastiDood:
Currently, CI does not run in all pull requests. In fact, the only time it runs is when the branch of the forked repository also happens to be named `main` (i.e., the default branch). Since it is generally a best practice to branch out of `main` even in a fork, resulting PRs back into Valibot does not trigger CI.

This PR simply removes the `main` branch restriction on the `pull_request` hook for GitHub Actions, thereby allowing the `node` workflow to run in PRs.

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -2,9 +2,8 @@ name: Node.js CI
 
 on:
   push:
-    branches: ['main']
+    branches: main
   pull_request:
-    branches: ['main']
 
 jobs:
   test:


fabian-hiller:
Thank you very much! Does this have any disadvantages? Is there a limit that can be reached if the CI is run too often?

BastiDood:
Fortunately, [GitHub Actions is free and unlimited for public repositories](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#about-billing-for-github-actions)! 🥳

Should we ever go beyond any undocumented limits (for some reason?), I'm sure GitHub is open to sponsor more compute minutes, especially for popular open-source projects.

Issue #73, opened 2023-08-07
Tracking issue: Performance improvement


naruaway:
This is a tracking issue to discuss & improve the performance of Valibot.
I created a benchmark suite for Valibot so that we can track its runtime performance in many situations since existing benchmarks were failing to capture the real world schema and data diversity.

The current latest result can be checked from https://naruaway.github.io/valibot-benchmarks/

Currently we see the following interesting observations:

- Most of the cases, Valibot is slower than Zod
- Valibot performs better in Bun (JavaScriptCore)
- When the data is invalid, Valibot is much slower than Zod, which is probably because we rely on exception handling for failure case

I'll continue improving [my benchmark suite](https://github.com/naruaway/valibot-benchmarks) and will post any interesting information here.

We already have several performance improvement PRs: https://github.com/fabian-hiller/valibot/pulls?q=is%3Apr+is%3Aopen+label%3Aperformance

I'll post benchmark result for each PR later

fabian-hiller:
Thanks for your research and the benchmark suite. This week I will focus exclusively on performance for a few days and probably release a new version with performance improvements.

@BastiDood had recently checked the performance of `.forEach` vs. `for...of` and we have merged the first changes with PR #68. Let's use this issue to share more insights here to make Valibot maximally fast. 🔥

fabian-hiller:
I expect to start my performance research on Wednesday or Thursday and then share my findings with you here. I plan to test the validation for the success and failure case under normal and extreme conditions between different libraries. For Valibot, the most important thing is to find the vulnerabilities in the code that are currently affecting performance. Thank you @naruaway for the vulnerabilities you have already found.

naruaway:
Hi @fabian-hiller, I updated (actually, a lot :) ) https://github.com/naruaway/valibot-benchmarks and now it can easily compare different versions of Valibot as well, not only Zod, for example, here is the excerpt of the performance of https://github.com/fabian-hiller/valibot/pull/46 🎉 :

<img width="1440" alt="reduce-spread" src="https://github.com/fabian-hiller/valibot/assets/2931577/2ce8047c-40d0-4059-9f80-88626960978c">

Also, I confirmed that in my MacBook, after my changes Valibot is faster than Zod except for "failure" data. I plan to set up GitHub actions to run it also in Linux and Windows.

[...]

Issue #48, opened 2023-08-04
Consider "ctx.addIssue" style rather than throwing ValiError


naruaway:
Throwing errors is really expensive operation so even if we can make `parse` fast for "success" case, "failure" case will be still slow. This means if we use Valibot for server side validation logic, malicious people might be able to attempt DoS by sending "failure" payload.

I created this issue to discuss architectural / philosophical possibility to stop throwing ValiError without compromising other factors such as bundle size and code simplicity.

fabian-hiller:
Thank you for this insight. Do you have an idea how the implementation could look like here? With a global object or array? Or a parameter that is passed through and modified directly?

fabian-hiller:
Thank you for the code example! I will give it some thought.

PR #77, opened 2023-08-07
First writeup of migration from Zod guide

FlorianDevPhynix:
Planning on maybe migrating to valibot, so I thought I could write a guide for it.
Not entirely sure how well I did.
Probably forgot some things. Especially with the Special Examples / Equivalents


fabian-hiller:
Thank you! Will have a look at it in the coming days.

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
```diff
@@ -1,7 +1,101 @@
...
+### Separate imports
+
+If your already using the different Schemas separately, then you only need to additionally import some of the <Link href="/guides/pipelines/">Pipeline</Link> methods. Again, many of the Schemas are compatible.
+
+## Example
+
+Let me explain it to you with an example.
+
+```ts
+import { z } from 'zod'; // 13.1 kB
```
naruaway, line 40: Probably not only this place but I found this way of showing size is kind of confusing since this is gzipped size. If not gzipped, this should be around 52 kB. It's confusing since gzip is not the only option (e.g. brotli)

Maybe we should use raw size in every place in the docs

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
```diff
@@ -1,7 +1,101 @@
...
+After this setup you will have to change every **Zod Schema**,
+but you don't have to change all of them just yet.
+
+> **This is a Process, so take your time!** *You can always take a break, and I will still be here, waiting for you.*
+
+Theres usually two ways to write these Schemas with Zod:
+
+### The `z` Object
+
+> Zod makes heavy use of it's `z` object export, which contains every schema type and validation function. This is one of Zod's disadvantages, because modern bundlers can't remove unused functions. Learn more <Link href="/guides/comparison/">here</Link>.
```
naruaway, line 27: This might be slightly confusing since the primary reason why Zod cannot support efficient tree shaking is not because it's using `z` object export. Improving this should be a low hanging fruits for Zod since they can just export symbols separately and enforce people to use `import * as z from 'zod'`. The primary reason is that they are using classes and most of features are implemented as methods, which cannot be tree shakable. They even have some inlined logic, which is even harder to tree shake as shown in [this screenshot](https://twitter.com/naruaway/status/1687006117430525952) (a tweet in Japanese)

(BTW finding these was the main reason why I got interested in Valibot)

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
fabian-hiller: I understand your argument. However, I think that specifying 52 kB will cause more confusion than the gzip value. As a compromise, we could add "minified + gziped" to the comments in the docs.

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
MineLPPhynix: Why are they even using classes? I'm currently writing a custom translation library and I tried using classes but for every usecase objects just worked much better.

BTW, we could use something like the `addIssueToContext` function for removing the Exception performance penalty.
We also have to return some `INVALID` value, like Zod, to signal validation failure. I think this will make the code more complex and "ugly" but that should be ok for the performance benefit. Maybe there's an elegant solution?

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
naruaway: > we could add "minified + gziped" to the comments in the docs.

Sure, I think this is reasonable 👍

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
naruaway: > Why are they even using classes?

I have no idea! But probably partially because to make it easier to implement "method chain". Or maybe just coding style preference. I personally always avoid classes though 😅 

> BTW, we could use something like the addIssueToContext function

Yes, I guess we can discuss this in https://github.com/fabian-hiller/valibot/issues/48 ! Now I am confident that Valibot can be faster than Zod except for failure case (https://github.com/fabian-hiller/valibot/issues/73#issuecomment-1669292338)

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
MineLPPhynix: Im working on this File again and I want to improve this. I don't want to lie here, I want to be precise. Zod is 13.1 kB minified + gziped but im not sure if some of the classes will be treeshaken. What's the best way of calculating these numbers?

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
MineLPPhynix, line 1: I tried to rewrite it. Is this better? @naruaway

MineLPPhynix (commented):
I tried to rewrite it. Is this better? @naruaway

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
naruaway: It also depends on which bundler we use to bundle / minify (and gzip) and what to import.
I checked the following tools:

- https://pkg-size.dev/zod: "55KB minified (13KB gzip)"
- https://bundlephobia.com/package/zod@3.21.4 "57kB MINIFIED 13.1kB MINIFIED + GZIPPED"

To get more precise result, we can just build it with real source code by trusted production grade tools like Webpack.

I created [this StackBlitz](https://stackblitz.com/edit/stackblitz-starters-tvsaau?file=src%2Findex.js) and it shows gzipped one is **10.8 kB**. I used `import * as z from 'zod'` to get some tree-shaking.
Probably this is the most fair metrics for Zod. BTW in terms of notation, I think something like "10.8 kB gzip" is easy to understand ("minified" is redundant when we say "gzip" since there is no reason not to minify when doing gzip)

BTW I also checked [the size of Valibot using StckBlitz](https://stackblitz.com/edit/stackblitz-starters-yufn18?file=src%2Findex.js), which is **562 bytes gzipped**. Notice that using `import * as v from 'valibot'` is just fine

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
```diff
@@ -1,7 +1,101 @@
...
+After this setup you will have to change every **Zod Schema**,
+but you don't have to change all of them just yet.
+
+> **This is a Process, so take your time!** *You can always take a break, and I will still be here, waiting for you.*
+
+Theres usually two ways to write these Schemas with Zod:
+
+### The `z` Object
+
+> Zod recommends the use of it's `z` object export, which contains every schema class. Modern bundlers can remove ([treeshake](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking)) unused functions and objects, but this does not work with JavaScript classes. Zod Schemas and validation chains are build using class methods,  this is one of it's disadvantages compared to me. Learn more <Link href="/guides/comparison/">here</Link>.
```
naruaway, line 27: > Zod recommends the use of it's `z` object export,

Is it actually true? I do not see any reason since `import * as z from 'zod'` will produce slightly smaller bundle. Maybe we should not talk about Zod too much since this part can be also out of date quickly. I think the remaining sentences make sense to me but let me try to potentially improve:

`Tree-shaking does not work well with JavaScript classes. Zod schemas and validation chains are built using class methods, which is one of its disadvantages compared to me in terms of the final bundle size`

How about this?

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
```diff
@@ -1,7 +1,101 @@
...
+
+You need to import every Schema and <Link href="/guides/pipelines/">Pipeline</Link> method you are using. Then you can just remove all references of the `z` object, and many of the Schema functions should be compatible.
```
naruaway, line 29: This might be confusing since `import * as v from 'valibot'` just works. Note sure about which character is the "official" one though. In my personal projects I am using this form now

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
```diff
@@ -1,7 +1,101 @@
...
+
+*That's it! Pretty easy, right?*
+
+## Special Examples / Equivalents
+
+Here are some example usecases, that are used different than in Zod.
+
+| Zod     | Valibot |
+| ----------- | ---------------- |
+| `z.string().default("hello")` | `useDefault(string(), 'hello') ` |
```
naruaway, line 100: Using `useDefault` in examples might not be a good idea since we need to change the name https://github.com/fabian-hiller/valibot/issues/80

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
```diff
@@ -1,7 +1,101 @@
 ---
 title: Migrate from Zod
+description: Im based on Zod, so migrating should not require much work, because the naming and behaviour of methods is very similar.
```
naruaway, line 3: typo? Im -> I'm
But I am not sure whether it's true that "Valibot is based on Zod" 🤔 I am sure it's inspired by Zod at least though

fabian-hiller:
Sorry that I have not yet had time to check the PR. Unfortunately I have a lot to do at the moment. Therefore, do not expect a quick reaction.

MineLPPhynix:
> Sorry that I have not yet had time to check the PR. Unfortunately I have a lot to do at the moment. Therefore, do not expect a quick reaction.

No problem, we're still working on some parts of the text.

website/src/routes/guides/(migration)/migrate-from-zod/index.mdx
```diff
@@ -1,7 +1,101 @@
...
+First you need to install me. See <Link href="/guides/installation/">Installation instructions</Link>.
+
+## Schema Migration
+
+After this setup you will have to change every **Zod Schema**,
+but you don't have to change all of them just yet.
+
+> **This is a Process, so take your time!** *You can always take a break, and I will still be here, waiting for you.*
+
+Theres usually two ways to write these Schemas with Zod:
```
estubmo, line 23: There's

fabian-hiller:
Is the guide final on your part? Should I review it in the next week and give feedback?

FlorianDevPhynix:
> Is the guide final on your part? Should I review it in the next week and give feedback?

Yes, feedback would be great.

fabian-hiller:
It would be perfect if we had a tool that could rewrite a Zod schema into a Valibot schema: https://sinclairzx81.github.io/typebox-workbench/

fabian-hiller:
As soon as I find the time, I will revise this guide and then merge the PR.

fabian-hiller:
Sorry, but I probably won't get around to reviewing this PR and finalizing the guide until January.

FlorianDevPhynix:
> Sorry, but I probably won't get around to reviewing this PR and finalizing the guide until January.

no worries

fabian-hiller:
Hey everyone. I want to apologize. I had some different ideas and rewrote the Zod migration guide from scratch. Feel free to take a look and give me feedback. https://valibot.dev/guides/migrate-from-zod/

FlorianDevPhynix:
Looks great! Your list with name changes is much better than the list of changes that I added. People can just click on the links to the documentation and see how each works, no need to explain all of them on this page.
I can close this, don't think it is needed anymore.

Issue #56, opened 2023-08-04
feature: add branded types


zkulbeda:
Introduce `brand` method
```ts
const BrandType: unique symbol = Symbol()

interface Brand<K extends string | symbol> {
  readonly [BrandType]: {
    readonly [k in K]: K
  }
}

type Branded<T, K extends string | symbol> = T & Brand<K>

function brand<TSchema extends BaseSchema, const K>(
  schema: TSchema,
  brand: K
): BaseSchema<Input<TSchema>, Branded<Output<TSchema>, K>>{
  return schema
}

const schema = brand(number([maxValue(20)]), "UserId");
parse(schema, 123) // number & Brand<"UserId">

fabian-hiller:
Can you explain in which specific use cases `brand` is useful and offers an advantage? I have not quite understood it yet.

fabian-hiller:
I think I understand it now. Cool idea. Once I'm done with my bachelor thesis and hopefully have some more time, I can take a closer look at it.

fabian-hiller:
If you have time and interest, you can try to implement this feature and submit a PR. I will check your code and give you feedback.

fabian-hiller:
The `brand` feature is now available 🎉

danielo515:
@fabian-hiller if you have finished your degree and you are not yet familiar with thi, then you are missing one of the best type safety features. This exists in several languages with different names, but the same concept:
- ocaml: phantom types
- rust: newtype pattern 
- F# constrained types (I think) 

Since you seem to be interested in the type safety topic I really recommend taking a look at "domain modeling made functional", for some eye-opening usage examples. For me they are super valuable for IDs (what is that string? A user ID or a car ID?)

fabian-hiller:
Thank you. I will have a look at it.

PR #85, opened 2023-08-10
Add brand type

mrsekut:
## Introduction

This PR introduces the concept of "branded types" to the `valibot` library, a feature inspired by the [`.brand` in `zod`](https://github.com/colinhacks/zod#brand).

## Motivation

TypeScript is structurally typed, which means it doesn't distinguish between types with the same structure. This can be problematic in scenarios where we want to differentiate between types that have the same underlying structure but represent different concepts.

For instance, consider a project where we're working with multiple units of length, like millimeters (mm) and centimeters (cm). Both would typically be represented as `number` in TypeScript. However, with the introduction of branded types, we can now differentiate between them:

```typescript
const MM = brand(number(), 'MM');
const CM = brand(number(), 'CM');
```

## Implementation Details

While designing the interface for this feature, I considered an approach like `brand<'MM'>(number())`. However, due to TypeScript's nature, when there are multiple generic type arguments, we can't specify just one. Hence, the current implementation was chosen. This is an area where I believe there's room for discussion, especially considering usability.

It's important to note that the `brand` function doesn't alter the value in any way. Therefore, the result of the `parse` function remains unchanged before and after using `brand`. Consequently, the test cases added in this PR focus on type-level checks.

## On the Naming: "brand"

The term "brand" seems to be frequently used within the TypeScript community. In fact, it even appears in TypeScript's own codebase. You can find a reference to it [here](https://github.com/Microsoft/TypeScript/blob/7b48a182c05ea4dea81bab73ecbbe9e013a79e99/src/compiler/types.ts#L693-L698).

While there are similar terms like "opaque" that could be used, aligning with `zod` and using "brand" seems appropriate and consistent.

## Personal Note

I've been using `zod` in my projects and heavily relied on the `brand` feature. I felt the need for this feature in `valibot` as I'm transitioning to it.


fabian-hiller:
I think that `brand` is a method and not a schema or? Can you move it to the methods? This PR fixes #56.

mrsekut:
Thank you for the feedback.

I understood your point and have moved `brand` to the `methods/` directory as suggested.

I've added the changes in a subsequent commit. If there were any issues, please feel free to squash the commits during the merge. If necessary, I could also rebase and force push the changes.

fabian-hiller:
Thank you! I will check and merge your code tonight or in the coming days.

fabian-hiller:
I am about to release a new version that will include the `brand` method. Let me know then if it works in practice for you as expected. Thanks for your contribution!

PR #88, opened 2023-08-11
Allow multiple brands and add Unbranded utility

### LINKED ISSUE #87: Multiple brands schema results to never
Hi! Current implementation of branded types uses single symbol for single branded type, that results to never type in tests

https://github.com/fabian-hiller/valibot/blob/ce49b0b4809a62035767c4b7697fabddd871392c/library/src/methods/brand/brand.test.ts#L32-L38

image

It happends because typescript resolves `{ [symbol]: "1"} & { [symbol]: "2" }` to `never`. 

I'll make a pr for fix, that replaces single value to object of values

zkulbeda:
fixes #87

I found `UnionToIntersection` type in keyof.ts and decided to just copy it in brand.ts. Should it imports type for code deduplication?


library/src/methods/brand/brand.ts
```diff
@@ -84,9 +86,46 @@ type BrandName = string | number | symbol;
  * Brand type.
  */
 export type Brand<TBrandName extends BrandName> = {
-  [symbol]: TBrandName;
+  [BrandSymbol]: {
+    [N in TBrandName]: N;
```
fabian-hiller, line 90: Can you explain what this line does and why it is required?

library/src/methods/brand/brand.ts
```diff
@@ -633,6 +672,20 @@ export function brand<
 /**
  * Brands the output type of a schema.
  *
+ * @example
```
fabian-hiller, line 675: To keep the source code as simple as possible, I prefer not to add examples. I will add examples in the future to the API reference on the website.

library/src/methods/brand/brand.test.ts
```diff
@@ -31,9 +32,51 @@ describe('brand', () => {
...
+
+    expectTypeOf<BrandedId>().not.toBeNever();
+    expectTypeOf<BrandedUserId>().not.toBeNever();
+    expectTypeOf<BrandedPostId>().not.toBeNever();
+
+    expectTypeOf<BrandedUserId>().not.toMatchTypeOf<BrandedPostId>();
+    expectTypeOf<BrandedPostId>().not.toMatchTypeOf<BrandedUserId>();
+  });
+
+  test('should support unbranding', () => {
```
fabian-hiller, line 65: Why is unbranding and this test necessary? Is there a practical use case for this?

library/src/methods/brand/brand.ts
```diff
@@ -84,9 +86,46 @@ type BrandName = string | number | symbol;
...
+      }[keyof BrandedType[BrandSymbol]]
+    >
+  : never;
+
+/**
+ * Removes all brands from a branded type or returns the type if it is not branded.
+ * @example
+ * Unbranded<string & Brand<'foo'> & Brand<'bar'>> = string
+ */
+export type Unbranded<BrandedType> = BrandedType extends infer T &
```
fabian-hiller, line 124: If the newly added types are only needed in the tests, they should not be in the `brand.ts` file, but directly under `brand.test.ts`. Or alternatively under `types.ts` in the `brand` folder if there is a practical use case for it outside the library code.

library/src/methods/brand/brand.test.ts
fabian-hiller: Can you explain the practical use case? Does Zod have this feature?

fabian-hiller:
I will review the PR next week.

fabian-hiller:
I have now checked your code. We can extend the `Brand` type so that `brand` can be applied multiple times. However, currently I would not add `Brands` and `Unbranded` as it complicates Valibot's source code and testing and I don't quite understand the benefit yet. I have not found any issue or PR at Zod in this regard. So I assume that this feature is rarely needed. If you disagree here, please let me know.

PR #93, opened 2023-08-13
Reduce spread operators for perf round 3: new ValiError()

fabian-hiller:
In my test environment, this change had no effect. @naruaway can you test if it makes a difference in your benchmarks? If it doesn't make a difference in yours either, we should find out why that is the case.

### DIFF
--- a/library/CHANGELOG.md
+++ b/library/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to the library will be documented in this file.
 - Refactor loops and array destructuring for performance reasons
 - Add `getPipeInfo` util and refactor code for performance reasons (pull request #46)
 - Add `getPathInfo` util and refactor code for performance reasons (pull request #92)
+- Add `getIssue` util and refactor code for performance reasons (pull request #93)
 - Rename `getCurrentPath` util to `getPath` (pull request #92)
 
 ## v0.12.0 (August 11, 2023)
--- a/library/src/methods/strict/strict.ts
+++ b/library/src/methods/strict/strict.ts
@@ -1,5 +1,6 @@
 import { ValiError } from '../../error/index.ts';
 import type { ObjectSchema } from '../../schemas/object/index.ts';
+import { getIssue } from '../../utils/index.ts';
 
 /**
  * Creates a strict object schema that throws an error if an input contains
@@ -32,14 +33,12 @@ export function strict<TSchema extends ObjectSchema<any>>(
       // Check length of input and output keys
       if (Object.keys(input as object).length !== Object.keys(output).length) {
         throw new ValiError([
-          {
+          getIssue(info, {
             reason: 'object',
             validation: 'strict',
-            origin: 'value',
             message: error || 'Invalid keys',
             input,
-            ...info,
-          },
+          }),
         ]);
       }
 
--- a/library/src/methods/strict/strictAsync.ts
+++ b/library/src/methods/strict/strictAsync.ts
@@ -1,6 +1,7 @@
 import { ValiError } from '../../error/index.ts';
 import type { ObjectSchemaAsync } from '../../schemas/object/index.ts';
 import type { Output } from '../../types.ts';
+import { getIssue } from '../../utils/index.ts';
 
 /**
  * Creates a strict async object schema that throws an error if an input
@@ -33,14 +34,12 @@ export function strictAsync<TSchema extends ObjectSchemaAsync<any>>(
       // Check length of input and output keys
       if (Object.keys(input as object).length !== Object.keys(output).length) {
         throw new ValiError([
-          {
+          getIssue(info, {
             reason: 'object',
             validation: 'strict',
-            origin: 'value',
             message: error || 'Invalid keys',
             input,
-            ...info,
-          },
+          }),
         ]);
       }
 
--- a/library/src/schemas/array/array.ts
+++ b/library/src/schemas/array/array.ts
@@ -3,6 +3,7 @@ import type { BaseSchema, Input, Output, Pipe } from '../../types.ts';
 import {
   executePipe,
   getErrorAndPipe,
+  getIssue,
   getPath,
   getPathInfo,
   getPipeInfo,
@@ -94,14 +95,12 @@ export function array<TArrayItem extends BaseSchema>(
       // Check type of input
       if (!Array.isArray(input)) {
         throw new ValiError([
-          {
+          getIssue(info, {
             reason: 'type',
             validation: 'array',
-            origin: 'value',
             message: error || 'Invalid type',
             input,
-            ...info,
-          },
+          }),
         ]);
       }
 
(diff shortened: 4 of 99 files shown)


naruaway:
(on mobile, do not have PC now) 'throw' it self can be too slow compared to the spread ops

I'll check this in later today as well

fabian-hiller:
`throw` did not have a negative effect in my tests. Validating invalid data is almost as fast as with valid data.

naruaway:
@fabian-hiller Ok I ran https://github.com/naruaway/valibot-benchmarks to compare 29f78b4b854dbab9168a2636c287bf4ed46a506c against 27278fe99be29dfa96fe950b126ad7fca54218d5 and it looks like the result is consistent with your observation. 
<details><summary>Benchmark result of 29f78b4b854dbab9168a2636c287bf4ed46a506c against 27278fe99be29dfa96fe950b126ad7fca54218d5</summary>
<pre>
Running benchmarks using nodejs
Showing the result when running runtime environment (e.g. Node.js) 100 times for each:
case: "deep", data: "valid"
 0% slower
  baseline: 357921 ops/s, +- 3% stddev
  target: 356630 ops/s, +- 4% stddev
case: "many_features", data: "invalid"
 4% slower
  baseline: 18500 ops/s, +- 2% stddev
  target: 17839 ops/s, +- 2% stddev
case: "many_features", data: "valid"
 1% faster
  baseline: 104606 ops/s, +- 5% stddev
  target: 105324 ops/s, +- 4% stddev
case: "optional_nullable", data: "valid"
 0% slower
  baseline: 374669 ops/s, +- 5% stddev
  target: 374159 ops/s, +- 4% stddev
case: "wide", data: "invalid"
 1% faster
  baseline: 4021 ops/s, +- 6% stddev
  target: 4081 ops/s, +- 4% stddev
case: "wide", data: "valid"
 0% faster
  baseline: 235519 ops/s, +- 4% stddev
  target: 236182 ops/s, +- 5% stddev
Running benchmarks using bun
Showing the result when running runtime environment (e.g. Node.js) 100 times for each:
case: "deep", data: "valid"
 0% slower
  baseline: 389875 ops/s, +- 3% stddev
  target: 389212 ops/s, +- 2% stddev
case: "many_features", data: "invalid"
 3% faster
  baseline: 85253 ops/s, +- 5% stddev
  target: 87828 ops/s, +- 3% stddev
case: "many_features", data: "valid"
 3% faster
  baseline: 406345 ops/s, +- 4% stddev
  target: 417007 ops/s, +- 4% stddev
case: "optional_nullable", data: "valid"
 0% faster
  baseline: 443788 ops/s, +- 7% stddev
  target: 444826 ops/s, +- 7% stddev
case: "wide", data: "invalid"
 3% faster
  baseline: 22390 ops/s, +- 3% stddev
  target: 23113 ops/s, +- 3% stddev
case: "wide", data: "valid"
 1% faster
  baseline: 287371 ops/s, +- 8% stddev
  target: 290755 ops/s, +- 8% stddev
out of 12 cases, 8 cases (67%) were faster
</pre>
</details>


> Validating invalid data is almost as fast as with valid data

Which schema / data did you use? In `many_features` schema in my benchmark, parsing "invalid" data is 82% slower than parsing "valid" data. This schema and data is [stored here](https://github.com/naruaway/valibot-benchmarks/tree/ac99da10597c9b4c98e20a4d1a5df74aba663896/test_data/many_features)

fabian-hiller:
> Validating invalid data is almost as fast as with valid data.

I had forgotten to replace the test data with invalid data. I noticed it just now. Now validating erroneous data is extremely much slower than valid data. You are probably right that `throw` could influence this. I will investigate this today.

fabian-hiller:
With this PR, I see now an improvement from 8 seconds to 6 seconds in my test.

fabian-hiller:
The investigation with a performance profiler shows that the instantiation and throwing of the `ValiError` is the biggest bottleneck. It probably relates to the parent class `Error`, which collects error information in the background. If I remove the parent class, I see a 30% improvement in performance. You are right that if we change the error handling to a context, we will see a much better performance.

fabian-hiller:
My idea, based on your suggestion, is the following. Instead of throwing errors in the `parse` or `safeParse` method, we add the key `issues` to the `info` object with an empty array. Now instead of throwing errors, they are added to the `issues` array, and only at the very end in the `parse` or `safeParse` method a `ValiError` is created. With these changes, we could see much better performance.

naruaway:
@fabian-hiller Yeah I had kind of similar idea, but there is one issue. Since `.parse` still need to throw to retain backward compatibility. Composite schema such as `object` will be still slow (e.g. [this try-catch](https://github.com/fabian-hiller/valibot/blob/29f78b4b854dbab9168a2636c287bf4ed46a506c/library/src/schemas/object/object.ts#L118-L142) will be executed many times for ["wide" type of schema like this](https://github.com/naruaway/valibot-benchmarks/tree/ac99da10597c9b4c98e20a4d1a5df74aba663896/test_data/wide)). If we really need to avoid throwing, probably we need to switch some sort of "mode" for .parse to avoid throwing or we could implement alternative method for internal usage or we could turn `.safeParse` first class instead. Anyway, at this point I feel like it can be better to encourage people to use `v.parse(v.string(), x)` rather than `v.string().parse(x)` so that we can change internal API

fabian-hiller:
Yes, `.parse` is an internal API and should not be used externally. In the docs, I have taken this into account. In my opinion, we don't need to be backward compatible as we are still in v0. Also, fixing this with `parse(schema, data)` should be quick. Do you have an alternative idea for naming `.parse`? Should we use an underscore to indicate that it is an internal API? My idea from above includes that `.parse` no longer throws an error but fills the `.issues` array.

naruaway:
> Yes, .parse is an internal API and should not be used externally. In the docs, I have taken this into account. In my opinion, we don't need to be backward compatible as we are still in v0. 

Oh I see! I missed this and I was just using `v.string().parse()` a lot 😅  Yes I think we can rename it to `._parse()` and we should not expose this in TypeScript definition. (we might come up with hard protection like [hard private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) such as `#parse` but I really do not like this. We should still allow people to access this internal API if they really want and they know what they are doing )

> we are still in v0

definitely 👍 we should not be afraid of good breaking changes

fabian-hiller:
I have already started the improvements and will create a draft PR in an hour or two.

fabian-hiller:
It took longer than I thought. But now I am finished. Next, I have to revise all the tests. That will probably take another 1 to 2 hours.

fabian-hiller:
The first tests show that the validation time for invalid data could be reduced by up to 90% by changing the error handling from `try...catch` and `throw new ValiError(...)` to an `issues` array.

fabian-hiller:
I have just added the changes to the `main` branch. Feel free to review the code and give me feedback. However, before we release a new version, we need to ensure that we don't break any existing integrations, e.g. with tRPC and React Hook Form.

PR #99, opened 2023-08-13
Initial support of vitest benchmark

zkulbeda:
This separated pr is required to merge into main branch for testing github actions workflow in next coming pr #100. I just copied one of the tests from typescript-runtime-type-benchmark just for demo purposes.

Is `benchmarks` folder in ./library/src good place for storing benchmarks files?


fabian-hiller:
Thank you very much! The next few days I have to focus on my bachelor thesis. As soon as I find time, I'll get back to you.

fabian-hiller:
Do you have any idea how we can make the result comparable to the latest version of Valibot, so we can see if changes have a positive or negative impact on the performance?

fabian-hiller:
We could add the latest version of Valibot as a dev dependency and run the benchmark for both to have a comparison of the codebase to the latest Valibot version.

fabian-hiller:
Yes, I will try to check it in the next few days.

fabian-hiller:
Sorry for not coming back to this, but other things had a higher priority. I am also not sure if we really need it, since the basic structure of the library is set and changes will most likely not have a major impact on performance. It is still useful, but other things have a higher priority for me.

fabian-hiller:
I think that we will add benchmark tests in the long run but right now it seems not too important. I will close this PR to cleanup the repo before our v1 release.

Issue #97, opened 2023-08-13
Allow boolean literal


zkulbeda:
Currently you can't create true or false as const in valibot because function generic argument must extend string or number

https://github.com/fabian-hiller/valibot/blob/ea4e6f39dce43cfc49eb542b1b200fb5a904b1ae/library/src/schemas/literal/literal.ts#L24-L27

![image](https://github.com/fabian-hiller/valibot/assets/24959171/e16f37b0-2e5d-491c-af7a-b0452c1db31f)

fabian-hiller:
Thanks for the hint! I will improve that.

PR #102, opened 2023-08-14
Primitives values in literal schema

fvckDesa:
This PR add all primitives values on literal schema.

Solve Issue #97

### DIFF
--- a/library/CHANGELOG.md
+++ b/library/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to the library will be documented in this file.
 > Note: The library has been revised and refactored. There is a migration guide in the [release notes](https://github.com/fabian-hiller/valibot/releases/tag/vX.X.X).
 
 - Add `getPipeInfo`, `getPathInfo` and `getIssue` util (pull request #46, #92, #93)
+- Add support for `boolean`, `bigint` and `symbol` to `literal` (pull request #102)
 - Expand `flatten` so that issues are also accepted as first argument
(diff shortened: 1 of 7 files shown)


fabian-hiller:
Great! Thanks a lot! I'll check your code as soon as I find time and publish this improvement with the next version.

kassiomaia (approved):

kassiomaia (approved):
I just do not sure if null should be a primitive considering that null extends from object!

fabian-hiller:
I think `null` and `undefined` should be removed because a literal with this datatype makes no sense. This would also have the consequence that the type `Primitive` should be renamed to `Literal`. The tests would then have to be updated. Other than that, the code looks good.

fvckDesa:
I have take the primitives data types from [MDN Documentation](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) that includes null and undefined between primitives values.
I think that null and undefined is not very usefull as literal value, but I don't see why someone shouldn't have the possibility to use them

fabian-hiller:
Since this does not change the schema's implementation, it is not so important to me whether we include `null` and `undefined`. Both have their advantages and disadvantages. If we leave it like it is, `literal` is more flexible to use on the one hand. However, the API of Valibot becomes more confusing, on the other hand, as there are now two ways to achieve the same thing. Also, strictly speaking, there is no `null` and `undefined` literal since these datatypes cannot be further specified.

Feel free to give me feedback on this again. I'll probably decide tomorrow how I want to proceed.

Issue #74, opened 2023-08-07
Catch/Fallback


FlorianDevPhynix:
zod's [`default()`](https://zod.dev/?id=default) only works with undefined and not null, which this library perfectly mimics. [Zod Sandbox Example](https://stackblitz.com/edit/node-zod-default-catch?file=index.ts)

Thanks to `catch()` on the other hand, it is possible to have a default value when validation fails. But Valibot seems to not have [`catch()`](https://zod.dev/?id=catch). 
It is one of the most useful functions of zod (atleast to me) and I use it a lot. Especially with react, where I want the ensure the app always has certain pieces of data.

Thanks to the simplicity of Valibot is was able to quickly write a catch method.

``` ts
import type { BaseSchema, Output } from 'valibot';
import { ValiError } from 'valibot';

function Catch<TSchema extends BaseSchema>(
  schema: TSchema,
  value: Output<TSchema>,
  callback?: (err: Error) => void
): TSchema {
  return {
    ...schema,
    parse(input, info) {
      try {
        return schema.parse(input, info);
      } catch (err: any) {
        if (typeof callback === 'undefined') {
          console.error(err);
        } else {
          callback(err);
        }
        return value;
      }
    },
  };
}
```
[Sandbox](https://stackblitz.com/edit/valibot-node?file=fallback.ts) with usage examples
`npm run fallback`

This method could also be called fallback. I originally called it that, but to stay true to zod's naming I changed it to Catch, just not catch (reserved word).

fabian-hiller:
Thank you very much for the hint. I will add this functionality as soon as possible. I think `fallback` is a good alternative name for `catch`.

MineLPPhynix:
I read some of the performance issues and the proposed changes to move away from error throwing to some form of context being passed around, might make this function more complex. Should still work very similar though, if the previous validation issues are not passed down, only up. Catching those would be almost the same.

PR #103, opened 2023-08-14
Fallback method

FlorianDevPhynix:
Now that exceptions are gone, I can finally, properly implement the `fallback` method. #74 
Also added some tests.

Still not sure about the issues that are caught though. Adding a centralized way to collect these warning, just for fallback, is unnecessary. Printing or being able to receive them through the callback might work for now, but I did not try collecting them.


FlorianDevPhynix:
I added a test where a function is used to collect all the warnings into a list. This could also be used as an example in documentation.

FlorianDevPhynix:
The example is probably too big for the method guide page. Maybe it should be added to its api page, but the api reference has no written pages, so I'm not sure how those are supposed to look.

fabian-hiller:
Great! Thanks a lot! I'll check your code as soon as I find time and publish this improvement with the next version.

PR #104, opened 2023-08-15
Perf improvements

jussisaurio:
Hey,

Recently popped into [this issue thread](https://github.com/fabian-hiller/valibot/issues/73) to discuss a fairly significant perf improvement idea, but found out @fabian-hiller had already done almost exactly the same thing!

However, I just finished a 4 hour plane ride and I spent the whole flight looking for more perf improvements and found a few! 

**Sizeable optimizations**

Both of these generally just involved doing less whenever possible:

1. Evaluate full issue paths lazily instead of eagerly
2. Avoid creating PipeInfo objects when `pipe.length === 0`

I'd say these contributed 90% to the total performance improvement in this PR.

**Micro-optimizations**

1. Make `cachedEntries` for the `object` schema a flat array instead of an array of 2-tuples -- mainly just less indirection which is more friendly to the CPU.
2. Simpler "is object" check using `input?.constructor === Object`

Here are the results on my machine using [typescript-runtime-type-benchmarks]():

```
*** valibot @ current head of main, i.e. e6c53c8 ***
Running "parseSafe" suite...
Progress: 100%

  valibot:
    1 784 335 ops/s, ±1.51%   | fastest

Finished 1 case!
Running "parseStrict" suite...
Progress: 100%

  valibot:
    1 715 007 ops/s, ±0.15%   | fastest

-------------------------------------------

*** THIS PR ***
Running "parseSafe" suite...
Progress: 100%

  valibot:
    2 616 009 ops/s, ±0.21%   | fastest

Finished 1 case!
Running "parseStrict" suite...
Progress: 100%

  valibot:
    2 317 246 ops/s, ±0.25%   | fastest
```

This run amounts to +47% on `parseSafe`, +35% on `parseStrict`.


fabian-hiller:
Thank you very much for your contribution! At the moment I have to focus on the text of my bachelor thesis. As soon as I have more time, I will check and merge your changes.

milankinen:
Just a quick notice about path if we're really trying to optimize the performance. Usually path is only needed in case of failures so that the path can be added to the issues. If this is the case, it can be allocated lazily only if the validation (or sub-validation) returns any issues instead of the [current implementation where the path is allocated eagerly](https://github.com/jussisaurio/valibot/blob/perf-improvements/library/src/schemas/object/object.ts#L136-L141). This should yield some small performance improvements due to decreased number of allocations and gc overhead. 

If the failure types are something like 
```ts
type Issues = { 
  pathItem: PathItem
  issues: (IssueErrorMessageAndMeta | Issues[])[] 
}
````
Then the `_parse` function validation logic in failure cases becomes extremely simple:
```ts
// object validation pseudo
let issues: Issues[] | undefined
for each key in objectSchema {
  const result = parseKeySchema(object[key])
  if (result.issues) {
   (issues = issues ?? []).push(result.issues)
  }
}
return issues ? { valid: false, issues } : { valid: true, value }
```

Of course the issues structure (and paths along with it) must be flattened at the top level parse before returning the issues to the library user, but that should be very straightforward task to do (similar to the current `getPath`).

That said, I'm not familiar with this library or its features so take this with grain of salt... 😄

fabian-hiller:
Yes, that is correct. Thank you very much for pointing it out. I will take it into account for the final implementation.

jussisaurio:
Yep that's a very good point, interested to see how much this speeds things up further

fabian-hiller:
Thank you for your contribution! Due to my bachelor thesis I have a lot to do at the moment. I'll try to check the changes next week at the latest.

jussisaurio:
> Thank you for your contribution! Due to my bachelor thesis I have a lot to do at the moment. I'll try to check the changes next week at the latest.

Sure, no pressure nor hurry, it's your library dude :) I just rebased and fixed the conflicts, nothing else.

If I have extra time I can try to find time to implement @milankinen 's suggested refactor as well, but it's kind of 50/50 whether I can find the time

jussisaurio:
Found an hour of idle time and started shoveling, looking pretty promising... :D

```
Removing previous results
Executing "valibot"
Loading "valibot"
Running "parseSafe" suite...
Progress: 100%

  valibot:
    3 657 558 ops/s, ±0.11%   | fastest

Finished 1 case!
Running "parseStrict" suite...
Progress: 100%

  valibot:
    3 213 696 ops/s, ±0.10%   | fastest
```

@milankinen you absolute GOAT! 🐐 

I'm not quite finished with the refactor yet - must make sure all the tests pass and none of the public API is broken.

fabian-hiller:
Can you explain why it is necessary to make a distinction between `LeafIssue`, `NestedIssue` and `UnionIssue` for the issues?

jussisaurio:
> Can you explain why it is necessary to make a distinction between `LeafIssue`, `NestedIssue` and `UnionIssue` for the issues?

Sure! We can obviously iterate on the data model, but the idea in a nutshell is to accurately model the different kinds of objects according to the information that is actually relevant to each:

`LeafIssue` is an issue with a "primitive": the input was supposed to be a string, but was a number instead, or was supposed to be an object, but was a string instead (important to distinguish this from other issues with objects -- here the specific problem is that the input is _not an object at all_). So for these types of issues it makes sense to include `input` , `message` etc, but it does not include other issues, so they shouldn't be in its datatype. The primitive also doesn't really have to be aware of `path` -- a string is a string, the path is its containing parent's concern.

---

`NestedIssue` is an issue with a "composite": an input may be a valid object (i.e. it is not a string), but it can have e.g. the following problem: it should have contained key `a` of `string`, but `a` was a `number` instead. It's still a valid object, but something within its composition is wrong. More precisely, one or more sub-`path`s within this composite contained issues. Hence, it makes sense to include `path` and a list of `issues` from each subpath. Likewise, for maps and records, `origin` belongs here, but nowhere else, since `origin` is nearly always `value`. 

- Looking at the current implementation of `NestedIssue`, it should probably still have the `validation` property which I neglected to add. But does it need to have `reason`? The reason in the above example is that one of its child properties was not a string. So I'm thinking the `reason` belongs to the child, not the container.

---

`UnionIssue` was something that I identified as a special case: It's basically a "pathless" NestedIssue. Not 100% sure what to do with it. The current data model for it may well not be the right one.

milankinen:
Wow nicely done @jussisaurio 😮 👌!

>Can you explain why it is necessary to make a distinction between LeafIssue, NestedIssue and UnionIssue for the issues?

Sorry my fault, separation was just my initial thought because of separation of concerns. 😅  It's not absolutely necessary to have different types of issues if the issues are flattened "in-place" in schemas. E.g. something like:

```ts
const createIssue = (msg: string) => ({msg, path: []})
const appendIssues = (issues: Issue[], childIssues: Issue[], pathItem: PathItem) => {
  issues.push(...childIssues.map(issue => issue.path.unshift(pathItem) && issue))
  return issues
}

// string.ts
if (typeof value !== 'string') {
  return createIssue('not a string')
}

// object.ts
if (result.issues) {
  issues = appendIssues(issues, result.issues, key)
}
```

fabian-hiller:
I don't have time to think about it in detail at the moment. However, I want to keep the library as simple as possible, as this also reduces the bundle size, and splitting it into different issues seems a bit complicated. It is already possible to extract all important information with the current implementation, isn't it?

A path is still needed for a `LeafIssue` if it is in an object. Or can a `LeafIssue` not concern nested data? If so, the name is not optimal.

The only thing that would make sense to me would be:
- `RootIssue`: Issue without path
- `NestedIssue`: Issue with path
- `UnionIssue` (special case): With sub-issues (stems from `RootIssue` or `NestedIssue`)

With the current implementation, the path is only constructed once an issue occurs, which is great. Does this lead to bugs with deeply nested issues? We may construct the path retroactively e.g. with `path.unshift(…)`? Then `LazyPath` would not be necessary, and the code would be simpler constructed or?

fabian-hiller:
Tomorrow or Sunday, I will try to release a new version with all performance improvements. For time reasons (because of my bachelor thesis), I will then weigh whether I build on this issue or implement the ideas independently for the time being. I appreciate your contribution and will mention you in the release notes. 🤝

jussisaurio:
@fabian-hiller you are free to do what you like ofc :) the important point is the lazy construction of paths.

as you can see in the most recent commit i removed the LazyPath class entirely because its not needed at all. ParseInfo no longer contains path because it's only constructed when an issue actually arises. that is the key detail.

wish you the best with your thesis, those are a lot of work!

jussisaurio:
> I don't have time to think about it in detail at the moment. However, I want to keep the library as simple as possible, as this also reduces the bundle size, and splitting it into different issues seems a bit complicated. It is already possible to extract all important information with the current implementation, isn't it?
> 
> A path is still needed for a `LeafIssue` if it is in an object. Or can a `LeafIssue` not concern nested data? If so, the name is not optimal.
> 
> The only thing that would make sense to me would be:
> 
> * `RootIssue`: Issue without path
> * `NestedIssue`: Issue with path
> * `UnionIssue` (special case): With sub-issues (stems from `RootIssue` or `NestedIssue`)
> 
> With the current implementation, the path is only constructed once an issue occurs, which is great. Does this lead to bugs with deeply nested issues? We may construct the path retroactively e.g. with `path.unshift(…)`? Then `LazyPath` would not be necessary, and the code would be simpler constructed or?

And yes in the most recent commit the path is constructed retroactively in the flatten() function. In non flattened issues the path is only the key or the index, because nested issues have an issues array, which may have their own issues array, etc. So you can always build the full path string from those as necessary. Again, this was just my shot at the implementation, but it does avoid doing any extra work on the paths before an issue actually happens.

If you want you can play around with it in a branch to see how it works. But anyway, do whatever you think is best!

fabian-hiller:
Thank you both for the research and ideas. Since I wanted to do a few things differently, I implemented it on my own and added you guys as co-authors. Feel free to check the result. You may see something we can do better.

The paths are now created lazy, and the pipeline is only executed when required. Also, I changed the object check to `input.constructor !== Object`.

However, I kept the previous implementation of `cachedEntries` because there was hardly any difference in performance in my test environment, and the code became more confusing with the new implementation.

The changes helped to improve performance by about 30% in my test environment. That's great! Thanks a lot!

jussisaurio:
Hey @fabian-hiller , nicely done! 🥳 

I made a simpler version of the flat cachedEntries [here](https://github.com/fabian-hiller/valibot/compare/main...jussisaurio:valibot:cachedentries-flat), trying to be a bit more sensitive towards eventual bundle size and readability. It still gets a tad confusing in the `objectAsync` case.

Perf-wise: I'm still consistently getting a 6% perf boost using this strategy on typescript-runtime-type-benchmarks compared to current `main`, so feel free to throw it in at your discretion if you consider it a net win.

fabian-hiller:
Thanks for the hint. For `objectAsync` I would probably leave it as is, but for `object` I can imagine the change. @naruaway can you check in your benchmark suite how the change affects performance?

naruaway:
I guess many of (similar) changes in this PR are already incorpolated in the latest `main`. 
So I compared `Make cachedEntries for the object schema a flat array instead of an array of 2-tuples -- mainly just less indirection which is more friendly to the CPU.` part (I extracted out this part on top of the current main: https://github.com/naruaway/valibot/pull/1) against the current `main`.

And I ran the benchmark suite for the following commits on my MacBook Air (M2 chip):

- `latest-main`: [00d234f4438250349d71d2cf154bef38649bc769](https://github.com/fabian-hiller/valibot/commit/00d234f4438250349d71d2cf154bef38649bc769) current `main` branch
- `use-flat-object-cache`: [bad8b32ed0d1bcab69cc901e6cbd82d7fb08a5ca](https://github.com/naruaway/valibot/commit/bad8b32ed0d1bcab69cc901e6cbd82d7fb08a5ca) I tried using speread operator on top of the `main`: https://github.com/naruaway/valibot/pull/1

<details>
<summary>Benchmark results</summary>
<pre><code>
nodejs, deep, valid
  use-flat-object-cache : 1095916.5 ops/s (fastest, 5.2% faster)
  latest-main           : 1042200.5 ops/s

nodejs, many_features, invalid
  use-flat-object-cache : 294343.5 ops/s (fastest, 1.2% faster)
  latest-main           : 290939.5 ops/s

nodejs, many_features, valid
  use-flat-object-cache : 1183446 ops/s (fastest, 3.8% faster)
  latest-main           : 1140184 ops/s

nodejs, optional_nullable, valid
  use-flat-object-cache : 798971.5 ops/s (fastest, 5.1% faster)
  latest-main           : 760427.5 ops/s

nodejs, wide, invalid
  use-flat-object-cache : 175466.5 ops/s (fastest, 2.2% faster)
  latest-main           : 171669 ops/s

nodejs, wide, valid
  use-flat-object-cache : 369301 ops/s (fastest, 5.6% faster)
  latest-main           : 349862 ops/s

bun, deep, valid
  use-flat-object-cache : 1814776 ops/s (fastest, 9.8% faster)
  latest-main           : 1652201.5 ops/s

bun, many_features, invalid
  use-flat-object-cache : 1224463.5 ops/s (fastest, 4.9% faster)
  latest-main           : 1167401.5 ops/s

bun, many_features, valid
  use-flat-object-cache : 1818182.5 ops/s (fastest, 9.1% faster)
  latest-main           : 1666586.5 ops/s

bun, optional_nullable, valid
  use-flat-object-cache : 1001718.5 ops/s (fastest, 15.0% faster)
  latest-main           : 870855 ops/s

bun, wide, invalid
  use-flat-object-cache : 494294 ops/s (fastest, 12.8% faster)
  latest-main           : 438072 ops/s

bun, wide, valid
  use-flat-object-cache : 685653 ops/s (fastest, 18.9% faster)
  latest-main           : 576454.5 ops/s
</code></pre>
</details>


Interestingly, in the above result, we see that this "flat cache" strategy **consistently improves the runtime performance** for both Bun and Node.js and the performance improvement is more significant for Bun (up to **18.9%** for "wide, valid" case). We might want to consider merging this strategy as long as the maintainer thinks the code is readable & maintainable enough.

fabian-hiller:
Thank you @naruaway. I will think about it.

Issue #96, opened 2023-08-13
Input type should be optional after withDefault


ooga:
When using `withDefault()`, the inferred input type of the defaulted value should be optional

<img width="948" alt="image" src="https://github.com/fabian-hiller/valibot/assets/3911114/f8183f8a-ad92-4669-aebd-c02399707415">

Maybe wrap returned schema with `optional()` ?

fabian-hiller:
That is correct. I will improve that. Thanks for the hint!

ooga:
What do you think about adding default value as parameter of `optional()` ?
It will be more explicit when reading schema

fabian-hiller:
Thanks for the idea. I'll give it some thought.

fabian-hiller:
I am not sure anymore if the input type should be `undefined`. It depends on how `withDefault` and the input type is interpreted. If the passed value in `withDefault` is the input in case of `undefined`, then the input is not `undefined`. I hope you understand what I mean.

The input type defines the initial data set that runs through the schema. When using `withDefault`, this type/value is no longer `undefined`. As the documentation describes, the input type can differ from the output using `transform`.

If you want the input type to be `undefined` as well, you can pass `optional(string())` as the first parameter instead of `string()`. It would then be necessary that `withDefault` changes the output type to `string` since it can no longer be `undefined`.

Please let me know your thoughts on this. I have not yet made a decision in this regard.

ooga:
> The input type defines the initial data set that runs through the schema. When using `withDefault`, this type/value is no longer `undefined`.

From my point of view, using `withDefault` creates a new schema, and we need to reflect the input of this newly created schema. 
The type/value that is no longer `undefined` is not the input anymore. 
The `withDefault` is like a `transform` but acting before the validation.

fabian-hiller:
This procedure, however, would have the consequence that also, in the [pipeline](https://valibot.dev/guides/pipelines/), the data type is optionally `undefined`, whereby almost none of the validation functions, like `minLength` and `regex`, would work anymore.

ooga:
The input of the pipeline and the "main" input are not the same, so the type shouldn't be the same. It might be challenging, but that is how it should work in my opinion

fabian-hiller:
Will think about it. Thank you for your feedback.

octet-stream:
[...]

PR #106, opened 2023-08-15
Add getDefault method 

### LINKED ISSUE #105: Add a way of getting default values from a schema
It is really useful to populate fields in a form.

Example of this method in `Yup` https://github.com/jquense/yup#schemagetdefaultoptions-object-any

Demivan:
Closes #105

### DIFF
--- a/library/src/methods/getDefault/getDefault.test.ts
+++ b/library/src/methods/getDefault/getDefault.test.ts
@@ -1,8 +1,14 @@
-﻿import { describe, expect, test } from 'vitest'
-import { number, object, record, recursive, string } from '../../schemas/index.ts'
-import { withDefault } from '../withDefault/index.ts'
-import { getDefault } from './getDefault.ts'
-import { minLength } from '../../validations'
+﻿import { describe, expect, test } from 'vitest';
+import {
+  number,
+  object,
+  record,
+  recursive,
+  string,
+} from '../../schemas/index.ts';
+import { withDefault } from '../withDefault/index.ts';
+import { getDefault } from './getDefault.ts';
+import { minLength } from '../../validations';
 
 describe('getDefault', () => {
   test('should get default value for simple schema', () => {
@@ -24,7 +30,7 @@ describe('getDefault', () => {
     const output3 = getDefault(schema2);
     expect(output3).toStrictEqual({ test: 'test' });
   });
-  
+
   test('should work with recursive schemas', () => {
     const schema = recursive(() => withDefault(string([minLength(3)]), 'test'));
     const output = getDefault(schema);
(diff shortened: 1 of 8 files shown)


fabian-hiller:
Thank you for your contribution! I think that it is not necessary to implement a function for this, since the default value can be made directly accessible or not?

Demivan:
I have made it the same as in `Yup`. `Yup` actually clones defaults if it is an object or an array, I think `Valibot` should clone it too, in case it is mutated.

https://github.com/jquense/yup#schemadefaultvalue-any-schema

Function default are also useful for defaults that change over time (`new Date`).

fabian-hiller:
With #96 there is another issue. I think we need to add a separate type definition for each schema, like I did for `transform` and `brand`. In this definition we could then add the key `default` to make the default value accessible there.

fabian-hiller:
> Function default are also useful for defaults that change over time (new Date).

When a date is created, it has a fixed value or doesn't it?

Demivan:
> When a date is created, it has a fixed value or doesn't it?

`new Date()` returns current date and time.

Function defaults are useful for nested schemas too.

```ts
const userSchema = object({ name: withDefault(string(), 'John Doe') })

// Populate default users array with one default user
const usersSchema = withDefault(array(userSchema), () => [getDefault(userSchema)])
```

Demivan:
> With https://github.com/fabian-hiller/valibot/issues/96 there is another issue. I think we need to add a separate type definition for each schema, like I did for transform and brand. In this definition we could then add the key default to make the default value accessible there.

I don't think it needs a definition for each schema. It would actually be the same signature as with `optional` method.
I like idea by @ooga https://github.com/fabian-hiller/valibot/issues/96#issuecomment-1678715229. Passing a default value (or a function to get one) to `optional` method would make schema definition simpler, and `withDefault` method can be removed. @fabian-hiller, what do you think?

fabian-hiller:
Yes, I think the idea of `optional` is not bad. However, I currently have to focus on my bachelor's thesis, and I do not want to make this decision quickly.

If we also allow a function for the default value with `optional` or `withDefault`, we could return the value with a setter. That way, it can still be accessed with `.default`.

Demivan:
> I currently have to focus on my bachelor's thesis, and I do not want to make this decision quickly.

No pressure :) This can wait.

> This is only true if instead of `new Date()` a function `() => new Date()` is passed and this is currently not supported by Valibot.

I have added support for this in this PR. Using a function to return a default value is useful when the default value depends on some external state too. 

> And in the code example, I would write the following:
```ts
const userSchema = object({ name: withDefault(string(), 'John Doe') })

// Populate default users array with one default user
const usersSchema = withDefault(array(userSchema), [userSchema.object.name.default])
```

This would create an array with just a name `['John Doe']` and not a user object `[{ name: 'John Doe' }]`.

> If we extend `optional`, perhaps we should do the same with `nullable` and `nullish`.

I agree.

fabian-hiller:
I deprecated `withDefault` and revised `optional`, `nullable` and `nullish`. Also, I added a `getDefault` method and added you as co-author.

fabian-hiller:
v0.15.0 is now available.

Demivan:
@fabian-hiller I haven't tried 0.15 yet, but I don't see implementation of `default` getter for wrapping schemas (`object`, `recursive`) etc. Would `getDefault` work on `object` schema that has properties with `default`?
You can see implementation for `object` schema in this PR.

fabian-hiller:
Yes I have looked at your implementation. No, that doesn't work because I think that is only needed in rare cases and therefore I don't want to add it globally to every schema. My plan was to wait to get feedback on this.

Demivan:
What I was thinking about is using `getDefault` for getting default values for the entire form. And because form would almost always use an `object` schema, that would not work.

Issue #122, opened 2023-08-25
cuid2 built in support


samualtnorman:
zod has built in support for `z.string().cuid2()`
it would help with migration if valibot also supported this out of the box

fabian-hiller:
Do you want to create a PR? You can follow the implementation of [`uuid`](https://github.com/fabian-hiller/valibot/blob/main/library/src/validations/uuid/uuid.ts).

samualtnorman:
I wouldn't mind creating a PR, the official package itself has a validator https://github.com/paralleldrive/cuid2/blob/fb07094487ba5ad038a51ef67bcb9c5f525c173b/src/index.js#L103, is the plan to bundle in that already existing function (to keep the 0 dependencies) or to reimplement the function like zod does.

if we go down the route implementing another validator, something note is that the official validator is more forgiving than the zod one (https://github.com/colinhacks/zod/blob/f59be093ec21430d9f32bbcb628d7e39116adf34/src/types.ts#L549).

the official implementation allows a-z on the first character and the zod version doesn't.
however the zod version also doesn't appear to check the length of the cuid2 (which is normally 24 unless configured otherwise)

fabian-hiller:
I prefer zero dependencies. We can make 24 the default and implement the ability to change it with the first optional parameter.

PR #130, opened 2023-08-29
Add support for cuid2s

samualtnorman:
See #122.


fabian-hiller:
I would prefer if we could avoid `{ checkLength: false }` altogether. I will probably work on this PR next week, as I am currently still busy with my bachelor thesis.

samualtnorman:
What are you thinking of doing instead?

fabian-hiller:
Maybe we should start without validating the length at all, since this can be done additionally with `length`, `minLength` and `maxLength` as you described. Then we wait if people create issues with questions or problems.

samualtnorman:
Is this what you're looking for?

fabian-hiller:
Yes seems to look good. I'll check your code in the next few days. Thanks a lot!

samualtnorman:
np

fabian-hiller:
I have your docs cached locally. I want to work out a consistent format first before we fill the API reference with content.

fabian-hiller:
v0.14.0 is now available.

Issue #131, opened 2023-08-29
strict() does not work with object that contains optional properties


FabienDehopre:
The `strict()` method throws a `ValiError` when the schema contains an optional property and the validated object did not contain that optional property.
Here is an example:
```ts
const schema = strict(
  object({
    environment: enumType(['dev', 'acc', 'prd']),
    logLevel: enumType(['ALL', 'TRACE', 'DEBUG', 'INFO', 'WARN', 'ERROR', 'OFF']),
    appInsights: optional(
      strict(
        object({
          instrumentationKey: string(),
          cloudRoleName: string(),
          endpointUrl: optional(string()),
          // more optional properties
        })
      )
    ),
  })
);
const obj = {
  environment: 'dev',
  logLevel: 'INFO',
};
const result = safeParse(schema, obj);
```
The `result.success` property is `false` and the `issues` property contains one issue with the message "Invalid keys".
After some debugging in the library code, I see that the `strict` method is looking for different property count between the input and the output.
Unfortunately, the output looks like this after all the other validations (in the `strict`'s `_parse` function):
```ts
{
  environment: 'dev',
  logLevel: 'INFO',
  appInsights: undefined,
}
```
So, the input has 2 properties and the output has 3 properties.

fabian-hiller:
You are right. We need to improve the tests and implementation. Are you interested in creating a PR?

FabienDehopre:
Sure. I can try to do a PR. Not sure when it would be available because I don't have a lot of free time.

fabian-hiller:
No problem. Don't feel obligated. I can take care of it next week.

FabienDehopre:
I think I've fixed the issue by updating the object schema parser to not assign undefined value to optional properties.

fabian-hiller:
Thank you for your contribution. In the next version the problem is fixed.

fabian-hiller:
v0.14.0 is now available.

FabienDehopre:
Great! I'll have a look later this weekend or next week.

PR #137, opened 2023-09-02
Fix optional strict by not assigning undefined value to optional properties in object(Async) function

FabienDehopre:
This PR should fix the issue #131.
The main issue wasn't in the `strict()` (or `strictAsync()`) function but more in the `object()` (or `objectAsync()`) function that always assigned the value to a property even when the value was `undefined` and the property was "optional".
The PR fix both the `object()` and `objectAsync()` function by not assigning optional properties when the value is undefined.

### DIFF
--- a/library/src/methods/strict/strict.test.ts
+++ b/library/src/methods/strict/strict.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from 'vitest';
-import { number, object, string } from '../../schemas/index.ts';
+import {number, object, optional, string} from '../../schemas/index.ts';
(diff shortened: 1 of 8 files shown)


fabian-hiller:
Thank you! I will check your changes in the next days.

fabian-hiller:
I think that your implementation still leads to problems. If a person now passes a property with the value `undefined` as input, it will be filtered out in `object`, causing `strict` to still throw an error. Also, I'm against making a change in different places when really only `strict` is affected. I think it is easiest to exclude `undefined` keys at `strict`. I will do the implementation and add you as co-author.

PR #138, opened 2023-09-02
fix: add InstanceSchema to transformAsync

Yovach:
I can't do
```typescript
transformAsync(instance(File), async (input) => {
  return file.length;
});
```

but 
```typescript
transform(instance(File), async (input) => {
  return file.length;
});
```
is working


fabian-hiller:
Thank you! I will try to release a new version with the fix in the next days.

fabian-hiller:
v0.14.0 is now available.

PR #141, opened 2023-09-04
feat: custom typeguard based validation

y-nk:
The motivation is to allow users to pass their own type guards which is the ultimate form of freedom imho.

My usecase is that i want to validate country codes, an enum which i pull from an external repo (here: https://github.com/lukes/ISO-3166-Countries-with-Regional-Codes/blob/master/all/all.json)

I don't want to have a pass of authoring on that list to avoid introducing corruption of data, so I'd rather write a type guard instead such as:

```ts
import iso3166Data from './all.json'

type Alpha2 = typeof iso3166Data[number]['alpha-2']
const Alpha2Codes = iso3166Data.map(entry => entry['alpha-2'])

export const isAlpha2(arg: any): arg is Alpha2 => Alpha2Codes.includes(arg)
```

The proposed features would allow this:

```ts
const Alpha2Schema = typeGuard(isAlpha2),
const countryCode = parse(Alpha2Schema, 'FR') // typed as Alpha2
```

of course you may argue that i could use the typeguard directly, but i'm trying to validate a form using `object()`, so i'd need it as a schema.

### DIFF
--- a/library/src/schemas/typeGuard/index.ts
+++ b/library/src/schemas/typeGuard/index.ts
@@ -1,0 +1,1 @@
+export * from './typeGuard.ts';
--- a/library/src/schemas/typeGuard/typeGuard.test.ts
+++ b/library/src/schemas/typeGuard/typeGuard.test.ts
@@ -1,0 +1,19 @@
+import { describe, expect, test } from 'vitest';
+import { parse } from '../../methods/index.ts';
+import { typeGuard } from './typeGuard.ts';
+
+describe('typeguard', () => {
+  type FooBar = 'foo' | 'bar'
+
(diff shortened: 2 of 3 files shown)


fabian-hiller:
For your use case you could also use `enumType` or `special`:

```ts
const FOO_BAR = ['foo', 'bar']

type FooBar = (typeof FOO_BAR)['number']

function isFooBar(arg: any): boolean {
  return FOO_BAR.includes(arg)
}

const EnumSchema = enumType(FOO_BAR);
const SpecialSchema = special<FooBar>(isFooBar);
```

y-nk:
I didn't know `special` would exactly act as what i was trying to achieve, but it looks like it. closing in favor 🙏

Issue #143, opened 2023-09-05
feature request: allow additional properties in object schema and allow to refuse additional properties via strict


FabienDehopre:
Actually, the object schema does not allow additional properties from the input to be "copied" to the output as-is (like zod).
I propose to allow additional properties to be transferred as-is from the input to the output.
Additionally, the strict method should allow to "block" those additional properties (again like zod).
I can provide a PR to implement these feature(s) request(s)

FabienDehopre:
I've added a comment to my PR to answer at your remark but here it is as well:

Here what I propose instead: Leave the default behavior of the object schema parser as it is actually but add a new `extraEntries` array property to the `_parser` result that can be validated by the `strict` method.
Additionally, I would also create 2 new "methods" for the object schema:
- `strip` (or any other name that you see fit): ignores the additional properties given in the input (so no additional properties in the output).
- `passthrough` (or any othe name that you see fit): copies the additional properties from input to the output as-is.

And update the `strict` method to throw an error if the `extraEntries` array isn't empty (more or less the same behavior as now).

What do you think about this proposition?

fabian-hiller:
Thanks for your feedback on this. I will probably have more time for open source again next week and will answer you then in detail.

fabian-hiller:
I added `passthrough` and `strip` in v0.14.0. Thank you for your contribution!

Issue #17, opened 2023-07-26
Discussion about `passthrough` to parse object keeping extra props


dmorosinotto:
Add  `passthrough` method inspired by [Zod](https://zod.dev/?id=passthrough)  to allow `parse object()` keeping extra props  that are present on original input object but not validated through Schema.

This maybe useful if you want to validate/trasform only _"a portion of an object"_ that you receive from/send to and API, but want to keep all the original extra props present in the object that are not specified in object() schema.

PS: some more discussion points:
1. if you prefer other name please let me know, I just used the same name of Zod... but I'm open to alternative/refactor
2. do you think that it's correct to implement it as a _"wrapper"_  and my modification to `unwrap` or do you prefer other way to [implement it](https://github.com/fabian-hiller/valibot/pull/15/commits/36c06d4a3443755b2df4bcc3cb51f885f2adb3f6) ?
3. about this I was even thinking to implement it changing the `object()` implementation adding another overload with an optional `strict: boolean` param  before error - but to do it I need to change the implementation of `getErrorAndPipe` handling 3 params to discriminate boolean, string, Pipe[] ...  I don't know if you'll approve/prefer this kind of implementation, maybe the result is more clear because the real nature of passthrough is just an `ObjectSchema` that keep the extra params, but don't know if you like it... 
4. Or maybe on the same "idea" i can create other 2 function `objectWithExtraProps` that return ObjectSchema but simply use my parse code to keep props... 

This are all the ideas I came up when implementing it, what's your opinion?

_Originally posted by @dmorosinotto in https://github.com/fabian-hiller/valibot/issues/15#issuecomment-1652418243_

fabian-hiller:
If we decide on a method. Can you think of alternative names besides `passthrough`?

dmorosinotto:
About alternative names I can suggest `keepExtras` or `keepExtraProps`

But I have another question do you think that we can change the implementation of `merge` to make it works with Record + Object? Because actually it allow to merge only `ObjectSchema[]`, but if we can make it works with Record+Object we can achieve the "passtrough" by simply `merge( record(string(),any()) , object({...schema...}) );`

I looked at merge code and record() constructor but the two are not compatible to simply join the parse as it's implemented now... 
Or if you think that this idea maybe good, we can create a new method `extends` to specify merge Record+Object correctly handle the parse and input/output type ;-)

Waiting for your feedback @fabian-hiller  about how do you prefer to procede with this issue

[...]

PR #144, opened 2023-09-05
Allow additional properties in object schema and allow to block those additional properties via the strict method

FabienDehopre:
This PR implements the feature request #143

### DIFF
--- a/library/src/methods/strict/strict.test.ts
+++ b/library/src/methods/strict/strict.test.ts
@@ -11,10 +11,9 @@ describe('strict', () => {
     const output1 = parse(schema, input1);
     expect(output1).toEqual(input1);
 
-    const keysError = 'Invalid keys';
     const input2 = { key1: 'test', key2: 123, key3: 'unknown' };
-    expect(() => parse(schema, input2)).toThrowError(keysError);
+    expect(() => parse(schema, input2)).toThrowError('Invalid keys: key3');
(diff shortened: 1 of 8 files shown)


fabian-hiller:
This seams related to https://github.com/fabian-hiller/valibot/pull/15

Since this this a modular library and current most users expect the current behaviour, I prefer to implement a method instead of changing the `object` schema. I am happy to receive feedback on this.

FabienDehopre:
Indeed. I didn't spot that PR (and related issue/discussion #17).
I understand your point.
Here what I propose instead: Leave the default behavior of the object schema parser as it is actually but add a new `extraEntries` array property to the `_parser` result that can be validated by the `strict` method.
Additionally, I would also create 2 new "methods" for the object schema:
- `strip` (or any other name that you see fit): ignores the additional properties given in the input (so no additional properties in the output).
- `passthrough` (or any othe name that you see fit): copies the additional properties from input to the output as-is.

And update the `strict` method to throw an error if the `extraEntries` array isn't empty (more or less the same behavior as now).

What do you think about this proposition?

fabian-hiller:
Thanks for your feedback on this. I will probably have more time for open source again next week and will answer you then in detail.

fabian-hiller:
I added `passthrough` and `strip` in v0.14.0. Thank you for your contribution!

PR #151, opened 2023-09-10
feat: add a validator for the ulid spec

jmcdo29:
[ULIDs][ulid] are another approach for sortable, unique identifiers. I've been using them in a personal project or two and would love to see official support from valibot rather than having to define the same custom check myself in these projects.

[ulid]:https://github.com/ulid/spec


fabian-hiller:
Thank you for your contribution! I will check your code later.

library/src/validations/ulid/ulid.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @link https://github.com/ulid/spec
+ *
+ * @param error The error message.
+ *
+ * @returns A validation function.
+ */
+export function ulid<TInput extends string>(error?: string) {
+  return (input: TInput): PipeResult<TInput> => {
+    if (
+      input.length !== 26 ||
```
fabian-hiller, line 15: Is this necessary? The regex already checks the length. I'll probably remove it. Let me know if there is an important reason to keep it.

jmcdo29: I actually added the length for the specific case of having a valid ULID _plus_ extra characters. The Regex could probably be updated instead to `^[0123456789ABCDEFGHJKMNPQRSTVWXYZ]{26}$` so that it's _only_ those 26 characters. Let me see if that works for the tests I've added

jmcdo29: Yeah, good call. Using the `^regex$` format works as expected. Thanks for the callout here

fabian-hiller:
I minified the regex, revised the tests and updated the website. If everything looks good to you, I'll merge it later.

jmcdo29:
Thanks for adding the exports and website info and making the regex cleaner :smile: It all looks good to me

I wanted to also mention that I created an integration with the [NestJS](https://docs.nestjs.com) framework through the use of [typeschema](https://github.com/decs/typeschema). Would I be good to add that under the Ecosystem > Frameworks, or should it go under Libraries?

fabian-hiller:
Thanks! I would add NextJS under Frameworks or API Libraries and if you also want a link to nest-lab, that chould best fit under Utilities or?

fabian-hiller:
The new version is available.

adoublef:
There is a slight issue with the import paths for deno. #181 is a PR for that

Issue #161, opened 2023-09-15
[✨ Feature Request]: Can we return an `issues` object on the custom dev defined validation pipeline methods?


demarchenac:
Hi @fabian-hiller! 

I hope you're having a great day!

Ok, basically what I'm asking is if I could have a custom pipeline validation method to return multiple issues, _why?_ you may ask.

So I can summarize common-wise validations within an `object()` schema such as:

```ts
const mySchema = object({
    field1: optional(string()),
    field2: optional(string()),
    field3: optional(string()),
    ...,
    fieldN: string(),
}, [
    (input) => {
        if(input.fieldN && !input?.field1){
            return { issue: {...} } // return some issue related to field1
        }
       return { output: input }
    },
    (input) => {
        if(input.fieldN && !input?.field2){
            return { issue: {...} } // return some issue related to field2
        }
       return { output: input }
    },
    (input) => {
        if(input.fieldN && !input?.field3){
            return { issue: {...} } // return some issue related to field3
        }
       return { output: input }
    },
])
```

into something more compact like:
```ts
const mySchema = object({
    field1: optional(string()),
    field2: optional(string()),
    field3: optional(string()),
    ...,
    fieldN: string(),
}, [
    (input) => {
        const issues = []

       ['field1', 'field2', 'field3'].forEach((fieldKey) => {
            if(input.fieldN && !input?.[fieldKey]){
                 issues.push({ issue: {...} }) // some issue related to fieldKey*
            }
       })

       return issues.length > 0 ? { issues } : { output: input }
    },
])
```

It's worth noting that this one was one basic example that I thought to explain the feature request, but it could be any kind of complex validations.

What do you think about this one?

fabian-hiller:
Thank you for creating this issue! I plan to change `issue` to `issues` with the next version.

fabian-hiller:
However, I recommend using our validation functions such as `custom` whenever possible. This usually makes the code more readable.

demarchenac:
Hey absolutely, this was just a silly example I came by to explain my feature request, thanks for working through this one! 👏🏻

fabian-hiller:
The new version is available.

PR #162, opened 2023-09-15
feat(executePipe): add issues

demarchenac:
Hi, @fabian-hiller!

Here's a proposal I thought for #161, please let me know if there's anything that could be improved. It is worth noticing that I may need a little bit of help with testing `result.issues`

### DIFF
--- a/library/src/types.ts
+++ b/library/src/types.ts
@@ -123,10 +123,17 @@ export type PipeResult<TOutput> =
   | {
       output: TOutput;
       issue?: undefined;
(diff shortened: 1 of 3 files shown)


fabian-hiller:
I think that the code is easier to unterstand if we remove `issue` and replace it with `issues`. However, since this is a little more work, I can do it on Sunday.

fabian-hiller:
I have implemented the changes. With the next version it will be possible.

PR #164, opened 2023-09-16
Add option to skip pipelines when parsing

jonlambert:
Firstly, thanks for Valibot 😊 It's a great library and I'm excited to use it in projects moving forward! 

## Overview

When validating a value, it would be useful to be able to confirm the type of the object without throwing if the values don't pass formatting checks. Valibot could support that by accepting an option to skip all pipeline checks when parsing.

For example: 

```ts
const schema = object({
  name: string(),
  email: string([email()]),
});

const initialValues: unknown = { name: '', email: '' };

parse(schema, initialValues); // will throw
parse(schema, initialValues, { skipPipes: true }); // won't throw
```

This could be really useful in circumstances such as initial form values, where we do care about the type of the values, but want to run formatting checks separately.

If I understand correctly, Valibot's rules for not transforming the data type inside pipelines means that this should work for any schema.

## Alternatives

Users could maintain two schemas: one for the type-level checks, one for the formatting. But that would be additional overhead, and cause issues if their Valibot schema is auto-generated.

## Notes

Please forgive me if this functionality already exists, or if I've overlooked a caveat here :) Also happy to rename the option too if another name is more appropriate.

### DIFF
--- a/library/CHANGELOG.md
+++ b/library/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the library will be documented in this file.
 
+## vX.X.X (Month DD, YYYY)
+
+- Add `skipPipe` option to skip execution of pipeline (pull request #164)
+
 ## v0.16.0 (September 16, 2023)
 
 - Add `ulid` validation (pull request #151)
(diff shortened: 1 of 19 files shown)


fabian-hiller:
Thank you for creating this PR. The idea is good. I'll check your code as soon as I find time and give you feedback then.

jonlambert:
Thanks for your swift consideration, @fabian-hiller! Singular `skipPipe` is much better.

fabian-hiller:
I also find `skipPipes` suitable. It was only important to me to make it consistent to `abortPipeEarly`.

PR #166, opened 2023-09-17
Skip pipeline when calling is()

jonlambert:
As `is()` is primarily designed for use as a type guard, it would make sense to skip pipelines during this check too 🙂

Currently: 

```ts
is(string([email()]), 'foo') // this check would fail, though types are correct
```

Proposed: 

```ts
is(string([email()]), 'foo') // this check would pass, as it's a string
```

## Notes

**This is a breaking change to `is()`**

Following on from #164.

### DIFF
--- a/library/src/methods/is/is.test.ts
+++ b/library/src/methods/is/is.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, test } from 'vitest';
 import { number, object, string } from '../../schemas/index.ts';
+import { minLength } from '../../validations/index.ts';
 import { is } from './is.ts';
 
 describe('is', () => {
@@ -16,5 +17,9 @@ describe('is', () => {
     expect(is(string(), 123)).toBe(false);
     expect(is(number(), 'hello')).toBe(false);
     expect(is(object({ test: string() }), {})).toBe(false);
+  });
+
+  test('should skip pipelines', () => {
+    expect(is(string([minLength(1)]), '')).toBe(true);
   });
 });
--- a/library/src/methods/is/is.test.ts
+++ b/library/src/methods/is/is.test.ts
@@ -19,7 +19,8 @@ describe('is', () => {
     expect(is(object({ test: string() }), {})).toBe(false);
   });
 
-  test('should skip pipelines', () => {
-    expect(is(string([minLength(1)]), '')).toBe(true);
+  test('should accept parsing options', () => {
+    expect(is(string([minLength(1)]), '', { skipPipe: true })).toBe(true);
+    expect(is(string([minLength(1)]), '', { skipPipe: false })).toBe(false);
   });
 });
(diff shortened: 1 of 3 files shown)


fabian-hiller:
To just check the data type, you are right and it can be added. However, I would add it as a configuration and leave it as it is by default. I strongly suspect that this is the expected behavior of most users of Valibot.

```ts
// Default behaviour with pipes
is(Schema, input);

// Skip pipes and only check data type
is(Schema, input, { skipPipe: true });
```

jonlambert:
Yep fair enough! I guess it depends on the primary purpose of `is()`. Personally I'd use it as a typecheck, so would prefer the simple boolean, but I defer to your judgement there, @fabian-hiller 🙂 Thanks again for your consideration!

fabian-hiller:
If you are interested in this feature, you can implement it. Otherwise we can wait for feedback from other people.

jonlambert:
Yep happy to implement, @fabian-hiller 🙂 I do think it might be worth waiting for some more opinions though, if that's ok. I'm undecided about what might be the best option.

To add to the options you've listed above:

```ts
// Might be more clearly represented with an opt-out approach
is(Schema, input, false);

// Function signature would be
export function is<TSchema extends BaseSchema>(
  schema: TSchema,
  input: unknown,
  runPipe = true
): input is Input<TSchema> {
  return !schema._parse(input, { abortEarly: true, skipPipe: !runPipe }).issues;
}
```

Would be good to hear from others what they might expect the behaviour of `is()` to be, as in the docs it's very much presented as a "type guard" 🙂

fabian-hiller:
I think that if someone defines the schema `string([email()])` and then uses `is`, they expect `false` to be returned for an empty string. If we only validated `string` and not `email`, I suspect this would lead to bugs in many cases.

Maybe we should implement it explicitly with `is(schema, input, { skipPipe: true })` for now, leaving the possibility to add more configurations in the future. Furthermore this way it is consistent with `parse` and `safeParse`.

jonlambert:
Yeah I do agree with you, @fabian-hiller. Definitely good to keep consistent with `parse` and `safeParse`. I think you're right that it would be unexpected for it to ignore checks invisibly.

I've added the `info` argument to `is()` 🙂

jasikpark:
> I think that if someone defines the schema `string([email()])` and then uses `is`, they expect `false` to be returned for an empty string. If we only validated `string` and not `email`, I suspect this would lead to bugs in many cases.
> 
> Maybe we should implement it explicitly with `is(schema, input, { skipPipe: true })` for now, leaving the possibility to add more configurations in the future. Furthermore this way it is consistent with `parse` and `safeParse`.

I would agree, I would expect a schema passed to `is()` to fully check the schema shape - if we were clever `email()` could further refine the shape of the string 🤷

fabian-hiller:
I added `skipPipe` in v0.22.0 as a configuration to `is`.

```ts
// Skip pipes and only check data type
is(Schema, input, { skipPipe: true });
```

Issue #155, opened 2023-09-11
[feat]: Parse schema with partial fallback values


hermanseder:
It would be nice to have a functionality to parse a schema with partial fallback values.

For example, that this input:
```js
const schema = object({
  name: fallback(string(), 'test'),
  age: number(),
});

console.log(parse(schema, {}));
```
results in this output:
```js
{
   age: undefined
   name: 'test'
}
```

Or is such a scenario already possible by combining multiple functions together?

As idea, I implemented here a custom version of partial, which uses `fallback` instead of `optional`:
https://stackblitz.com/edit/valibot-partial-fallback-example

fabian-hiller:
Maybe with `optional`? If not, please describe in more detail what behavior you are looking for. I don't quite understand it yet.

```ts
const schema = object({
  name: optional(string(), 'test'),
  age: number(),
});
```

fabian-hiller:
For what exactly do you need the fallback values? Technically this is possible, but not that easy. Especially it is difficult to make it type safe.

hermanseder:
My intention is, that I can parse a schema, where some of the properties have a fallback value, and some not. From this schema I want to extract, only the properties which has a fallback value, into an object. All other properties, which have no fallback value, should be undefined in the extracted object.

As background, I want to use `Valibot` that different components can define in a generic way their required configuration schema. A further requirement is, that these component can also define a partial default value, and therefore I need to extract these values. Another solution would be to separate the schema and the default configuration. But I thought it would be nice to combine them into one (valibot) configuration object.

hermanseder:
Thanks for your detailed explanation. Then I will implement a custom logic in my project. I thought, such a functionality could maybe be helpful for other users.

fabian-hiller:
Yes, feel free to share your implementation. Maybe I will integrate it in `getDefault` or add another method called `getDefaults`.

fabian-hiller:
Were you able to find a solution to your problem?

fabian-hiller:
I have implemented `getDefaults`. The API will be available in the next version.

fabian-hiller:
v0.20.0 is now available

BThomann:
Awesome thank you! In the end the `getFallbacks` was exactly what we have been looking for!

PR #169, opened 2023-09-18
[feat]: Parse schema with partial fallback values #155

BThomann:
Implements https://github.com/fabian-hiller/valibot/issues/155 including unit tests to reflect behavior.

Happy for feedback 😀


fabian-hiller:
I think that `getDefaults` should return the default value of `optional`, `nullable` and `nullish` and not the value of `fallback`. I think that `fallback` was only an example in issue #155.

fabian-hiller:
I have implemented `getDefaults`. The API will be available in the next version. Thank you for your contribution!

fabian-hiller:
v0.20.0 is now available

PR #177, opened 2023-09-21
Add `fallback` property to `fallback()` return value

### LINKED ISSUE #174: Include fallback value in `fallback()` return
https://github.com/fabian-hiller/valibot/blob/f3679064037eae9bd31c97d32aa78517d12b76d9/library/src/methods/fallback/fallback.ts#L15-L37

Similar to how `optional()` returns a `default` key, I'd like to be able to access `fallback()`'s fallback value before parsing because I'd like to supply the fallback value as an initial value to a CLI prompt. Would you be okay if I created a PR to add this?

brandonpittman:
- Make info arg optional
- Add `fallback` property to return value

Closes #174

### DIFF
--- a/library/src/methods/fallback/fallback.test.ts
+++ b/library/src/methods/fallback/fallback.test.ts
@@ -27,4 +27,9 @@ describe('fallback', () => {
     const output3 = parse(schema3, input2);
     expect(output3).toEqual(input2);
   });
+
+  test('should have fallback property', () => {
+    expect(schema1.fallback).toEqual('test');
+    expect(schema2.fallback).toEqual('test');
+  });
 });
--- a/library/src/methods/fallback/fallback.ts
+++ b/library/src/methods/fallback/fallback.ts
@@ -12,7 +12,7 @@ import type { FallbackInfo } from './types.ts';
  */
 export function fallback<TSchema extends BaseSchema>(
   schema: TSchema,
-  value: Output<TSchema> | ((info: FallbackInfo) => Output<TSchema>)
+  value: Output<TSchema> | ((info?: FallbackInfo) => Output<TSchema>)
 ): TSchema {
   return {
     ...schema,
--- a/library/src/methods/fallback/fallback.ts
+++ b/library/src/methods/fallback/fallback.ts
@@ -16,7 +16,8 @@ export function fallback<TSchema extends BaseSchema>(
 ): TSchema {
   return {
     ...schema,
-
+    fallback:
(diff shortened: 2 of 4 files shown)


fabian-hiller:
I think that it can make sense to implement it with a getter, similar to `optional`, `nullable` and `nullish`:

https://github.com/fabian-hiller/valibot/blob/v0.17.0/library/src/schemas/optional/optional.ts#L48-L52

fabian-hiller:
Please reuse the getter in the `_parse` method.

brandonpittman:
The getter won't be able to take any arguments though, right?

fabian-hiller:
Ahh... you are right

brandonpittman:
Getters can't take args, right? We wouldn't be able to pass the {info, issues} object to it.

fabian-hiller:
Yes, you can leave it as it is for now. I will try to check the PR over the weekend. Thanks for your contribution!

fabian-hiller:
Do you think that we should change the getter to a `getFallback` method? This would allow a unified procedure if we do the same with the `default` getter at `optional`, `nullable` and `nullish`. With a getter, the API is more sexy but also more complicated for developers with less experience.

brandonpittman:
> Do you think that we should change the getter to a `getFallback` method? This would allow a unified procedure if we do the same with the `default` getter at `optional`, `nullable` and `nullish`. With a getter, the API is more sexy but also more complicated to understand for developers with less experience.

Do you mean like [`getDefault()`](https://github.com/fabian-hiller/valibot/blob/main/library/src/methods/getDefault/getDefault.ts)?

Kinda makes me think we should make have a `default` getter on `fallback()` and let `getDefault()` work with `fallback()` schemas.

fabian-hiller:
For `optional` is is a default value but for `fallback` it is a fallback value. I think that we should not mix that.

fabian-hiller:
The current question is whether we should replace `get default() {...}` with `getDefault() {...}` in `optional`, `nullable` and `nullish` so that we can improve the implementation of `fallback` with `getFallback() {…}` in a uniformed way.

fabian-hiller:
> The current question is whether we should replace `get default() {…}` with `getDefault() {…}` in `optional`, `nullable` and `nullish` so that we can improve the implementation of `fallback` with `getFallback() {…}` in a uniformed way.

fabian-hiller:
Unfortunately, I could not push changes to this PR. So I committed the changes directly to the main branch and added you as a co-author.

fabian-hiller:
v0.20.0 is now available

PR #178, opened 2023-09-22
Fix deno import issue for missing file ext and add new linting rule

syhol:
I tried importing valibot from deno.land and had issues with ulid.ts missing the `.ts` in its imports.

It's easy enough to recreate the problem, just run `deno run https://deno.land/x/valibot@v0.17.0/mod.ts` in your shell today and you'll see the error. It's the same error when trying to use it as an import.

This PR fixes the issues with missing extensions and adds a new eslint rule to prevent it from happening again.


fabian-hiller:
Thank you for your contribution! I will review your PR over the weekend.

PR #181, opened 2023-09-23
deno requires `.ts` prefix in-order for this to work

adoublef:
Seems that the import paths for the `ulid` module are different to the other validators and so there is a break in compatibility with Deno


fabian-hiller:
Thanks for your contribution! I think that this PR is a duplicate of #178.

adoublef:
ah missed that, then this one can be closed, good to know it'll be fixed in next release

Issue #156, opened 2023-09-13
Incorrect TS types  for `recrod`


abd2561024:
Hi! I have a problem with TS types after `Output` following construction. 

```TS
const SOME_NAME = {
  FOO: "FOO", 
  BAR: "BAR"
} as const;

const ValibotSchema = record(nativeEnum(SOME_NAME), array(string([minLength(1)]), [minLength(1)]));

type ValibotOutType = Output<typeof ValibotSchema>;

const valibotData: ValibotOutType  = {}; // here error with TS ->  "Type  {}  is missing the following properties from type  { FOO: string[]; BAR: string[]; } :  FOO, BAR"

valibotData[SOME_NAME.FOO] = ['1'];
```

I expect, that  `record(nativeEnum(SOME_NAME), array(string([minLength(1)]), [minLength(1)]))` will get me something like `Partial<Record<keyof typeof SOME_NAME, [string, ...string[]]>>` as it returns in `Zod` with follow code:

```TS
const ZodSchema = z.record(z.nativeEnum(SOME_NAME), z.array(z.string().min(1)).nonempty());
  
type ZodOutType = z.infer<typeof ZodSchema>;

const zodData: ZodOutType  = {}; // here all fine

zodData[SOME_NAME.FOO] = ['1'];

```
I'm not sure, than `Zod` correct outputs `record` keys with an `optional` generic (Here is the [issue](https://github.com/colinhacks/zod/issues/1215) with this problem).  So, is this a correct behavior of `record()` logic and I jsut miss something? Or maybe `valibot` has some ability to make `record` keys not `required`?
Thanks a lot!

P.S. Technically, the `parse` and `safeParse` works correctly, you can specify only part of the keys in `record`

fabian-hiller:
Thank you for creating this issue. I will investigate the types in the coming days.

abd2561024:
@fabian-hiller I created a [PR](https://github.com/fabian-hiller/valibot/pull/184) with fixing types for record schema. Now it would be work as in `Zod` implementation (I mean types, technically `parse` and `safeParse` work great as I wrote early)

Otherwise it could be something like other implementation for `partial` method, which would make `record` keys optional,
```TS
declare function partial<TRecordKey extends RecordKey, TRecordValue extends BaseSchema>(
  schema: RecordSchema<TRecordValue, TRecordKey>,
  pipe?: Pipe<Partial$2<RecordOutput<TRecordKey, TRecordValue>>>
[...]

PR #184, opened 2023-09-25
fix: change record keys types to optional

abd2561024:
Fix for  `record` type as it work in `Zod` implementation. This is a variouse of resolving follow [issue](https://github.com/fabian-hiller/valibot/issues/156)

```TS
const SOME_NAME = {
  FOO: "FOO", 
  BAR: "BAR"
} as const;

const ValibotSchema = record(nativeEnum(SOME_NAME), string());

type ValibotOutType = Output<typeof ValibotSchema>;

const valibotData: ValibotOutType  = {}; // here error with TS ->  "Type  {}  is missing the following properties from type  { FOO: string[]; BAR: string[]; } :  FOO, BAR"

valibotData[SOME_NAME.FOO] = 'SOME';
```
Expected, that `record(nativeEnum(SOME_NAME), string())` will get `Partial<Record<keyof typeof SOME_NAME, string>>` as it returns in `Zod` with follow code:
```TS
const ZodSchema = z.record(z.nativeEnum(SOME_NAME), z.string());
  
type ZodOutType = z.infer<typeof ZodSchema>;

const zodData: ZodOutType  = {}; // here all fine
zodData[SOME_NAME.FOO] = 'SOME'

### DIFF
--- a/library/src/schemas/record/record.ts
+++ b/library/src/schemas/record/record.ts
@@ -21,10 +21,10 @@ export type RecordKey =
 export type RecordSchema<
   TRecordValue extends BaseSchema,
   TRecordKey extends RecordKey = StringSchema,
-  TOutput = RecordOutput<TRecordKey, TRecordValue>
+  TOutput = Partial<RecordOutput<TRecordKey, TRecordValue>>
 > = BaseSchema<RecordInput<TRecordKey, TRecordValue>, TOutput> & {
   schema: 'record';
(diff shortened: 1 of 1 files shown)


fabian-hiller:
Basically I have done the following... Whenever an enum was passed I apply `Partial` in `RecordInput` and `RecordOutput`. This way we fix the existing problem without changing the types for other record schemas.

PR #194, opened 2023-10-03
Add new validator notValue

### LINKED ISSUE #193: Add notEqual validator
Since there is already an `equal` validator I think having the negated version of it would be a useful addition.

lo1tuma:
Fixes #193


fabian-hiller:
Thank you! I will review your code as soon as I find some time.

fabian-hiller:
What do you prefer? Is there an argument for one or the other?

```ts
// Option 1
const Schema1 = string([notValue('foo')]);
const Schema2 = string([notLength(10)]);

// Option 2
const Schema3 = string([valueNot('foo')]);
const Schema4 = string([lengthNot(10)]);
```

lo1tuma:
I would prefer option 1, although I don’t have very strong opinion on that.

fabian-hiller:
Thanks. At the moment, I think that's better, too. I'll try to review this PR over the weekend.

PR #207, opened 2023-10-08
feat(dot-notation): add partial implementation

ecyrbe:
Hello Fabian,

Like promised, here an example of how to add dot notation for valibot.
Users would need to add a registration step to register some notations.
a vite plugin could add them automatically if needed.

This is using:
- a type registry declared globally (interfaces are automatically overloaded by typescript).
- a runtime registry to add mixins at runtime

if you don't import the registry, you don't pay the cost and you don't have the dot notation.
so this would be optin for users that want it

if you want to explore more, happy to help make it a reality and improve DX, since this is a quick demo hack.

### DIFF
--- a/library/src/registry/registry.ts
+++ b/library/src/registry/registry.ts
@@ -1,9 +1,5 @@
 import type { PipeItem } from '../types';
 
-// register  a validator function by it's name , it's validation function and the type of the schema it validates
-// you can register many validators for the same schema type
-// e.g. register('number', min) // min is the function to register and the name of the function is extracted to register it
-
 type Validator<Args extends unknown[], T> = (...args: Args) => PipeItem<T>;
 type RegistryItem = (...args: unknown[]) => unknown;
 
--- a/library/src/registry/registry.ts
+++ b/library/src/registry/registry.ts
@@ -1,0 +1,41 @@
+import type { PipeItem } from '../types';
+
+// register  a validator function by it's name , it's validation function and the type of the schema it validates
+// you can register many validators for the same schema type
+// e.g. register('number', min) // min is the function to register and the name of the function is extracted to register it
+
+type Validator<Args extends unknown[], T> = (...args: Args) => PipeItem<T>;
+type RegistryItem = (...args: unknown[]) => unknown;
+
+const registry: Record<string, Record<string, RegistryItem>> = {};
+
+export function register<Args extends unknown[], T>(
+  type: string,
+  validator: Validator<Args, T>
+) {
+  const name = validator.name;
+  if (!name) {
+    throw new Error('Cannot register anonymous function');
+  }
+  if (!registry[type]) {
+    registry[type] = {};
+  }
+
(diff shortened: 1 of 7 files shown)


fabian-hiller:
Thank you! Can you describe in more detail with code examples how your solution works?

ecyrbe:
> Thank you! Can you describe in more detail with code examples how your solution works?

yes, 

this part put one and only one validation function type to a type registry that is global.
```ts
declare global {
  export interface StringRegistry {
    email: typeof email;
  }
}
```
allowing it to be injected in schema string type definition only if you import it.

this part put the function in a runtime registry so in can  injected to the pipe of validbot :
```ts
register('string', email);
```
The thing here is that every registration is independant of one another (they are isolated). So they can still be treshaked independently.
Registering the types and the runtime are opt-in. meaning if you don't want to use it, there is no impact in current valibot way of using pipes. You can use both if you want.

Now here is how types are injected in the type string schema : 
```ts
// helper that take a type registry (that is only updated if you register a validator) and add the signature to the schema
type SchemaValidators<TRegistry, TThis> = {
  [K in keyof TRegistry]: (
    ...args: Parameters<Extract<TRegistry[K], (...args: any) => any>>
  ) => TThis; // this is a definition for a mixin method that returns this so they are chainable
};

// schema with augmented validators (like .email() ) only for registered ones
export interface StringSchema<TOutput = string>
  extends BaseSchema<string, TOutput>,
    SchemaValidators<StringRegistry, StringSchema<TOutput>> {
  schema: 'string';
}
```

and for the `dot` validators they are injected with a mixin that comes from the runtime registry, again only adding those that are registered : 
```ts
return {
    schema: 'string',

    _pipe: pipe, // added to allow the mixins to update the pipe

    async: false,
    _parse(input, info) {
      // Check type of input
      if (typeof input !== 'string') {
        return getSchemaIssues(
          info,
          'type',
          'string',
          error || 'Invalid type',
          input
        );
      }
      return executePipe(input, this._pipe, info, 'string'); // using this._pipe to have access to the updated pipe from the mixins
    },
    ...(mixinsOf('string'), // inject the mixins, thats are the methods the user registered
  };
```

To sumarize, the solution is to use a registry to only register the functions you want to have with dot notation.
the registry then injects a function that adds the mixin to the pipe and allows to chain them

So the solution is not transparent for the user. they have to register the methods they want to have chainable methods.
To improve DX, we could register the all the types automatically since they don't contribute to bundle size.
And then have a vite plugin that automatically registers only the used methods, so that it would be transparent for users

fabian-hiller:
Thank you for your explanation. I think that I have understood it. The idea is interesting. This could be the basis for a separate open source project that emulates Zod with a much smaller bundle size. For now, I don't plan to implement it for Valibot, because I like our API design and the disadvantage in tooling can also be compensated by AI tools or a VS code extension.

ecyrbe:
closing then.

PR #205, opened 2023-10-08
chore: add eslint-plugin-jsdoc

kazizi55:
Implements https://github.com/fabian-hiller/valibot/issues/185.

I chose not `plugin:jsdoc/recommended-typescript`, but `plugin:jsdoc/recommended-typescript-error`, so that the review time and effort can be reduced. 😄 

FYI
https://github.com/gajus/eslint-plugin-jsdoc#readme:~:text=typescript%22%5D%0A%7D-,...or%20to%20report%20with%20failing%20errors%20instead%20of%20mere%20warnings%3A,-%7B%0A%20%20%22extends%22

PR #208, opened 2023-10-09
ci: add lint and format check

kazizi55:
Since the number of implemented linter rules is increasing (https://github.com/fabian-hiller/valibot/pull/205 https://github.com/fabian-hiller/valibot/pull/202) and will be increased very much, I added lint (and format) check to CI.

This will reduce the review time and effort 😄


kazizi55:
@fabian-hiller 
Thank you too 😄

>I think it makes sense to do the format and lint checks before the tests. What do you think about that?

Yes, I think so too, because the format and lint checks don’t usually take so much time, therefore there is a slight time difference between running all in series  and running each check in parallel.

>Do you know how other projects do this?

Here are some examples:

- zod: running lint checks and tests in parallel
    - https://github.com/colinhacks/zod/blob/f59be093ec21430d9f32bbcb628d7e39116adf34/.github/workflows/test.yml#L11-L71
- react hook form: running all in series
    - https://github.com/kazizi55/react-hook-form/blob/0d565fcf944621f5510da4623129ffbd238f9530/.github/workflows/build-test.yml#L30-L38
- typescript-eslint: running lint checks and tests in parallel
    - https://github.com/typescript-eslint/typescript-eslint/blob/main/.github/workflows/ci.yml#L74-L294

.github/workflows/node.yml
```diff
@@ -42,18 +43,38 @@ jobs:
...
+    needs: install
+    runs-on: ubuntu-latest
+
+    steps:
+      - run: pnpm lint
         working-directory: website
-        run: pnpm lint
 
-      - name: Run formatter in website
+  format_website:
```
fabian-hiller, line 55: I think that the job `format_library` is missing.

fabian-hiller:
Thank you for your work. I think something is not configured properly yet. At least not all the checks are running yet.

fabian-hiller:
Did you configure that the job `install` must run first before the others are allowed to start?

kazizi55:
Thank you for your quick response!
I'll check the detail 🙏

kazizi55:
Fixed! 😄 
https://github.com/fabian-hiller/valibot/pull/208/commits/b7ec31d52563750b39f60c8f5adad28a5967bddf

> Right now I think it's best to run everything in parallel and list each check individually on a PR so you know exactly what doesn't fit. What do you think?

I think both are acceptable now, but in the long term, running everything in parallel will be better, because of the expectation of the increase of CI check, which causes delay.

fabian-hiller:
Thank you for your contribution! I will review your changes in the next 24 hours.

fabian-hiller (approved):
Thank you so much for improving this. I find this very useful for all future PRs. The code looks good. I just want to rename a few labels. Once I have done that, I will merge the PR.

kazizi55:
Thank you for your words! 😄

>  I just want to rename a few labels. Once I have done that, I will merge the PR.

OK, go ahead!

Issue #191, opened 2023-10-01
Custom issue formatting: how to determine expected parameters of validations


lo1tuma:
Let’s say I have the following schema:

```ts
const schema = string([minLength(3)]);`
```

I want to implement a custom formatter, that would output the following error message I case the validation fails:

```ts
const result = safeParse(schema, 'a');
console.log(myFormatter(result.issues)); // Expected string to have at least 3 character(s) but instead got 1
```

While the `input` data is available and I can easily get the length of the input, it seems like there is no way to get the expected amount of characters from the `Issue`. Or am I missing something?

fabian-hiller:
Yes, that is correct. This information is not included. We could add the optional property `requirement` to an issue which contains this information.

lo1tuma:
Sounds like a good idea. Would it be a `string` or would the type of `requirement` depend on what `validation` is? So in case of `min_length` it would be a `number` and for `includes` it would be `string`.

fabian-hiller:
I think we need to declare them as `any` or `unknown` because the requirement can be anything. For example, with `custom` it can be a function. We should investigate if there is a way to make it type safe in the long run.

lo1tuma:
In this case I would prefer to make it `unknown`, since I try to avoid using `any` completely. Another idea would be to make it at least type safe for the builtin validations and only `unknown` in case of `custom` validations. So that the `validation` property acts as a discriminator, so I can be sure that when I check `issue.validation === 'min_length'` that `requirement` will be always a number.

fabian-hiller:
Yes, as soon as I find the time, I will take a look at it.

lo1tuma:
I’ve looked a little bit into this. I would love to get this feature type safe, but I think one issue is that on `Issue` the `validation` property is defined as `string` and that custom validation functions can put any value in there. I order to use `validation` as a discriminator for the new `requirement` we would need to change it to union type of string literals. But I guess this would be a breaking change. And then the question is how do we handle it for custom validations? I though about using a template string literal type, so that we could enforce e.g. a strict prefix (e.g. `custom_${string}`), but this still doesn’t give us a type safe `requirement` for custom validations 🤔.

fabian-hiller:
Thanks for your research! Unfortunately, I don't have a solution for this either. Breaking changes are ok as long as we are still in v0. Being able to write your own validation functions and name them in issues seems important to me for Valibot.

[...]

Issue #23, opened 2023-07-28
Valibot to JSON Schema


huypham50:
Will there a native `valibot ↔ json schema` conversion?

I think this is what missing from `zod` (you have to use `zod-to-json-schema`, which is a third party package with some [known issues](https://github.com/StefanTerdell/zod-to-json-schema#known-issues))

Whereas in the python ecosystem, `pydantic` has [native support](https://docs.pydantic.dev/latest/usage/json_schema/) to generate json schemas from models.

Lovely work so far!

shakyShane:
+ 1

but, I would also enjoy the other way around - json-schema to valibot schemas - perhaps we can leverage existing tools around zod/arktype for this...

fabian-hiller:
It's on my list to check out a possible implementation. Since I'm currently writing my bachelor's thesis, I won't get to it until September.

noxify:
Maybe https://github.com/sinclairzx81/typebox could be used as inspiration?

gcornut:
There is one existing solution to convert from valibot to json schema. You can generate json schema from the TS types with [`ts-json-schema-generator`](https://github.com/vega/ts-json-schema-generator). 
Of course it's not a perfect solution because **some of the features of valibot do not translate to TS types** but would translate to JSON schema (like `minLength` to constrain array size).

**An example on how to use this:** 
1. define a schema
> src/schema.ts
```
import { object, string, type Input } from 'valibot';

const FooElementSchema = object({ type: string('foo') });
export type FooElement = Input<typeof FooElementSchema>;
```
2. then execute `npx ts-json-schema-generator --path src/schema.ts -t FooElement -o dist/schema.json`

**The result:**

```json
{
  "$ref": "#/definitions/FooElement",
  "$schema": "http://json-schema.org/draft-07/schema#",
  "definitions": {
    "FooElement": {
      "additionalProperties": false,
      "description": "Input inference type.",
      "properties": {
        "type": {
          "description": "Input inference type.",
          "type": "string"
        }
      },
      "required": [
        "type"
      ],
      "type": "object"
    }
  }
}
```

fabian-hiller:
[...]

PR #211, opened 2023-10-09
Add Schema Reflection API

Saeris:
This PR is a proposal implementation for a schema reflection API submitted for consideration and refinement.

Adding a reflection API would enable development of third-party utility libraries for Valibot such as a mock generation library similar to [`@anatine/zod-mock`](https://github.com/anatine/zod-plugins/tree/main/packages/zod-mock), which is what initially inspired this effort.

At present, given Valibot's architecture utilizing anonymous arrow functions as the return values of Validation functions and the lack of access to the given Pipe in the Schema object, it would be difficult if not entirely impossible to gather the required metadata about a Valibot schema definition in order to generate mock data that would pass schema validation. This is because many of the required values, such as function arguments and internal variables, cannot be accessed at all once defined.

In this PR I've added what I believe to be a minimal API which would expose this Schema and Pipe metadata. Let's consider some examples.

First, let's take a look at the Email validation function, which doesn't accept a user-defined requirement:

```ts
  const kind = 'email' as const;
  const requirement =
    /^[\w+-]+(?:\.[\w+-]+)*@[\da-z]+(?:[.-][\da-z]+)*\.[a-z]{2,}$/i;
  const message = error ?? ('Invalid email' as const);
  return Object.assign(
    (input: TInput): PipeResult<TInput> =>
      !requirement.test(input)
        ? getPipeIssues(kind, message, input)
        : getOutput(input),
    {
      kind,
      requirement,
      message,
    }
  );
}
```
Note that `kind`, `requirement`, and `message` were all extracted to variables to keep the internals DRY, and that `Object.assign()` is being used to assign each of these values as static properties of the returned validator function.

This exposes two of the most critical pieces of information about a validation check: what kind of check is being performed and what condition needs to be met for success.

Since the function being returned here is anonymous, even if we had access to the Pipe array in a Schema, each Pipe Item wouldn't have a `name` property to identify what validation was used. Additionally, because of the security limitations of strict mode, this function's `arguments` property also cannot be accessed.

With this setup, we're safely exposing all of this information in a manner which these values can be accessed without risk of modifying the runtime behavior of the validation itself.

> [!NOTE]
>
> Originally I had considered wrapping this "metadata" object in `Object.freeze()` to ensure that it was read-only at runtime, but I don't think it provides much benefit for the performance tradeoff.

Let's take a look at one more Validation example:

```ts
export function minLength<
  TInput extends string | any[],
  const TRequirement extends number
>(requirement: TRequirement, error?: ErrorMessage) {
  const kind = 'min_length' as const;
  const message = error ?? ('Invalid length' as const);
  return Object.assign(
    (input: TInput): PipeResult<TInput> =>
      input.length < requirement
        ? getPipeIssues(kind, message, input)
        : getOutput(input),
    {
      kind,
      requirement,
      message,
    }
  );
}
```
User-defined requirements are roughly the same except for one important detail: `const TRequirement extends number`, where the use of the `const` modifier on this generic will instruct Typescript to treat the user-supplied value as narrowly as possible. This means that the return type of `minLength()` will show the user-defined values it was supplied rather than a wider `requirement: number` type.

> [!NOTE]
>
> For this initial implementation I wanted to introduce as few changes to the existing types as possible, so there's room for improvement still.

Next let's take a look at a Schema function:

```ts
export type StringSchema<TOutput = string> = BaseSchema<string, TOutput> & {
  schema: 'string';
  checks: PipeMeta[];
};

// ...

export function string(
  arg1?: ErrorMessage | Pipe<string>,
  arg2?: Pipe<string>
): StringSchema {
  const [error, pipe] = getDefaultArgs(arg1, arg2);

  return {
    schema: 'string',

    async: false,

    checks: getChecks(pipe),

    _parse(input, info) {
      // ...
    },
  };
}
```
New things here are the addition of the `checks` property, which always returns an array of `PipeMeta` objects, which are the values we added to each of the Validation functions. For every Schema that accepts a Pipe argument, a `checks` property was added in the return value to expose metadata about the Pipe array.

Notably every Transformation was untouched, so they don't provide any metadata. `checks` only contains metadata from Validation functions.

Last, here's another Schema example:

```ts
export type ObjectSchema<
  TObjectShape extends ObjectShape,
  TOutput = ObjectOutput<TObjectShape>
> = BaseSchema<ObjectInput<TObjectShape>, TOutput> & {
  schema: 'object';
  object: TObjectShape;
  /* entries: [key: PropertyKey, value: SchemaMeta][]; // Removed */
  checks: PipeMeta[];
};

// ...

export function object<TObjectShape extends ObjectShape>(
  object: TObjectShape,
  arg2?: Pipe<ObjectOutput<TObjectShape>> | ErrorMessage,
  arg3?: Pipe<ObjectOutput<TObjectShape>>
): ObjectSchema<TObjectShape> {
  const [error, pipe] = getDefaultArgs(arg2, arg3);

  let cachedEntries: [string, BaseSchema<any>][];

  return {
    schema: 'object',

    object,

    async: false,

    /* entries: getEntries(object), // Removed */

    checks: getChecks(pipe),

    _parse(input, info) {
        // ...
    },
  };
}
```
> [!NOTE]
>
> After further exploration writing a mocking library based on this PR, I no longer think the `entries`  prop / `SchemaMeta` type are necessary, so I've removed them.

~~For iterable schemas (`intersection`, `object`, `tuple`, and `union`), there's also a new `entries` property, which either returns an array of key/value pairs, or simply a `SchemaMeta[]` array. This property helps with introspection on nested schema types.~~

~~`SchemaMeta` varies by schema, but it breaks down like this:~~ This is what the `PipeMeta` object looks like:

```ts
export type PipeMeta<
  TKind extends string = string,
  TRequirement = unknown
> = Readonly<{
  kind: TKind;
  message: ErrorMessage;
  requirement?: TRequirement;
}>;

/* Removed
export type SchemaMeta =
  | {
      schema: string;
    }
  | {
      schema: 'literal';
      literal: string | number | bigint | boolean | symbol;
    }
  | {
      schema: 'any' | 'array' | 'map' | 'record' | 'set' | 'special';
      checks: PipeMeta[];
    }
  | {
      schema: 'intersection' | 'union';
      entries: SchemaMeta[];
    }
  | {
      schema: 'tuple';
      entries: SchemaMeta[];
      checks: PipeMeta[];
    }
  | {
      schema: 'object';
      entries: [key: PropertyKey, value: SchemaMeta][];
      checks: PipeMeta[];
    };
   */
```
These are all the fields which comprise the reflection API. I think there's room for some further simplification and improvements to the type definitions. Names for properties/types aren't set in stone, so feedback on those is welcomed and expected.

In this first pass I went and added a bunch of new tests to most of the affected schemas to validate the property additions and the a utility function `getChecks()` ~~and `getEntries()`~~.

There may be a few things I missed our haven't considered too. My goal here was adding just enough information about schemas to feed into a library such as Faker to generate mock data for tests, which is what I'm currently doing for my `zod` based schemas in [`@discordkit/client`](https://github.com/discordkit/discordkit/). I can see this also being useful for other use cases as well, such as JSON Schema or OpenAPI spec code generation tools.

Finally I want to acknowledge that I understand that this project is part of your Bachelor's thesis and that there wasn't a contributing guide that I could find, so I understand if you don't have much time for reviewing PRs or would prefer to pursue your own implementation. This was a fun weekend project for me and at the end of the day I'm just hoping this is more inspiring than just opening another feature request issue.


fabian-hiller:
Thanks a lot for this PR. Unfortunately, I don't have time to read everything right now. However, [this comment](https://github.com/fabian-hiller/valibot/issues/191#issuecomment-1753275007) I wrote this morning seems to be related. As soon as I find the time, I will read everything carefully and answer you in detail.

library/src/validations/ip/ip.ts
```diff
@@ -9,11 +9,19 @@ import { getOutput, getPipeIssues } from '../../utils/index.ts';
...
-      input
-    )
-      ? getPipeIssues('ip', error || 'Invalid IP', input)
-      : getOutput(input);
+  const kind = 'ip' as const;
+  const requirement =
+    /^((((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4})|((([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))))$/;
```
Saeris, line 14: Here I combined the regex expressions into a single value. It's a little hard to read but essentially it's just introducing an `|` OR between the two so that this can be exposed as a single requirement value instead of two.

Saeris:
Just wanted to pop in with an update:

Since opening this I started working on a refactor of [`@anatine/zod-mock`](https://github.com/anatine/zod-plugins/tree/main/packages/zod-mock) to work with this new API and I have some great coverage of most types already!

In working on this I realized that the `entries` property isn't necessary and so after a little more testing with this mocking library implementation, I'll probably simplify this PR a bit to remove the unnecessary bits.

fabian-hiller:
Thanks a lot for your contribution! I will answer you by the weekend at the latest.

Saeris:
No rush!

fabian-hiller:
Can you comment on my thoughts in [this comment](https://github.com/fabian-hiller/valibot/issues/191#issuecomment-1753275007)? Instead of a function with properties, what do you think about choosing a structure similar to the schemas using an object with a `._parse` method?

fabian-hiller:
This is my answer on https://github.com/fabian-hiller/valibot/issues/191#issuecomment-1758859954

If a validation function returns an object with `._parse`, this has the advantage that the implementation and structure is uniform with our schemas. The other advantage I see is that `.checks` is not needed this way and therefore we can speed up the initialization of a schema. Instead of `.check` I would add the property `.pipe`, which is 1 to 1 the same as the pipeline passed as an argument. This way no additional calculations are needed. `.pipe` has almost the same content as `.checks`. Basically, we have the same idea, just with two different implementations.

According to my idea, `minLength` would then be implemented as follows:

```ts
export function minLength<TInput extends string | any[]>(
  requirement: number,
  message: ErrorMessage = 'Invalid length'
): MinLengthPipe<TInput> {
  return {
    kind: 'validation',
    name: 'min_length',
    requirement,
    message,
    _parse(input) {
      return input.length < requirement
        ? getPipeIssues('min_length', message, input)
        : getOutput(input);
    },
  };
}
```

Calling `string([minLength(10)])` would then return the following data structure:

```ts
const schema = {
  kind: 'schema',
  name: 'string',
  async: false,
  pipe: [{
    kind: 'validation',
    name: 'min_length',
    requirement: 10,
    message: 'Invalid length',
    _parse(input) {
      // ...
    },
  }],
  _parse(input, info) {
    // ...
  },
};
```

This is just a first idea. I have not made a decision yet and am open to any ideas.

Saeris:
I think I'd rather go the opposite direction, get rid of `_parse()` entirely and adopt the `Object.assign()` format across the library.

One thing I'm concerned about is that `_parse` doesn't compress well because it's a property name. If the default behavior of each the Schemas, Transforms, Validations, etc was to be invoked as a function, that would shave additional bytes off a compiled app using `valibot`. Additional space savings could be had by aliasing `Object.assign()` behind a utility function, let's call it `assign()`. However, I'm not an expert on this subject, I just know that mangling property names is considered dangerous.

There's also to consider how much refactoring would be required by either approach. It would seem to me like with your proposal it would necessitate additional refactors to every Transform to have a consistent Pipe API. In setting out with this PR, I aimed to be purely additive and not introduce breaking changes to the existing API. So if possible, I'd like to keep the scope of changes here to a minimum.

---

I did some reading on JSON schema to get a better sense of how this might relate to #23, and I don't think that it's a good idea to tightly tie the metadata properties of Schemas, Validations, Transforms, etc to the JSON schema format, but rather pursue writing a schema serializer. You can read up on the JSON schema spec [here](https://json-schema.org/draft/2020-12/json-schema-validation#name-introduction)

What is important to keep in mind, I think, is to avoid the use of `type`, `title`, and `description` in the metadata for these functions, as they are keywords with specific meaning inside of JSON schema definitions. This also relates to the naming of various validation functions, as they could have 1:1 matches with the validation keywords that JSON schema uses.

One such example is `maxLength`, which in Valibot applies to both `string`s and `array`s, but in JSON schema these are divided into `maxLength` and `maxItems`. Additional thought on how to align those sorts of differences might be better explored in a different PR, but in theory a serializer can take the existing validations and translate them to the appropriate JSON schema validation constraints. In either case, a serializer would need to work on a Schema and translate it's `pipe` into the appropriate validation keywords. I would not attempt to set those properties on schemas themselves, as I think it would just make them heavier.

---

I think I could get behind exposing the entire `pipe` instead of `checks`, but my concern is that because `Pipe` can be either a Validation or a Transform, the user will then have to filter `pipe` on their end. That wouldn't be that difficult to do, just something worth considering.

However, I don't think repurposing `kind` here to distinguish between `schema`, `validation`, `transform` etc is necessary, because that can be inferred via the shape of the return type. TypeScript on its own can already tell us what we're working with because for example, all schema generators return a type that extends `BaseSchema`. To me this doesn't seem important at runtime, when really this is metadata that is mainly useful at authoring time.

---

For now, this is where my thoughts are on where to go from here.

tl;dr is that I'd prefer to keep this PR narrowly scoped and save larger refactors for another PR, and that we should be mindful of the tradeoffs at hand

fabian-hiller:
> One thing I'm concerned about is that `_parse` doesn't compress well...

I think `._parse` can be compressed very well, because the word "parse" also appears in the `parse` and `safeParse` methods, and these are needed to use a schema. Also, I suspect it takes less bytes than `Object.assign()`. In practice, however, the difference will be so small that it will not matter in the end, and it is not a deciding point for me.

> There's also to consider how much refactoring would be required by either approach...

We won't be able to prevent breaking changes while we're still in v0 and working on major features. I don't care about the effort either, because the best end product is what matters to me.

> What is important to keep in mind, I think, is to avoid the use of `type`, `title`, and `description` in the metadata for these functions, as they are keywords with specific meaning inside of JSON schema definitions...

Why would you consider the JSON schema definition? It doesn't matter for Valibot and doesn't play a role, or am I misunderstanding something here?
 
> One such example is `maxLength`, which in Valibot applies to both `string`s and `array`s, but in JSON schema...

I would develop Valibot completely independent of the JSON Schema definition, since serialization is not a relevant feature of Valibot, as it negates Valibot's advantages in terms of bundle size. If serialization and JSON Schema are important, I recommend using [TypeBox](https://github.com/sinclairzx81/typebox) instead of Valibot.

> I think I could get behind exposing the entire `pipe` instead of `checks`, but my concern is that because `Pipe` can be either a Validation or a Transform, the user will then have to filter `pipe` on their end. That wouldn't be that difficult to do, just something worth considering.

Yes, I agree, we should take that into account. However, since transformations are part of the Valibot pipeline, developers working with Valibot should be aware of this anyway. Exposing the pipeline as `.pipe` has the additional advantage that the pipeline can be modified afterwards, for example with methods like `removePipe` or `overwritePipe`.

> However, I don't think repurposing `kind` here to distinguish between `schema`, `validation`, `transform` etc...

Yes, with a schema it is not necessarily required. But for the pipeline I think it is, since transformations do not have `.requirement` and `.message`. Or do you have an alternative idea for this?

---

For now, the disadvantages of `Object.assign()` outweigh an object with `._parse` in my opinion:
- The code is harder for beginners to understand in usage and implementation
- If you can call a schema directly, we can't indicate that it's an internal API like we can with the underscore in `._parse'
- Also, the difference between `schema(input)` and `safeParse(schema, input)` is unclear, which makes the API more complicated for users
- Object.assign()` takes more CPU time than creating a simple JavaScript object
- Can TypeScript correctly type `Object.assign()`? If not, this is another drawback

Saeris:
Please take a look over the latest commit. I went and refactored everything to use a uniform return shape with a `_parse()` method.

I was going to continue advocating for `Object.assign()` but I spent some time this morning researching the performance impact it could possibly have, and while there might be when assigning a lot of properties or frequent execution, I figured it wasn't worth the risk. I did however do a refactor last night with this implementation if you want to compare to this branch: https://github.com/Saeris/valibot/tree/feat/return-parse-function-from-schemas (there were some test errors with `comparable()` that I punted on resolving, code works fine otherwise)

---

As far as my JSON schema comments go, you can ignore those. I assumed because of #23 that there was desire to support it somehow with this PR.

---

For reference, `Object.assign()` does work just fine with TypeScript. Under the hood it takes the types of each arg and returns an intersection of all of them. You can see for yourself in your IDE that the types come through exactly as expected.

---

> Yes, with a schema it is not necessarily required. But for the pipeline I think it is, since transformations do not have .requirement and .message. Or do you have an alternative idea for this?

I think it's rather simple to pattern match on unique properties to Schema, Validation, or Transform to determine what type they are. `getChecks()` has a utility function that does exactly this for Validations. `async` is unique to Schema, `message` is unique to Validation, etc.

Calling a night on contributions for now. Will catch up on any feedback tomorrow.

fabian-hiller:
Thank you for your contribution. I'll try to look at it tomorrow and think about it. I expect to give you feedback next week, with the goal of merging the improvements in the following days.

Saeris:
Here's the mocking library I'm working on to go with this PR, in case you'd like to try it out locally. https://github.com/Saeris/valimock

FYI I will be on vacation from the end of this week until the beginning of November, so if I'm slow to respond that is why.

fabian-hiller:
Can you try to sync the repo and merge the current main branch into your feature branch? I made a few structural changes to the object schema.

david-plugge:
I´m really excited for this! I´m working on a form validation library that is able to parse FormData objects using the provided valibot schema. With this feature it will also be able to generate html validation rules like

```html
<script>
  const schema = v.object({
    age: v.number([ v.minValue(5), v.maxValue(12) ])
  })
</script>

<!-- all properties can be generated by providing just the name -->
<input type="number" name="age" min="5" max="12"/>
```

david-plugge:
> What is the name of your library? I am interested in having a look at it.

I was going to name it valiform but the name is taken on npm. So i´m open to some suggestions!
The idea came up when i started using sveltekit-superforms in a project but wasn´t really satisfied with the dx and some decisions they made. That lib uses zod for validation and therefore teach you to only validate on the server.
With valibot that isnt really an issue due to its small bundle size. The idea is to use the same schema for both, server and client. My library will only bring a small abstraction with a few helpers to basicly reduce it to like 5 lines of code + configuration. This PR would make it possible to even generate native html form validation in cases where js doesnt work.

I will post the link to the repo when i´m done refactoring. The hardest part was to figure out how to efficiently extract the schema types and map them to the html input types eg transform `FormDataEntryValue` to the correct type (`text` and `number` is easy, but what about enums and other more complicated types?).

For now it will only work one layer deep, so no nested objects yet. I have some ideas to work around that limitation but we will see.

Another possible thing to add is the ability to get a json representation to build your html as your schema defines it.

fabian-hiller:
@david-plugge have you taken a look at my form library [Modular Forms](https://modularforms.dev/)? I plan something similar for all signal based form libraries and that includes Svelte 5. I am interested in an exchange on this.

library/src/schemas/any/any.ts
```diff
@@ -17,24 +21,10 @@ export type AnySchema<TOutput = any> = BaseSchema<any, TOutput> & {
...
-    schema: 'any',
-
-    /**
-     * Whether it's async.
-     */
+    kind: 'any',
     async: false,
+    pipe,
 
-    /**
```
fabian-hiller, line 30: Is there a reason why you removed this comment everywhere?

library/src/schemas/any/any.ts
```diff
@@ -5,7 +5,11 @@ import { executePipe } from '../../utils/index.ts';
  * Any schema type.
  */
 export type AnySchema<TOutput = any> = BaseSchema<any, TOutput> & {
-  schema: 'any';
+  kind: 'any';
+  /**
+   * Validation and transformation pipe.
```
fabian-hiller, line 10: For types, I don't think these inline comments are necessary.

library/src/schemas/bigint/bigint.ts
```diff
@@ -36,28 +40,13 @@ export function bigint(
...
   return {
-    /**
-     * The schema type.
-     */
-    schema: 'bigint',
-
-    /**
-     * Whether it's async.
-     */
+    kind: 'bigint',
```
fabian-hiller, line 47: Would you prefer `kind` to `type`? If so, for what reason?

library/src/schemas/bigint/bigintAsync.ts
```diff
@@ -42,28 +46,13 @@ export function bigintAsync(
   arg2?: PipeAsync<bigint>
 ): BigintSchemaAsync {
   // Get error and pipe argument
-  const [error, pipe] = getDefaultArgs(arg1, arg2);
+  const [error, pipe = []] = getDefaultArgs(arg1, arg2);
```
fabian-hiller, line 49: Maybe we should also allow `undefined` for `pipe` to avoid creating an object unnecessarily.

fabian-hiller (commented):
Thank you again for your contribution.

library/src/schemas/bigint/bigintAsync.ts
fabian-hiller: Also I would prefer to outsource this to `getDefaultArgs` if we keep that, instead of adding it everywhere.

library/src/schemas/any/any.ts
Saeris: All of the comments for the return types were moved to the base interface in the `types.ts` file. IntelliSense surfaces it everywhere in the codebase, so it's redundant to have them in each schema definition.

For properties specific to individual schemas, the comments relevant to those properties were moved to the type definition at the top of the file.

library/src/schemas/bigint/bigint.ts
fabian-hiller: If think that `type` is more generic and also reduces the bundle size a bit. However, the bundle size is negligible here. I am mainly interested in good naming.

Saeris: I think `kind` would be better so that it doesn't have a potential conflict if someone is trying to build out a JSON schema library on top of this.

It's also what I've seen commonly used to differentiate similar types in discriminated unions, which is basically what we have here to refer to all schemas/validations/transforms that extend a common base class.

library/src/schemas/bigint/bigintAsync.ts
Saeris: Either/or, I don't think it matters too much. My personal preference is to reduce the number of existence checks.

Saeris: And I'm open to moving it to the function itself. When I was writing all of this, I ran into type errors trying to modify `getDefaultArgs` so this was the least friction for me at the time.

library/src/schemas/bigint/bigint.ts
Saeris: If anything, this would be better read as `type: "schema", kind: "bigint"` for example. Where `type` refers to the base type we're extending from, and `kind` refers to the sub-type.

fabian-hiller: I understand the JSON argument, but would not consider it for Valibot as it seems too irrelevant at the moment. I think `kind` is fine and am happy with either name. I just thought `type` was more commonly used for differentiation. We can continue with `kind` and update if needed before v1 if we have better ideas or arguments.

library/src/schemas/any/any.ts
fabian-hiller: Ok, you are right, for documentation and tooling with IntelliSense it is actually better this way. I had added the comments to the return object of the schemas at that time to make the code more understandable when looking at a single schema file on GitHub. How would you proceed here?

library/src/schemas/any/any.ts
fabian-hiller: As a result of your comment, I have changed my mind on this. Thank you for this improvement. However, I would prefer to add a blank line before a comment that starts with `/**` for a better overview.

library/src/schemas/bigint/bigint.ts
fabian-hiller: > If anything, this would be better read as `type: "schema", kind: "bigint"` for example. Where `type` refers to the base type we're extending from, and `kind` refers to the sub-type.

That's a good argument!

library/src/schemas/bigint/bigintAsync.ts
fabian-hiller: Then I recommend to remove it completely and allow `undefined`. I think this is the best way to proceed for now. It might also make sense to add `TPipe` to the generics, so that the `.pipe` property is 1 to 1 the same as the `pipe` argument. However, we could also look at the generic thing in a future PR to reduce this on to the essentials.

library/src/utils/getChecks/getChecks.ts
```diff
@@ -0,0 +1,18 @@
+import type { Pipe, PipeAsync, PipeMeta } from '../../types';
+
+const hasPipeMeta = (val: unknown): val is PipeMeta =>
+  typeof val === `object` && val !== null && `kind` in val && `message` in val;
+
+export const getChecks = (pipe?: Pipe<any> | PipeAsync<any>) =>
```
lo1tuma, line 6: Wouldn’t `unknown` work instead of `any`? As far as I understand, there is a type-guard checking the value 🤔

library/src/types.ts
```diff
@@ -138,17 +172,72 @@ export type PipeResult<TOutput> =
...
+  TKind extends string = string,
+  TMessage extends ErrorMessage = ErrorMessage,
+  TRequirement = unknown
+> = {
+  /* The validation type */
+  kind: TKind;
+  /* The validation error message */
+  message: TMessage;
+  /* The requirement value to compare against (Optional) */
+  requirement?: TRequirement;
```
lo1tuma, line 186: I an optional `requirement` field really a good idea? If I implement a validation function that always sets the `requirement` field it is somehow awkward to always add a guard just because the type is not precise enough. If someone wants to have a validation function without a `requirement` field it might be better to explicitly set it to `undefined` or `null`. WDYT?

david-plugge:
> @david-plugge have you taken a look at my form library [Modular Forms](https://modularforms.dev/)? I plan something similar for all signal based frameworks and that includes Svelte 5. I am interested in an exchange on this.

While it looks quite nice for spa's i can't find info about server side validation. My main focus is to unify the client and server side validation and typesafety in meta frameworks (focusing on sveltekit for the moment).

fabian-hiller:
If you are interested, you can find more information about server-side validation [here](https://modularforms.dev/qwik/guides/handle-submission) and in the [Qwik docs](https://qwik.builder.io/docs/integrations/modular-forms/). I plan to rework Modular Forms in early 2024 to make this even easier.

library/src/schemas/any/any.ts
Saeris: I don't think it's necessary to add the extra comments everywhere. The codebase itself is relatively straight-forward to navigate. Once you have documentation added to the website it will make it easier to understand that patterns each of these functions follow as well.

fabian-hiller: Thank you for your feedback. I will think about it and then make a decision.

library/src/schemas/bigint/bigint.ts
fabian-hiller: I have been reading about the difference between `kind` and `type`. Several sources have indicated that `kind` refers to a group with certain attributes, while `type` refers to a specific item. So I prefer to continue with `type`.

fabian-hiller:
v0.20.0 brought a lot of structural changes. For example, I changed the `schema` property of a schema to `type` and nested properties like `map: { key, value }` in `map` to a flat hierarchy with just `key` and `value`. A merge is likely to introduce various conflicts. It may be faster to make the changes independently on the branch before merging. Here are the release notes: https://github.com/fabian-hiller/valibot/releases/tag/v0.20.0

fabian-hiller:
I plan to publish the improvements of this PR with the next version.

library/src/schemas/any/any.ts
Saeris: I'll make a pass over everything tonight to try and clean up the formatting.

library/src/schemas/bigint/bigintAsync.ts
Saeris: I'll take a look at implementing this generic tonight, though I agree it would be a good follow-up PR.

library/src/utils/getChecks/getChecks.ts
Saeris: I tried using `unknown` in a few places but ended up getting type errors so I punted on doing extra work. With the API still in flux ahead of a v1, I think it's safe to push those refinements out to a later PR.

library/src/schemas/any/any.ts
fabian-hiller: I have thought about this and would like to keep the comments for now, as I think they make the code more understandable.

Saeris:
v0.20.0 should be all merged up! Tests are passing, so looks good to me.

I made all of the `pipe`s optional except for `any`, which defaults to an empty array for its single argument.

Went and renamed `kind` to `type` everywhere, so should be good to go on that front.

Please give it a review for any cleanup that is missing so we can get this shipped!

library/src/schemas/any/any.ts
Saeris: I would advise against it. You're adding unnecessary maintenance burden to yourself to repeat these comments throughout the codebase. IntelliSense will surface the doc comments from the root `BaseSchema` type when a contributor is editing any of the schemas.

fabian-hiller: The reason I want to keep the comments is mainly so that people looking at a single schema on GitHub can immediately understand where to find what. The effort on my part is not high, as new APIs are added only occasionally. I would rather have one comment too many than one too few.

fabian-hiller:
Thanks @Saeris, I appreciate the work you put into this PR! I will try to review your code in the next few days.

fabian-hiller:
Short update: I try to review everything tomorrow or on the weekend. The goal is to release a new minor version next week.

fabian-hiller:
I have started to investigate and improve the code. I'll let you know as soon as I'm further along.

fabian-hiller:
I expect to share my changes here in the middle or end of this week so we can discuss the final details.

fabian-hiller:
@Saeris and @lo1tuma can review my changes?

Saeris:
I took a look over your latest commits (da3ca63, 775d539 and 5838643) and everything looks good to me.

A concern I have with `_parse` always taking the returned values for options like `pipe` and others is that they are no longer statically analyzable, because runtime logic could change their values. But that would be an exception rather than the norm, so I suppose it's okay because it's really a "don't change this unless you know what you are doing" kind of feature. Generally, I don't like this kind of unpredictability in something like a data validation schema, but then it's also not a feature I see myself using.

But otherwise, I like the addition of exposing error messages as well as the validation pipelines. That should help in form input validation.

I'm good to ship this if you are! Thanks for all the effort in reviewing @fabian-hiller!

fabian-hiller:
Thanks a lot for the work you put into this PR! I will wait 1 to 3 days and then do the final review to merge the changes.

lo1tuma:
I had a brief look at the changes. I still think that `requirement` itself is not typed is a bummer. If this is only a first iteration and there are plans to improve the precision of its type I’m fine. But as it is right now I would probably not use it.

fabian-hiller:
@lo1tuma can you describe in detail what you mean and how you would improve it?

lo1tuma:
What I mean is the problem that  on the `Issue` type `requirement is defined as an optional of `unknown`. So when I want to analyze an issue like this:

```
if (issue.type === 'stars_with') {
    someStringOperation(issue.requirement); // requirement is unknown but I would expect type string
}
```

I would need to do further runtime checks to verify what type `requirement` has. So ideally `Issue` could be a discriminated union. But right now I think it is hard to implement because we also have custom validations and therefore arbitrary values of `issue.type`. One idea would be not allow custom validations to overwrite `type` and make issue.type a known value in this case, for example `custom`.
I think this is a bit tricky to solve and there might be other solution, so maybe it is better to do it as a follow-up.

fabian-hiller:
Yes, I agree that we should address this in a follow-up. I plan to integrate Valibot more into Modular Forms and will probably find a lot of necessary improvements to Valibot along the way.

fabian-hiller:
I will think about your feedback and probably create an issue after this one is merged so that we do not forget about this in the long run.

fabian-hiller:
Thank you again for contributing to this PR!

ciscoheat:
This looks very promising, especially with my idea that [Superforms](https://superforms.rocks/) v2 should be able to use most validation libraries through [typeschema](https://github.com/decs/typeschema), in more or less complete versions of the validation format, where Zod is the benchmark, being fully introspectable.

I would be happy to use Valibot as the test library for experimentation, so is there an example how to use the Reflection API? This is how I do it with Zod: https://github.com/ciscoheat/sveltekit-superforms/blob/main/src/lib/schemaEntity.ts#L293-L358

It's not difficult, once the metadata is easily accessible. So an example of that, together with how to be able to access some other properties like nullable, optional, default value, would get me very far.

fabian-hiller:
I think the code for Valibot is very similar to the code for Zod. The biggest difference will be that you extract all the details of a data type from the pipeline. I have created an issue so that we add a guide on this to the documentation in the long run.

Issue #203, opened 2023-10-07
Change `.reduce` to for...of loop to improve performance


fabian-hiller:
In several places, for example in `partial`, we currently still use `.reduce`. For performance reasons, it may make sense to change this to a for...of loop. Since we use for...of in several other places in the library, this change should not affect the bundle size due to compression and should make the code more consistent.

Idea for this change comes from @kurtextrem in #202

lo1tuma:
Is the usage of `.reduce` really a bottleneck in applications that use `valibot`? I usually recommend to avoid micro-optimizations, especially when it is about language features, because the performance could change with every new release of the engine (e.g. `v8`).  
So I think a data-driven approach would be better. E.g. having a benchmark suite, which tests the real-world usage of valibot. This could then be used to identify the performance bottlenecks. Optimizing something that is not the bottleneck is usually waste or in the worst case it increases the complexity or maintainability of the codebase without any benefits.

fabian-hiller:
Thank you for your feedback! I will take it into consideration. We plan to add performance benchmarks to our CI in #100.

PR #213, opened 2023-10-15
chore: add eslint unicorn plugin

kurtextrem:
This PR is based on #202.

Unicorn really brings some advantages, such as auto optimizing regexps (makes them smaller) and sets rules for writing or using regexps in the most efficient way. https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-reduce.md?rgh-link-date=2023-10-07T09%3A37%3A41Z might also be interesting (for e.g. #203)

### DIFF
--- a/library/.eslintrc.cjs
+++ b/library/.eslintrc.cjs
@@ -8,7 +8,7 @@ module.exports = {
     'plugin:security/recommended',
   ],
   parser: '@typescript-eslint/parser',
-  plugins: ['@typescript-eslint', 'import', 'redos-detector'],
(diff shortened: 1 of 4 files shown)


fabian-hiller:
I will review this PR after we merged #202.

kurtextrem:
Good point. I think we briefly discussed that back then and preferred having one rule over many that improve regexps, but if it contains bugs, we should go for eslint-plugin-regexp (at Framer we're also using that). Wdyt Fabian?

PR #214, opened 2023-10-17
Error on invalid dates

sillvva:
One thing missing from the `date` and `dateAsync` schemas is `Invalid Date` handling. With `new Date("Not a date")` you get `Invalid Date`, but `dateObj instanceof Date` returns `true`. 

This is similar to the `NaN` issue, which is already handled by the `number` schema. 

A fix for this is to test whether `dateObj.getTime()` returns `NaN`.

Playground Demo: https://playcode.io/1631384


library/src/schemas/date/date.ts
```diff
@@ -60,7 +60,7 @@ export function date(
      */
     _parse(input, info) {
       // Check type of input
-      if (!(input instanceof Date)) {
```
fabian-hiller, line 63: Can you check if `isNaN(…)` also works to save some bytes?

library/src/schemas/date/date.ts
```diff
@@ -60,7 +60,7 @@ export function date(
...
+      if (!(input instanceof Date) || Number.isNaN(input.getTime())) {
```
fabian-hiller, line 63: Can you check if `isNaN(…)` also works to save some bytes?

fabian-hiller:
Thank you for noticing this issue!

library/src/schemas/date/date.ts
sillvva: It does. I was just following the  number schema example.

fabian-hiller: Theoretically, `Number.isNaN(...)` is safer because it does not transform the value into a number. However, in both cases it makes no difference because we already know that the value is a number or `NaN`. So it does not matter how we do it. We should just be consistent. What do you think is better?

sillvva: I agree. I think because we know it's a number or NaN, isNaN is better to save bytes. I have added a commit to the PR.

fabian-hiller:
v0.20.0 is now available

PR #219, opened 2023-10-20
Add export for any validation regex

fabian-hiller:


library/src/utils/isLuhnAlgo/isLuhnAlgo.ts
```diff
@@ -7,7 +12,7 @@
  */
 export function isLuhnAlgo(input: string) {
   // Remove any non-digit chars
-  const number = input.replace(/\D/gu, '');
+  const number = input.replace(NON_DIGIT_REGEX, '');
```
kurtextrem, line 15: needs `replaceAll` here (could be automatically captured by https://github.com/fabian-hiller/valibot/pull/213/files `unicorn/prefer-string-replace-all` I think.

reason is: a global regex (`/g`) will set `lastIndex` to the last index where the regex stopped. Which means if string a is "abcdef" and string b is "a", lastIndex will be set to e.g. 3, but string b has no index 3, so the regex now fails at all times.
If you want to avoid `replaceAll`, you'll need to call
`NON_DIGIT_REGEX.lastIndex = 0` before calling the replace

library/src/regex.ts
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * [IMEI](https://en.wikipedia.org/wiki/International_Mobile_Equipment_Identity) regex.
+ */
+export const IMEI_REGEX = /^\d{2}(?:[ /|-]?\d{6}){2}[ /|-]?\d$/u;
+
+/**
+ * [IPv4](https://en.wikipedia.org/wiki/IPv4) regex.
+ */
+// eslint-disable-next-line redos-detector/no-unsafe-regex -- false positive
+export const IPV4_REGEX = /^(?:(?:25[0-5]|(?:2[0-4]|1\d|[1-9])?\d)\.?\b){4}$/u;
```
kurtextrem, line 26:
```suggestion
export const IPV4_REGEX = /^(?:25[0-5]|(?:2[0-4]|1\d|[1-9]|)\d)(?:\.(?:25[0-5]|(?:2[0-4]|1\d|[1-9]|)\d)){3}$/u;
```

fabian-hiller:
v0.20.0 is now available

Issue #76, opened 2023-08-07
How to control 2 fields equal or not


dukeofsoftware:
in my case password and confirm password have to be equal but I can't find which method I have to use

dukeofsoftware:
I tried this with react-hook-form but I don't get " the passwords do not match." error

fabian-hiller:
Have a look at this comment: https://github.com/fabian-hiller/valibot/issues/5#issuecomment-1660056243

fabian-hiller:
I will try to improve the DX for this use case before v1.

Demivan:
@fabian-hiller, `Zod` uses the same `pipeline` (`refine` in case of Zod) functionality. The problem with this approach is that the entire schema needs to be valid before the pipeline is executed.
This is fine when you need to check whether the entire data is valid, but is inconvenient for use in forms. Only when the user fills all fields correctly, will he see a `"Passwords do not match."` validation error.

I have added PR to implement `partialRefine` method to Zod (https://github.com/colinhacks/zod/pull/2360), but it is not merged yet. Would be nice if `valibot` supported this, this is often overlooked by validation libraries.

fabian-hiller:
Do you have any ideas on how we could integrate this into Valibot's API design?

Demivan:
Few ideas:
* allow defining pipelines that are run on raw input. Input would have `unknown` type, so the developer would need to check everything they want to access.
* allow pipelines on partially validated input (for object schemas), like Zod's `partialRefine`. This would work like a combination of `pick` plus a pipeline. The benefit of this is that the pipeline can have a strongly typed input.
* `Yup` has [ref](https://github.com/jquense/yup#refpath-string-options--contextprefix-string--ref) to access sibling fields. Don't really like this solution - it is not typed and really breaks current Valibot's API design.

Demivan:
@fabian-hiller after some consideration, I think the cleanliest solution in terms of API would be something like this, inspired by Zod:

```ts
const schema = object({
  name: string(),
  password: string(),
  confirmPassword: string(),
}, [
  {
    type: 'pre', // Run before the validation
    transform: (data: unknown) => data.password === data.confirmPassword, // Input will be unknown
  },

  {
    type: 'post', // Run after validation (default)
    transform: (data: Typed) => data.password === data.confirmPassword, // Input will be typed

    // Allow specifying path and error message instead of throwing validation error
    path: ["confirmPassword"],
    message: "Both password and confirmation must match",
  },

[...]

PR #223, opened 2023-10-24
Add a simple customPartial implementation

Demivan:
Related #76

This is an extremely simplified implementation of the proposed “partial pipe” solution.

### DIFF
--- a/library/src/schemas/object/object.ts
+++ b/library/src/schemas/object/object.ts
@@ -155,6 +155,9 @@ export function object<
       let issues: Issues | undefined;
       const output: Record<string, any> = {};
 
+      const partialPipes = pipe?.filter(p => p.partial != null)
+      const validFields: string[] = []
+
       // Parse schema of each key
       for (const [key, schema] of cachedEntries) {
         const value = (input as Record<string, unknown>)[key];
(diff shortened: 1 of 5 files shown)


fabian-hiller:
Thank you for the draft!

fabian-hiller:
Do you have any alternative name ideas for `customPartial`. Will it only work or make sense in the pipe of `object`?

fabian-hiller:
Beside PR #211, I would like to focus on a solution to this problem in the next week. It will also be important to be able to specify the path where the error message should be applied.

Demivan:
> Do you have any alternative name ideas for `customPartial`. Will it only work or make sense in the pipe of `object`?

Maybe something like `whenValidated` or `whenFieldsValid`. 
It only makes sense with `object` pipe. So would be nice to have type system prevent usage with other validations.

fabian-hiller:
I have added a `partialCheck` action and think that we can close this PR. Here is a code example:

```ts
import * as v from 'valibot';

const RegisterSchema = v.pipe(
  v.object({
    email: v.pipe(
      v.string(),
      v.nonEmpty('Please enter your email.'),
      v.email('The email address is badly formatted.'),
    ),
    password1: v.pipe(
      v.string(),
      v.nonEmpty('Please enter your password.'),
      v.minLength(8, 'Your password must have 8 characters or more.'),
    ),
    password2: v.string(),
  }),
  v.forward(
    v.partialCheck(
      [['password1'], ['password2']],
      (input) => input.password1 === input.password2,
      'The two passwords do not match.',
    ),
    ['password2'],
  ),
);
```

PR #249, opened 2023-11-10
fix(deno): avoid Node-style import resolution

BastiDood:
This PR fixes the Deno version of this library by preferring `*.ts` imports rather than relying on the more implicit Node module resolution algorithm.


fabian-hiller:
Thank you for noticing that. Normally we should have an ESLint rule that checks this so that it doesn't happen in the first place. I'll have to check what's going on.

PR #220, opened 2023-10-21
docs: add string()

kazizi55:
Implements https://github.com/fabian-hiller/valibot/issues/209#issuecomment-1755624961

PR #252, opened 2023-11-14
docs: add contributor section

kazizi55:
Implements: https://github.com/fabian-hiller/valibot/pull/220#issuecomment-1793213310


kazizi55:
This is how contributor section looks:

スクリーンショット 2023-11-14 9 36 53

website/contributors.ts
```diff
@@ -0,0 +1,116 @@
+import { fetch } from 'undici';
```
kazizi55, line 1: badge
Copied from [Qwik doc](https://github.com/builderio/qwik/blob/1d71aec752b5122512e13044d3fe47092d6e92a4/packages/docs/contributors.ts#L1-L95), and modified a littile bit (e.g. changing directory path and dealing with github personal token)

kazizi55: To add contributors to index.mdx, we can simply execute `pnpm contributors` 😄 
https://github.com/fabian-hiller/valibot/pull/252/files#diff-fae242fbf77a8a9d52625664bc36ea12316586f8a716b41137d897a2b7e3df76R19-R20

website/src/routes/guides/(get-started)/comparison/index.mdx
```diff
@@ -1,6 +1,9 @@
 ---
 title: Comparison
-description: Even though my API resembles other solutions at first glance, the implementation and structure of the source code is very different.
+description: >-
```
fabian-hiller, line 3: Whats the reason for this change?

fabian-hiller:
Thank you very much! I will review it after the next version of Valibot is released.

website/src/routes/guides/(get-started)/comparison/index.mdx
kazizi55: @fabian-hiller 
Sorry for my late reply 🙏 
This is because of the built-in formatter of stringify of [grey-matter](https://github.com/jonschlinkert/gray-matter/tree/master).
And it seems unable to change the format, because grey-matter uses [js-yaml](https://github.com/nodeca/js-yaml) internally to stringify, and cannot change the option of js-yaml, through grey-matter.

If you don't like this format, I'm gonna search other libraries 😄

fabian-hiller: Isn't Qwik already able to read metadata with `useDocumentHead().frontmatter`? Do we even need an additional library for this?

kazizi55: @fabian-hiller 
I think we need this library, as long as `pnpm contributors` (tsm contributors.ts) can be executed to add contributors in each docs.

By using `useDocumentHead().frontmatter`, Owik can only read metadata, but in contributors.ts, metadata about contributors is read and also written.

Should I delete contributors.ts? 👀

fabian-hiller (approved):
Thank you again for your contributions!

Issue #257, opened 2023-11-19
Consider migrating from Type to Interface


Saeris:
Hey, opening this as a follow-up to #211 as I'm noticing some type errors that have come up as a result of changes to the base types that were made at the end of that PR. Here's a small example:

We have a root type most Validation functions extend from, BaseValidation:
```ts
type BaseValidation<TInput = any> = {
    async: false;
    _parse(input: TInput): PipeResult<TInput>;
};
```
And in the codebase these are "extended" via type intersections like so:
```ts
type EmojiValidation<TInput extends string> = BaseValidation<TInput> & {
    type: 'emoji';
    requirement: RegExp;
};
```
But this introduces a problem... because all Validation functions have a `type` and a `requirement`, but BaseValidation doesn 't include those! This causes what should be a good way to build a type guard for Validation functions to fail:

![image](https://github.com/fabian-hiller/valibot/assets/3144549/189f10b6-b1a5-4187-8b72-64adbc7552bd)

This is a case where we should be using `interface` instead of `type`, because here we're adhering to a common set of properties that each individual validation function narrows for its specific signature. This also applies to BaseSchema too, which suffers from this same problem.

So instead of the above, we should do this instead:
```ts
interface BaseValidation<TInput = any> {
    type: string;
    async: false;
    message: string;
    requirement: unknown;
    _parse(input: TInput): PipeResult<TInput>;
};

interface EmojiValidation<TInput extends string> extends BaseValidation<TInput> {
    type: 'emoji'; // Narrows the base property to a specific string
    requirement: RegExp; // Narrows the unknown property value to a specific type
};
```
And with that we can validate that all Validations share a common set of properties! (Also, we get a nice side benefit of [Typescript running slightly faster](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections) too!)

Saeris:
I can open a PR to resolve this later this afternoon maybe. Should be a relatively quick fix.

fabian-hiller:
My comment from issue #258:

[...]

Issue #258, opened 2023-11-19
[INTROSPECTION][DISCUSSION] allow to introspect with BaseSchema & BaseSchemaAsync


ecyrbe:
At the moment when introspecting generic schemas, typescript don't know about the existance of the schema type attribute .

exemple, with error at line 6: 
![image](https://github.com/fabian-hiller/valibot/assets/633115/fd5888b3-683f-4561-9d52-2fed4fb95c75)

### Add a generic attribute to BaseSchema and BaseSchemaAsync : 
```ts
/**
 * Base schema type.
 */
export type BaseSchema<TInput = any, TOutput = TInput> = {
  /**
   * The schema type. ie: 'string', 'number', 'object', 'array', 'boolean', etc.
   */
  type: string;

  /**
   * Whether it's async.
   */
  async: false;

  /**
   * Parses unknown input based on its schema.
   *
   * @param input The input to be parsed.
   * @param info The parse info.
   *
   * @returns The parse result.
   *
   * @internal
   */
  _parse(input: unknown, info?: ParseInfo): _ParseResult<TOutput>;
  /**
   * Input and output type.
   *
   * @internal
   */
  _types?: { input: TInput; output: TOutput };
};

/**
 * Base schema async type.
 */
export type BaseSchemaAsync<TInput = any, TOutput = TInput> = {
  /**
   * The schema type. ie: 'string', 'number', 'object', 'array', 'boolean', etc.
   */
  type: string;

  /**
   * Whether it's async.
   */
  async: true;
  /**
   * Parses unknown input based on its schema.
   *
   * @param input The input to be parsed.
   * @param info The parse info.
   *
   * @returns The parse result.
   *
   * @internal
   */
  _parse(input: unknown, info?: ParseInfo): Promise<_ParseResult<TOutput>>;
  /**
   * Input and output type.
   *
   * @internal
   */
  _types?: { input: TInput; output: TOutput };
};
```

### Add an introspection helper 

```ts
function getSchemaType<T extends BaseSchema|BaseSchemaAsync>(schema: T): string
```

### Add an introspection visitor
```ts
function matchSchema<T extends BaseSchema|BaseSchemaAsync>(schema: T): SchemaVisitorBuilder;
```
exemple : 
```ts
function introspectSchema<T extends BaseSchema|BaseSchemaAsync>(schema: T) {
  matchSchema(schema)
    .with(
      'string', (stringSchema)=> { // handle string }
   )..with(
[...]

Issue #250, opened 2023-11-10
Consider adding "requirement" to "issue"


amok:
It would be nice to have the validator arguments (aka requirements, f.i see [here](https://github.com/fabian-hiller/valibot/blob/main/library/src/validations/maxLength/maxLength.ts#L13)) specified in the issue object

Consider the following example

```typescript
const Schema = object({
  password: string([minLength(8)]),
});

console.log(safeParse(Schema, { password: 'p' }).error.issues)
```

it logs out 

```js
{
	reason: "string"
	validation: "min_length"
	origin: "value"
	message: "Invalid length"
	input: "p"
	...
}
```

the ask here is to also add minLength' args to the issue object, f.i.:

```json
{
	"reason": "string"
	"validation": "min_length"
	"origin": "value"
	"message": "Invalid length"
	args: [8],
	"input": "p"
	
}
```

so they're available for the code which calls `parse`

This could be useful for error handling and i18n. For example, instead of re-generating schemas per language with `minLength(8, MESSAGE_IN_MY_LANG)`, we could do something like:

```js
try {
    const data = parse(Schema, input)
    ...
} catch (error) {
    const issue = error.issues[0]
    
    if (issue.validation === 'min_length') {
         showNotification(translate('Length should be at least {{min}}'), { min: issue.args[0] })
    }
    ...
}
```

Sometimes message' text depends on the UI context and this approach helps defining error messages closer to the place of `parse` usage.

fabian-hiller:
Thank you for creating this issue. This is a duplicate of #191. We are in the process of improving this part of the library with #211.

amok:
Thanks for the reply @fabian-hiller. Can't believe I missed it searching through the issues. Feel free to close this one as a duplicate.

fabian-hiller:
That's not a problem. I am happy for everyone who contributes to Valibot, and you are now a part of it! Thank you! I close the issue once #211 is merged.

fabian-hiller:
v0.21.0 is now available.

PR #259, opened 2023-11-19
Use `interface` instead of `type` for all Base types

Saeris:
I'll add further details to this PR soon, for now see #257 for background

### DIFF
--- a/library/src/schemas/any/any.ts
+++ b/library/src/schemas/any/any.ts
@@ -4,7 +4,7 @@ import { executePipe } from '../../utils/index.ts';
 /**
  * Any schema type.
  */
-export type AnySchema<TOutput = any> = BaseSchema<any, TOutput> & {
+export interface AnySchema<TOutput = any> extends BaseSchema<any, TOutput> {
   /**
    * The schema type.
    */
@@ -13,7 +13,7 @@ export type AnySchema<TOutput = any> = BaseSchema<any, TOutput> & {
    * The validation and transformation pipeline.
    */
   pipe: Pipe<any> | undefined;
-};
+}
 
 /**
  * Creates a any schema.
--- a/library/src/schemas/any/anyAsync.ts
+++ b/library/src/schemas/any/anyAsync.ts
@@ -4,7 +4,8 @@ import { executePipeAsync } from '../../utils/index.ts';
 /**
  * Any schema type.
  */
-export type AnySchemaAsync<TOutput = any> = BaseSchemaAsync<any, TOutput> & {
+export interface AnySchemaAsync<TOutput = any>
+  extends BaseSchemaAsync<any, TOutput> {
   /**
    * The schema type.
    */
@@ -13,7 +14,7 @@ export type AnySchemaAsync<TOutput = any> = BaseSchemaAsync<any, TOutput> & {
    * The validation and transformation pipeline.
    */
   pipe: PipeAsync<any> | undefined;
-};
+}
 
 /**
  * Creates an async any schema.
--- a/library/src/schemas/array/array.ts
+++ b/library/src/schemas/array/array.ts
@@ -17,10 +17,10 @@ import type { ArrayPathItem } from './types.ts';
 /**
  * Array schema type.
  */
-export type ArraySchema<
+export interface ArraySchema<
   TItem extends BaseSchema,
   TOutput = Output<TItem>[]
-> = BaseSchema<Input<TItem>[], TOutput> & {
+> extends BaseSchema<Input<TItem>[], TOutput> {
   /**
    * The schema type.
    */
@@ -37,7 +37,7 @@ export type ArraySchema<
    * The validation and transformation pipeline.
    */
   pipe: Pipe<Output<TItem>[]> | undefined;
-};
+}
 
 /**
  * Creates a array schema.
--- a/library/src/schemas/array/arrayAsync.ts
+++ b/library/src/schemas/array/arrayAsync.ts
@@ -18,10 +18,10 @@ import type { ArrayPathItem } from './types.ts';
 /**
  * Array schema async type.
  */
-export type ArraySchemaAsync<
+export interface ArraySchemaAsync<
   TItem extends BaseSchema | BaseSchemaAsync,
   TOutput = Output<TItem>[]
-> = BaseSchemaAsync<Input<TItem>[], TOutput> & {
+> extends BaseSchemaAsync<Input<TItem>[], TOutput> {
   /**
    * The schema type.
    */
@@ -38,7 +38,7 @@ export type ArraySchemaAsync<
    * The validation and transformation pipeline.
    */
   pipe: PipeAsync<Output<TItem>[]> | undefined;
-};
+}
 
 /**
  * Creates an async array schema.
--- a/library/src/schemas/bigint/bigint.ts
+++ b/library/src/schemas/bigint/bigint.ts
@@ -8,7 +8,8 @@ import {
 /**
  * Bigint schema type.
  */
-export type BigintSchema<TOutput = bigint> = BaseSchema<bigint, TOutput> & {
+export interface BigintSchema<TOutput = bigint>
+  extends BaseSchema<bigint, TOutput> {
   /**
    * The schema type.
    */
@@ -21,7 +22,7 @@ export type BigintSchema<TOutput = bigint> = BaseSchema<bigint, TOutput> & {
    * The validation and transformation pipeline.
    */
   pipe: Pipe<bigint> | undefined;
-};
+}
 
 /**
  * Creates a bigint schema.
--- a/library/src/schemas/bigint/bigintAsync.ts
+++ b/library/src/schemas/bigint/bigintAsync.ts
@@ -12,10 +12,8 @@ import {
 /**
  * Bigint schema async type.
  */
-export type BigintSchemaAsync<TOutput = bigint> = BaseSchemaAsync<
-  bigint,
-  TOutput
-> & {
+export interface BigintSchemaAsync<TOutput = bigint>
+  extends BaseSchemaAsync<bigint, TOutput> {
   /**
    * The schema type.
    */
@@ -28,7 +26,7 @@ export type BigintSchemaAsync<TOutput = bigint> = BaseSchemaAsync<
    * The validation and transformation pipeline.
    */
   pipe: PipeAsync<bigint> | undefined;
-};
+}
 
 /**
  * Creates an async bigint schema.
--- a/library/src/schemas/blob/blob.ts
+++ b/library/src/schemas/blob/blob.ts
@@ -8,7 +8,7 @@ import {
 /**
  * Blob schema type.
  */
-export type BlobSchema<TOutput = Blob> = BaseSchema<Blob, TOutput> & {
+export interface BlobSchema<TOutput = Blob> extends BaseSchema<Blob, TOutput> {
   /**
    * The schema type.
    */
@@ -21,7 +21,7 @@ export type BlobSchema<TOutput = Blob> = BaseSchema<Blob, TOutput> & {
    * The validation and transformation pipeline.
    */
   pipe: Pipe<Blob> | undefined;
-};
+}
 
 /**
  * Creates a blob schema.
--- a/library/src/schemas/blob/blobAsync.ts
+++ b/library/src/schemas/blob/blobAsync.ts
@@ -12,7 +12,8 @@ import {
 /**
  * Blob schema async type.
  */
-export type BlobSchemaAsync<TOutput = Blob> = BaseSchemaAsync<Blob, TOutput> & {
+export interface BlobSchemaAsync<TOutput = Blob>
+  extends BaseSchemaAsync<Blob, TOutput> {
   /**
    * The schema type.
    */
@@ -25,7 +26,7 @@ export type BlobSchemaAsync<TOutput = Blob> = BaseSchemaAsync<Blob, TOutput> & {
    * The validation and transformation pipeline.
    */
   pipe: PipeAsync<Blob> | undefined;
-};
+}
 
 /**
  * Creates an async blob schema.
--- a/library/src/schemas/boolean/boolean.ts
+++ b/library/src/schemas/boolean/boolean.ts
@@ -8,7 +8,8 @@ import {
 /**
  * Boolean schema type.
  */
-export type BooleanSchema<TOutput = boolean> = BaseSchema<boolean, TOutput> & {
+export interface BooleanSchema<TOutput = boolean>
+  extends BaseSchema<boolean, TOutput> {
   /**
    * The schema type.
    */
@@ -21,7 +22,7 @@ export type BooleanSchema<TOutput = boolean> = BaseSchema<boolean, TOutput> & {
    * The validation and transformation pipeline.
    */
   pipe: Pipe<boolean> | undefined;
-};
+}
 
 /**
  * Creates a boolean schema.
--- a/library/src/schemas/boolean/booleanAsync.ts
+++ b/library/src/schemas/boolean/booleanAsync.ts
@@ -12,10 +12,8 @@ import {
 /**
  * Boolean schema async type.
  */
-export type BooleanSchemaAsync<TOutput = boolean> = BaseSchemaAsync<
-  boolean,
-  TOutput
-> & {
+export interface BooleanSchemaAsync<TOutput = boolean>
+  extends BaseSchemaAsync<boolean, TOutput> {
   /**
    * The schema type.
    */
@@ -28,7 +26,7 @@ export type BooleanSchemaAsync<TOutput = boolean> = BaseSchemaAsync<
    * The validation and transformation pipeline.
    */
   pipe: PipeAsync<boolean> | undefined;
-};
+}
 
 /**
  * Creates an async boolean schema.
--- a/library/src/schemas/date/date.ts
+++ b/library/src/schemas/date/date.ts
@@ -8,7 +8,7 @@ import {
 /**
  * Date schema type.
  */
-export type DateSchema<TOutput = Date> = BaseSchema<Date, TOutput> & {
+export interface DateSchema<TOutput = Date> extends BaseSchema<Date, TOutput> {
   /**
    * The schema type.
    */
@@ -21,7 +21,7 @@ export type DateSchema<TOutput = Date> = BaseSchema<Date, TOutput> & {
    * The validation and transformation pipeline.
    */
   pipe: Pipe<Date> | undefined;
-};
+}
 
 /**
  * Creates a date schema.
--- a/library/src/schemas/date/dateAsync.ts
+++ b/library/src/schemas/date/dateAsync.ts
@@ -12,7 +12,8 @@ import {
 /**
  * Date schema async type.
  */
-export type DateSchemaAsync<TOutput = Date> = BaseSchemaAsync<Date, TOutput> & {
+export interface DateSchemaAsync<TOutput = Date>
+  extends BaseSchemaAsync<Date, TOutput> {
   /**
    * The schema type.
    */
@@ -25,7 +26,7 @@ export type DateSchemaAsync<TOutput = Date> = BaseSchemaAsync<Date, TOutput> & {
    * The validation and transformation pipeline.
    */
   pipe: PipeAsync<Date> | undefined;
-};
+}
 
 /**
  * Creates an async date schema.
--- a/library/src/schemas/enum/enum.ts
+++ b/library/src/schemas/enum/enum.ts
@@ -12,10 +12,8 @@ export type Enum = {
 /**
  * Native enum schema type.
  */
-export type EnumSchema<
-  TEnum extends Enum,
-  TOutput = TEnum[keyof TEnum]
-> = BaseSchema<TEnum[keyof TEnum], TOutput> & {
+export interface EnumSchema<TEnum extends Enum, TOutput = TEnum[keyof TEnum]>
+  extends BaseSchema<TEnum[keyof TEnum], TOutput> {
   /**
    * The schema type.
    */
@@ -28,7 +26,7 @@ export type EnumSchema<
    * The error message.
    */
   message: ErrorMessage;
-};
+}
 
 /**
  * Creates a enum schema.
--- a/library/src/schemas/enum/enumAsync.ts
+++ b/library/src/schemas/enum/enumAsync.ts
@@ -5,10 +5,10 @@ import type { Enum } from './enum.ts';
 /**
  * Native enum schema async type.
  */
-export type EnumSchemaAsync<
+export interface EnumSchemaAsync<
   TEnum extends Enum,
   TOutput = TEnum[keyof TEnum]
-> = BaseSchemaAsync<TEnum[keyof TEnum], TOutput> & {
+> extends BaseSchemaAsync<TEnum[keyof TEnum], TOutput> {
   /**
    * The schema type.
    */
@@ -21,7 +21,7 @@ export type EnumSchemaAsync<
    * The error message.
    */
   message: ErrorMessage;
-};
+}
 
 /**
  * Creates an async enum schema.
--- a/library/src/schemas/instance/instance.ts
+++ b/library/src/schemas/instance/instance.ts
@@ -13,10 +13,10 @@ export type Class = abstract new (...args: any) => any;
 /**
  * Instance schema type.
  */
-export type InstanceSchema<
+export interface InstanceSchema<
   TClass extends Class,
   TOutput = InstanceType<TClass>
-> = BaseSchema<InstanceType<TClass>, TOutput> & {
+> extends BaseSchema<InstanceType<TClass>, TOutput> {
   /**
    * The schema type.
    */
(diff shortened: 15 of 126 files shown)


fabian-hiller:
Please have a look at https://github.com/fabian-hiller/valibot/issues/257#issuecomment-1817906464

fabian-hiller:
I had a look at this PR and want to thank you for the work you put into it. In general I prefer to use `type`, but in the case of Valibot it may make sense to switch to `interface` for some definitions. Before I make a decision, I would like to hear your opinion on #258 and what advantages you see in `interface` and if you have encountered any disadvantages during implementation.

Saeris:
Okay, here are a few reasons why I think `interface` is preferable to `type` in a *library* like Valibot:

1. To speak directly to the original issue, this change is being proposed as a way to help 3rd-party library authors to build their own libraries on top of Valibot. This is because `interface` supports [Declaration Merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html), an important feature that provides that developer a means to add new functionality to Valibot with their library without necessitating the end user to adopt usage of a type alias when consuming the library. Which could be further complicated if multiple libraries that extend Valibot are used, where this quickly becomes a mess for the end user to unravel. I think Declaration Merging has a good use-case too in that an end-user can add support for their project that might not make sense to contribute upstream to Valibot itself.

2. Another consideration from [Types vs Interfaces](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces) in the official docs is that:

    > Interface names will [always appear in their original form](https://www.typescriptlang.org/play?#code/PTAEGEHsFsAcEsA2BTATqNrLusgzngIYDm+oA7koqIYuYQJ56gCueyoAUCKAC4AWHAHaFcoSADMaQ0PCG80EwgGNkALk6c5C1EtWgAsqOi1QAb06groEbjWg8vVHOKcAvpokshy3vEgyyMr8kEbQJogAFND2YREAlOaW1soBeJAoAHSIkMTRmbbI8e6aPMiZxJmgACqCGKhY6ABGyDnkFFQ0dIzMbBwCwqIccabcYLyQoKjIEmh8kwN8DLAc5PzwwbLMyAAeK77IACYaQSEjUWY2Q-YAjABMAMwALA+gbsVjNXW8yxySoAADaAA0CCaZbPh1XYqXgOIY0ZgmcK0AA0nyaLFhhGY8F4AHJmEJILCWsgZId4NNfIgGFdcIcUTVfgBlZTOWC8T7kAJ42G4eT+GS42QyRaYbCgXAEEguTzeXyCjDBSAAQSE8Ai0Xsl0K9kcziExDeiQs1lAqSE6SyOTy0AKQ2KHk4p1V6s1OuuoHuzwArMagA) in error messages
    
    I think it's important for the developer experience that when an error crops up, you can trace it back to its source quickly. Because of this behavior with `interface` names always being present is that at a glance, you know where to look next when a type error occurs. Contrast that with `type` (and in my experience this was historically an issue with type intersections especially) in that you *might not know where the type conflict originates*. When there are multiple type intersections in the chain, especially with arbitrary anonymous types, you'll have to dig for the property you have the conflict with.

3. There is a perf cost associated with the two approaches where the type checker is concerned (meaning, it could affect editor tooling). I ran `tsc --noEmit --extendedDiagnostics` on my library [`discordkit`](https://github.com/discordkit/discordkit), which I am refactoring away from `zod`, and compared that build to a local build of this PR (w/Interfaces) and `v0.21.0` (w/Types) as a benchmark. One *significant* difference I noticed between these 3 builds is that `w/Interfaces` has less that *1/5th* the Instantiations than the `w/Types` build does, and is about 30% less than `w/Zod`. The rest of the differences aren't as pronounced, though the Check Time for `w/Interfaces` was about 1s faster than the other two.

    |                            |   w/Zod | w/Interfaces | w/Types |
    | -------------------------- | ------: | -----------: | ------: |
    | Files:                     |    2055 |         2172 |    2172 |
    | Lines of Definitions:      |  314503 |       351242 |  351142 |
    | Lines of TypeScript:       |   21216 |        22891 |   22891 |
    | Lines of JavaScript:       |     302 |          302 |     302 |
    | Lines of JSON:             |       0 |            0 |       0 |
    | Lines of Other:            |       0 |            0 |       0 |
    | Identifiers:               |  476096 |       434391 |  433892 |
    | Symbols:                   |  643182 |       525129 |  522803 |
    | Types:                     |  158165 |        71237 |   69277 |
    | Instantiations:            | 1338374 |       943901 | 5405629 |
    | Memory used:               | 800925K |      658503K | 676255K |
    | Assignability cache size:  |   67049 |        19106 |   20834 |
    | Identity cache size:       |    1486 |          746 |     632 |
    | Subtype cache size:        |     986 |         1744 |    1869 |
    | Strict subtype cache size: |     225 |          227 |     599 |
    | I/O Read time:             |   0.18s |        0.20s |   0.21s |
    | Parse time:                |   1.01s |        1.03s |   0.99s |
    | ResolveTypeReference time: |   0.02s |        0.02s |   0.02s |
    | ResolveLibrary time:       |   0.02s |        0.01s |   0.02s |
    | Program time:              |   2.02s |        2.88s |   2.89s |
    | Bind time:                 |   0.53s |        0.54s |   0.57s |
    | Check time:                |   3.28s |        2.22s |   3.52s |
    | printTime time:            |   0.00s |        0.00s |   0.00s |
    | Emit time:                 |   0.00s |        0.00s |   0.00s |
    | Total time:                |   5.83s |        5.64s |   6.98s |

    But i don't have a good understanding of what impact Instantiations has on the performance of the type checker. Personally I have noticed that the `w/Types` build has significantly slower IDE tooltips than `w/Zod`, which isn't present in `w/Interfaces`, but I'm unsure if this is related or not.
    
    One other benchmark I ran was to compare the output of `tsc --noEmit --generateTrace ./trace` on `w/Types` and `w/Interfaces` so that I could go into [Perfetto](https://ui.perfetto.dev/) and drill into the stack traces. One of note was comparing the `checkSourceFile` trace on `discordkit/packages/client/src/channel/__tests__/getchannelinvites.spec.ts` where the execution duration on `w/Types` was 205ms vs 24ms on the `w/Interfaces` build.
    
    
    image


    Now I don't know much about perf testing the Typescript compiler much more than what I could read in a day, but the comparison that comes to mind for me is that if I were playing say, Dota 2, I would be at a significant disadvantage with a ping of 200ms vs my opponents if theirs was only 25ms. If this difference in `checkSourceFile` (which was consistently longer for all schema related files in `w/Types`) has any impact on the responsiveness I was experiencing with my IDE, then I think it's reasonable to assume that switching to using `interface` here (or some other related change I made) is having a meaningful impact on the developer experience of using Valibot.
    
Generally, the advice I've been seeing lately is, `interface` is still preferable to `type` for describing the shape of an object in TypeScript, especially if you're extending objects and/or you are authoring a library. In other contexts, it's more a matter of personal preference. If it's your own application code, fine, but if its library code, then it could create perf issues for consumers.

I couldn't find much supporting evidence for the performance impact of `extends` vs intersections (specifically coupled with generics), but my hunch is that TypeScript doesn't always cache type intersections (historically this was an [issue](https://github.com/microsoft/TypeScript/pull/42149)) when generics are involved, so the pattern of `BaseSchema & { ... }` might cause a lot of extra instantiations. More digging is required on the huge difference in that metric. One [example](https://hackernoon.com/top-5-typescript-performance-tips) that did come up was a perf optimization where you need to extract out a complex conditional to its own named type so it can be cached. I had seen something similar in another article where an intersection was involved too, but I seem to have lost the link. I'll add it if I can find it. tl;dr was, if your intersection includes conditional types, extract the conditional to a named type with a generic, and intersect your base type with that type instead of doing it inline.

This [article](https://trpc.io/blog/typescript-performance-lessons) by the tRPC team was also an interesting read, and I recommend checking it out as got really buried in search results.

Another insightful [article](https://blog.openreplay.com/five-tips-to-improve-your-typescript-code/) pointed out that Unions `|` and Intersections `&` do not behave like logical OR `||` or AND `&&`, which can result in some unintended gotchas. Objects defined via `type` also implicitly have an index signature, where `interface`s do not, which can also lead to unexpected results.

fabian-hiller:
Thank you for your detailed response. I will examine the code and make a decision by the end of the month.

fabian-hiller:
Short update: I wasn't able to review this PR last month. It's planned to do so in the coming weeks.

ariskemper:
**Interfaces vs Types in TypeScript**

**Declaration Merging:**
**Interfaces:** Can be merged. This is useful when you're extending existing interfaces or working with third-party TypeScript definitions.
**Types:** Cannot be merged. Each type alias is unique.


**Extending:**
**Interfaces:** Easily extended using the extends keyword, promoting reusability.
**Types:** Can be extended using intersections, but it's not as intuitive as with interfaces.
Implementation Specifics:

**Interfaces:** More suited for defining public API's of classes.
**Types:** Better for complex type transformations or when using union types.

**Readability and Clarity:**

**Interfaces:** May be more readable and clear for object-oriented patterns.
**Types:** Sometimes more concise, especially with complex type relationships.

**Performance Considerations**
In terms of performance, there is no significant difference between interfaces and types in TypeScript. TypeScript compiles down to JavaScript, and both interfaces and types are used for type-checking during development and do not exist in the compiled JavaScript code. Therefore, they don't have a runtime performance impact.

**Recommendations**
Use interfaces when you need to define the shape of objects and you anticipate extending these shapes or integrating with third-party type definitions.
Use types for more complex type transformations or when working with unions and intersections.
Consistency is key. Stick to one approach within a project as much as possible to maintain readability and ease of maintenance.

Saeris:
I'll give this PR an update pass this weekend to merge in the recent changes on main.

ariskemper:
@Saeris we have done some optimisations for this via tsconfig. I hardly believe it has significant impact, because valibot is not big library and it doesn't have  dependencies.

Saeris:
> @Saeris we have done some optimisations for this via tsconfig. I hardly believe it has significant impact, because valibot is not big library and it doesn't have dependencies.

The perf issues add up when you build complex schemas such as those I've authored in [discordkit](https://github.com/discordkit/discordkit), as you end up doing many nested levels of type inference.

What specific setting in tsconfig affects this for the end user of Valibot? I'm unaware of a config setting that would have a perf effect downstream.

ariskemper:
@Saeris Please don't misunderstand my preference for interfaces over intersections. I'm aware that sometimes refactoring to interfaces may not be necessary and can consume additional development time. However, it's important to recognize the distinct advantages of interfaces in programming.

Interfaces provide a structured and conflict-sensitive method for creating object types, unlike intersections that simply combine properties and occasionally produce 'never' outcomes. This active detection and resolution of property conflicts by interfaces are vital for addressing issues early, saving time and effort in the long run.

Efficiency is another area where interfaces shine. The caching of type relationships in interfaces results in enhanced performance, a feature not found in the entirety of intersection types.

When it comes to type checking, interfaces offer a comprehensive approach. They ensure that each constituent is thoroughly checked against a target intersection type before moving to the "effective" or "flattened" type, leading to more precise and dependable type-checking.

Therefore, despite the extra time it might take initially, extending types using interfaces/extends is generally recommended over creating intersection types. This approach not only makes coding clearer and more efficient but also minimizes conflicts, thereby improving both the development process and the overall quality of the code.

What we added in terms of performance is skipLibCheck in tsconfig.

Saeris:
So, what I'm reading is that you're in agreement that `interface` is preferred over `type`.

I'm in agreement that I think `interface` is preferrable. So then, I don't know what we're in disagreement about?

`skipLibCheck` helps type-checking performance for authoring code within Valibot's codebase, but that would not carry upwards to end users unless they've enabled that setting themselves. In my diagnostics investigation above, I also had `skipLibCheck` enabled in my tsconfig.

ariskemper:
@Saeris the problem with this PR is that it is too large and could involve breaking changes, so i would prefer to do such refactoring in iteration of smaller changes and more PRs!

Saeris:
Valibot is still in alpha, is it not? This didn't seem to be an issue with #211 which I submitted previously. There are comparatively less changes in this PR, which by and large is just changing intersections to `extends`.

ariskemper:
@Saeris  We could connect via Discord and then do a video call and you could show me in your project the type inference problems you encounter? You could find me as ariskemper.

Saeris:
> have seen you are changing also some logic, replacing for loops etc..

Were my code comments explaining these changes not sufficient?

Take `getFallbacks()` for example, where on line 31 there's an undefined variable declaration: `let fallbacks: any;` which the subsequent conditionals possibly assign to, before it is returned at the end of the function.

This `any` in the internals of this function is not type-safe. It breaks the API contract of the return type of the function because `any` will always satisfy that return type. The changes I made to this function (and the others following the same pattern) makes each of them internally type-safe, adhering to the function return type.

I made sure to make these changes *without* altering any of the unit tests so that the existing functionality stayed the same. If you think the existing tests for these aren't sufficient, that's another issue we can address. But I was intentional with the logic changes I made so that it would *improve* the reliability of the code.

ariskemper:
@Saeris yes, but this are still part of PR and needed to be reviewed. We should just stick to smaller PRs, otherwise many times this kind of large PRs stays opened, that doesn't mean your contribution is not valued, but still needs to be reviewed and i also believe you don't like to review large PRs or do you?

Saeris:
There's a difference between reviewing a large PR and one that touches many files by implementing a pattern.

I understand there are a lot of files touched here. In the time since your last comment, I went back and counted the files touched in this PR that all change intersections to `extends`, which accounts for ~125 out of 148 files. I think this can be reviewed in less time that we've spent discussing when and whether it will be reviewed.

library/src/error/flatten/flatten.ts
```diff
@@ -78,7 +78,7 @@ type NestedPath<TSchema extends BaseSchema | BaseSchemaAsync> =
     TSchema extends
         | ObjectSchema<infer TEntries, infer TRest>
         | ObjectSchemaAsync<infer TEntries, infer TRest>
-    ? TRest extends BaseSchema | BaseSchemaAsync
```
Saeris, line 81: Because this change may be ambiguous, the type condition we have here is attempting to remove `undefined` from the union of types that `TRest` is assignable to. because `TRest` is inferred from the previous conditional. The utility type `NonNullable` takes whatever the values that `TRest` can be and removes `null` and `undefined` from that union.

It's been over a month since I had run into a type error here, so I don't remember the exact error message was, but the above ended up being the solution.

fabian-hiller:
Since we are still pre v1, a larger PR like this is fine. If I disagree with a change, I will revert it or provide feedback before merging the PR. I prefer not to complicate things at the moment. Any PR is better than no PR. In the long run, we will professionalize everything bit by bit.

ariskemper:
@fabian-hiller @Saeris could we split this at least in 2 PR, which exactly describe the changes, so we keep in this one
only refactoring to interfaces and everything else, we transfer to another one. Maybe you will return to PRs at some point in the future and it's good to note that keeping them in reasonable size is much better then having large changes. Not really sure if any PR is better then no PR, PRs also takes time to review them. I updated the Contributing guide for PRs so we work on the quality over quantity.

Saeris:
> this PR is here allready for few months

I mean, it was opened right before the holiday season. It was to be expected that it would sit open for this long?

If it helps, I will author smaller PRs in the future. (though to note, after a joke on Twitter, I have a branch refactoring everything to classes ready for consideration too @fabian-hiller lol)

It may be worth reviewing @ariskemper's other PRs that add new validations before rounding this off? I doubt it will take much time at all to apply the interface pattern here to those incoming changes, but that's a responsibility I am willing to take on rather than request him to do it.

fabian-hiller:
In the next few days I will focus on improving the documentation. I have already started to create more guides explaining different schema functions. After that I'll work on the i18n feature and the integration of Valibot into Qwik. After that I plan to investigate this PR. In this case, I don't think it matters which PRs we merge first.

library/src/methods/getDefaults/getDefaults.ts
```diff
@@ -22,36 +16,24 @@ import type { DefaultValues } from './types.ts';
...
-      for (let key = 0; key < schema.items.length; key++) {
-        defaults.push(getDefaults(schema.items[key]));
-      }
-    }
+    return getDefault(schema);
+  }
+  // Otherwise, check if schema is of kind object or tuple
+  // If it is an object schema, set object with default value of each entry
+  if (isObjectSchema(schema)) {
+    return Object.entries(schema.entries).reduce(
```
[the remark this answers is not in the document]
mxdvl: If using `Object.fromEntries` in the async version, why not use it here as well? Is it because of the performance cost of creating an extra array?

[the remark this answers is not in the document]
Saeris: I suppose it can be done with `Object.fromEntries` here too, since in either case a type assertion is required. I implemented one after the other and solved for type errors along the way, so these were just the solutions I arrived at then. Will make this change when I resolve merge conflicts.

fabian-hiller:
i18n is merged. I will review this PR soon.

fabian-hiller:
I plan to review and probably merge this PR in the next days.

fabian-hiller:
I investigated this PR but it does not solve what I was looking for. So I am not sure anymore if I want to merge it. Please see #257 to discuss with me there.

library/src/methods/getDefault/types.ts
```diff
@@ -1,6 +1,28 @@
-import type { Output } from '../../types/index.ts';
-import type { SchemaWithMaybeDefault } from './getDefault.ts';
-import type { SchemaWithMaybeDefaultAsync } from './getDefaultAsync.ts';
+import type { BaseSchema, BaseSchemaAsync, Output } from '../../types/index.ts';
+
+/**
+ * Schema with maybe default type.
+ */
+export interface SchemaWithMaybeDefault<TInput = any, TOutput = TInput>
```
fabian-hiller, line 6: Please move this into `getDefaults.ts`

library/src/methods/getDefault/types.ts
```diff
@@ -1,6 +1,28 @@
...
+  /**
+   * The optional default value.
+   */
+  default?: TOutput | (() => TOutput | undefined);
+}
+
+/**
+ * Schema with maybe default async type.
+ */
+export interface SchemaWithMaybeDefaultAsync<TInput = any, TOutput = TInput>
```
fabian-hiller, line 17: Please move this into `getDefaultsAsync.ts`

library/src/methods/getDefaults/getDefaults.ts
```diff
@@ -22,36 +16,25 @@ import type { DefaultValues } from './types.ts';
...
-      defaults = [];
-      for (let key = 0; key < schema.items.length; key++) {
-        defaults.push(getDefaults(schema.items[key]));
-      }
-    }
+    return getDefault(schema);
+  }
+  // Otherwise, check if schema is of kind object or tuple
+  // If it is an object schema, set object with default value of each entry
+  if (isObjectSchema(schema)) {
```
fabian-hiller, line 28: Please refactor this kind of code with your new util function. I am not sure about the name of the util. Something like `match`, `fit` or `comply` could work. Maybe you have better ideas.

Example:

```ts
if (match('string', schema)) {
  // ...
}
```

library/src/methods/getFallback/types.ts
```diff
@@ -1,6 +1,27 @@
-import type { Output } from '../../types/index.ts';
-import type { SchemaWithMaybeFallback } from './getFallback.ts';
-import type { SchemaWithMaybeFallbackAsync } from './getFallbackAsync.ts';
+import type { BaseSchema, BaseSchemaAsync, Output } from '../../types/index.ts';
+import type { FallbackInfo } from '../fallback/types.ts';
+
+/**
+ * Schema with maybe fallback type.
+ */
+export interface SchemaWithMaybeFallback<TInput = any, TOutput = TInput>
```
fabian-hiller, line 7: Please move that into `getFallback.ts`.

library/src/methods/getFallback/types.ts
```diff
@@ -1,6 +1,27 @@
...
+  /**
+   * The optional fallback value.
+   */
+  fallback?: TOutput | ((info?: FallbackInfo) => TOutput);
+}
+
+/**
+ * Schema with maybe fallback async type.
+ */
+export interface SchemaWithMaybeFallbackAsync<TInput = any, TOutput = TInput>
```
fabian-hiller, line 18: Please move that into `getFallbackAsync.ts`.

library/src/utils/isSchema.ts
```diff
@@ -0,0 +1,12 @@
+import type { BaseSchema, BaseSchemaAsync } from '../types/schema.ts';
+import { isNonNullable } from './isNonNullable.ts';
+
+export const isSchema = (val: unknown): val is BaseSchema | BaseSchemaAsync =>
```
fabian-hiller, line 4: Why do we need this util?

fabian-hiller (changes requested):

Saeris:
Just merged in the latest from the main branch, will address your feedback and requested changes as soon as I have a chance.

This latest commit adds the `hasType` utility we discussed in #250 (I may need to resolve some linting errors)

library/src/methods/getDefaults/getDefaults.ts
Saeris: Updated this with the `hasType` utility. Feel free to rename it and modify the usage code comment as you see fit.

library/src/utils/isSchema.ts
Saeris: I removed this when adding `hasType`. Strictly speaking I think this was a more thorough type guard for runtime validation that an object extends a Schema type, but it's not entirely necessary.

library/src/methods/getDefault/types.ts
Saeris: I moved these types because they created a circular dependency reference. IMO they are better kept in this file.

fabian-hiller: Why does it create a circular dependency reference? I prefer to move it to `getDefaults.ts` to be consistent with the rest of the source code. Types that are related to a particular function are placed above the function in the same file.

fabian-hiller (changes requested):
There are a few minor details we need to work out, but otherwise this PR seems ready to merge.

library/src/methods/getDefault/types.ts
Saeris: because for example:

```ts
export type DefaultValue<
  TSchema extends SchemaWithMaybeDefault | SchemaWithMaybeDefaultAsync,
>
```
in `types.ts` would import from `getDefaults.ts`, which itself imports `DefaultValue` from `types,ts`

library/src/methods/fallback/fallback.ts
```diff
@@ -8,17 +8,13 @@ import type { FallbackInfo } from './types.ts';
...
  * Schema with fallback type.
  */
-export type SchemaWithFallback<
-  TSchema extends BaseSchema = BaseSchema,
-  TFallback extends
-    | Output<TSchema>
-    | ((info?: FallbackInfo) => Output<TSchema>) = Output<TSchema>,
-> = TSchema & {
+export interface SchemaWithFallback<TInput = any, TOutput = TInput>
+  extends BaseSchema<TInput, TOutput> {
```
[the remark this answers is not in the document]
Saeris: I'll investigate and get back to you when I can. I don't think I have the time at the moment to properly answer your concern here.

library/src/methods/getDefault/types.ts
fabian-hiller: I will have a look at it. I don't think this is a problem but I could be wrong.

library/src/methods/fallback/fallback.ts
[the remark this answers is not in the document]
fabian-hiller: No worries. Maybe I will investigate and fix it this evening.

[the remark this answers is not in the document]
fabian-hiller: This does not seem to be possible with `interface`. So I will switch back to the previous implementation.

fabian-hiller (approved):

fabian-hiller:
I will do a final review later and probably merge this PR. 🙌

fabian-hiller:
v0.30.0 is available

library/src/methods/getDefaults/getDefaults.ts
```diff
@@ -26,32 +27,27 @@ export function getDefaults<
   TSchema extends SchemaWithMaybeDefault<
     BaseSchema | ObjectSchema<ObjectEntries, any> | TupleSchema<TupleItems, any>
   >,
->(schema: TSchema): DefaultValues<TSchema> {
-  // Create defaults variable
-  let defaults: any;
-
-  // If schema contains a default function, set its default value
+>(schema: TSchema): DefaultValues<TSchema> | undefined {
```
aboqasem, line 30: Hello, as per the [documentation for `getDefaults`](https://valibot.dev/api/getDefaults/):

> The difference to [getDefault](https://valibot.dev/api/getDefault/) is that for objects and tuples without an explicit default value, this function recursively returns the default values of the subschemas instead of undefined.

Did that change?

fabian-hiller: No, it should work just the same. All unit tests passed, so I do not expect any bugs, but maybe we are missing something. Please create an issue if you encounter any problems.

aboqasem: Ok thanks. It just broke the type when I upgraded so I was wondering.

fabian-hiller: Feel free to share your code if you thing something is wrong.

PR #270, opened 2023-11-27
feat: add mac address validation

ariskemper:


fabian-hiller:
Thank you for creating this PR! I plan to include this in our next release.

ariskemper:
@fabian-hiller thanks, valibot looks like nice lightweight validation library. Looking forward to contribute more to the project.

fabian-hiller:
Thank you so much! Your words mean a lot to me. I plan to reach v1 beta in January. In about 3 weeks I will be able to work full time on my open source projects again.

fabian-hiller:
Your implementation looks good. I'm just not sure what a valid MAC address is. Do all MAC addresses have to be 12 hexadecimal numbers long? Who defines this standard? Is there a clear source?

The following formats are clear:

- `00:00:00:00:00:00`
- `00-00-00-00-00-00`
- `0000.0000.0000`

But I am still unsure about the following formats:

- `0000:0000:0000:0000` // Colons instead of dots and 16 digits
- `00:00:00:00:00:00:00:00` // 16 instead of 12 digits

fabian-hiller:
Also I am unsure if we should rename the function to `macAddress`. Do you have an opinion on this?

ariskemper:
There are 48 bit (IP4 style ) and 64 bit (IPv6 style ) MAC addresses. MAC addresses are formed according to the principles of two numbering spaces based on extended unique identifiers (EUIs) managed by the [Institute of Electrical and Electronics Engineers](https://en.wikipedia.org/wiki/Institute_of_Electrical_and_Electronics_Engineers) (IEEE): EUI-48—which replaces the obsolete term MAC-48—and EUI-64.

> Your implementation looks good. I'm just not sure what a valid MAC address is. Do all MAC addresses have to be 12 hexadecimal numbers long? Who defines this standard? Is there a clear source?
> 
> The following formats are clear:
> 
> * `00:00:00:00:00:00`
> * `00-00-00-00-00-00`
> * `0000.0000.0000`
> 
> But I am still unsure about the following formats:
> 
> * `0000:0000:0000:0000` // Colons instead of dots and 16 digits
> * `00:00:00:00:00:00:00:00` // 16 instead of 12 digits

fabian-hiller:
I think both `mac` and `macAddress` are appropriate names. We can start with `mac` and rename it later if necessary. Are you sure the regex complies the standard? Then I would rely on you. I have not found a source on the Internet that describes the standard in a concise way. [This document ](https://standards.ieee.org/wp-content/uploads/import/documents/tutorials/eui.pdf)probably contains everything but I don't have the time to read it in detail right now.

ariskemper:
@fabian-hiller i believe current implementation should be ok, maybe we should just split to different validations, how it is done for ip, ipv4, ipv6 or leave it as it is for now.

fabian-hiller:
How would you divide it and also name it?

fabian-hiller:
Should we split it into `mac48` and `mac64`, similar to IP validation, and also offer `mac` if both are accepted?

ariskemper:
@fabian-hiller per my research this are valid formats for 48 and 64 bit MAC address.

fabian-hiller:
v0.23.0 is now available.

PR #281, opened 2023-12-02
feat: TS Config optimisation settings

ariskemper:


fabian-hiller:
Thank you! I plan to review and merge this PR in the next days.

ariskemper:
@fabian-hiller you could check difference in parse time and check time with `tsc --extendedDiagnostics`

library/tsconfig.json
```diff
@@ -6,7 +6,15 @@
...
     "noEmit": true
   },
-  "include": ["."]
+  "include": [
+    "src/**/*.ts",
+    "mod.ts",
+    "tsup.config.ts",
+    "vite.config.ts"
+  ],
+  "exclude": ["node_modules"]
```
ariskemper, line 19: @fabian-hiller will remove this, because it is done by default.

library/tsconfig.json
```diff
@@ -6,7 +6,14 @@
     "module": "ESNext",
     "moduleResolution": "node",
     "allowImportingTsExtensions": true,
+    "forceConsistentCasingInFileNames": true,
```
fabian-hiller, line 9: I think that this is not necessary because `true` is the default value.

PR #290, opened 2023-12-09
feat: IBAN validation

ariskemper:


fabian-hiller:
Thanks for working on this! I will review it in the next few days. Another validator that's missing is `card` or `creditCard`. There is already an `isLuhnAlgo` utility implemented for this.

ariskemper:
@fabian-hiller will check isLuhnAlgo and add validator for credit cards.

ariskemper:
@fabian-hiller could you review it and merge this PR?

fabian-hiller:
Next week is the last week of this semester and I have one exam left. That's why I'm not that fast with reviewing right now. I will try to do it over the weekend or next week.

fabian-hiller:
I am not an IBAN expert, but I suspect that your code does not detect a wrong IBAN in every case. For example, if you enter a letter instead of a number that happens to be converted to a valid number by `(char.charCodeAt(0) - 55).toString()`, `isIBAN` returns `true`.

I found [validator.js](https://github.com/validatorjs/validator.js) on GitHub. Maybe we can get some inspiration from there [`isIBAN`](https://github.com/validatorjs/validator.js/blob/master/src/lib/isIBAN.js) function.

ariskemper:
@fabian-hiller could you give a case? This functon first check the length and then rearanges, so it moves first 4 chars at the end and after replaces chars from A-Z to a 2 digit number. If there would be 4 chars, it would replace those 4 with 2 digit number. Then on this string of numbers we perform mod97.  I tested it against values from [validator.js](https://github.com/validatorjs/validator.js)
Validator.js has also whitelist and blacklist functionality.

fabian-hiller:
Thank you for your feedback. I will check the code again.

fabian-hiller:
Unfortunately, due to our rewrite a few months ago, this PR is no longer up to date with our source code. I still plan to add this action, but other things have a higher priority for me at the moment. If you or anyone else in the community is interested in creating a new PR, that would be great.

fabian-hiller:
I will close this PR. However, I will keep it in mind when adding this action in the future.

PR #295, opened 2023-12-11
feat: add downloads, license, npm, ci badges

ariskemper:


fabian-hiller:
Thank you! That's great!

PR #300, opened 2023-12-16
feat: bump TypeScript version to 5.3.3.

ariskemper:


fabian-hiller (changes requested):
We use a pnpm monorepo. There should only be one top level `pnpm-lock.yaml` file.

fabian-hiller (approved):

PR #301, opened 2023-12-16
feat: add match regex validation

ariskemper:


fabian-hiller:
Is there a difference to our `regex` validation function?

ariskemper:
@fabian-hiller somehow i missed regex....I still believe match is better naming since you could read that value should match regular expression.

fabian-hiller:
Yes, I agree. `match` makes the API more readable but `regex` is on the other side more clear and easier to find.

PR #304, opened 2023-12-17
feat: add hexadecimal string validation

ariskemper:


fabian-hiller:
Thank you so much! This is great! Another idea I have is to add a validation function for strings containing color codes in HEX or RGB format. I recommend supporting the same color codes there as CSS does.

ariskemper:
@fabian-hiller i plan to add hex color validation, octa string validation.

PR #306, opened 2023-12-17
feat: add flag --no-isolate to improve performance

ariskemper:


ariskemper:
@fabian-hiller maybe we should have another script like test.dev to run it locally with --no-isolate for faster development and in the pipeline we run with vitest.

fabian-hiller:
I will give you feedback later.

fabian-hiller:
Is there a reason why you would run it in isolation on GitHub?

ariskemper:
Since Valibot doesn' use global variables we could use it for Github. It would speed up our branches pipelines too.

fabian-hiller:
Yes, it speeds the process by more than 10x. So thanks a lot for the PR.

PR #307, opened 2023-12-18
feat: hash validation

ariskemper:


fabian-hiller:
I think for Valibot it might be better to change the argument to an array to be able to allow different hash types. What do you think?

```ts
// Just one hash type
hash(['md5']);

// Multiple hash types
hash(['md5', 'sha1', 'crc32']);
```

ariskemper:
Do you have use case for different hash types?

fabian-hiller:
I don't have this use case in my own projects, but I believe there could be, for example, APIs that accept more than two hash types as a parameter. We could also accept both, a string or an array as the first argument.

```ts
// With string arg (slower and more complicated)
const Schema1 = union([string([hash('md5')]), string([hash('sha1')])]);

// With array arg (faster and easier to write)
const Schema2 = string([hash(['md5', 'sha1'])]);
```

ariskemper:
Good example, will add support for argument with multiple hash types.

fabian-hiller:
Thanks! I will review it later.

fabian-hiller:
I can't thank you enough for all the work you put into Valibot! Soon you will be the second top contributor. Feel free to use the project as a reference.

PR #324, opened 2023-12-27
feat: ISSN validation

ariskemper:


fabian-hiller:
I will close this PR. However, I will keep it in mind when adding this action in the future.

Issue #325, opened 2023-12-27
Add Elo as a supported credit card provider


raisiqueira:
Elo is a prominent [credit card](https://github.com/fabian-hiller/valibot/pull/292) provider in Brazil, and it would be greatly appreciated if Valibot could provide support for this.

fabian-hiller:
Thank you for creating this issue and helping to improve the library!

fabian-hiller:
Hey 👋 I still plan to support this is in the long run. I just prioritised other things in the past because of time constraints.

raisiqueira:
@fabian-hiller No worries.

PR #292, opened 2023-12-10
feat: credit card validation

ariskemper:

PR #326, opened 2023-12-27
feat: add Elo as a supported credit card provider

raisiqueira:
This is an additional merge request for the #292 adding Elo as a credit card provider.

closes https://github.com/fabian-hiller/valibot/issues/325


fabian-hiller:
Thank you very much! Where did you get the regex from? Is there a good source that I can use to check what's a valid Elo credit card?

raisiqueira:
Hi @fabian-hiller, The source is from this project: https://github.com/raisiqueira/react-payment-inputs/blob/master/src/utils/cardTypes.js. I've used this library when I was working on a Brazilian fintech. To check if Elo credit card number is valid, you can use [this website](https://geradornv.com.br/gerador-cartao-credito/) (Portuguese only) to generate a valid credit card number.

ariskemper:
@fabian-hiller here are tests cards also for Elo https://docs.adyen.com/development-resources/testing/test-card-numbers/

fabian-hiller:
Thank you for this PR. @ariskemper feel free to create another PR for other providers. How would you output the provider name? Not sure if this should be part of Valibot. But we could export the specific provider regex we use so that the same regex can be used to determine the provider with custom code.

ariskemper:
@raisiqueira would be enough maybe shorter and less complex Regex, because Elo credit card numbers usually starts with the following prefixes and are 16 digits in length:

4011
4312
4389
4514
4573
4576
5041
5066
5067
5090
6277
6362
6363
6500
6504
6505
6507
6509
6516
6550
we need to account for these prefixes and ensure that the total length of the card number is 16 digits?

raisiqueira:
Hey @ariskemper I believe that this regex that you shared is enough. I'll update the merge request.

ariskemper:
@raisiqueira if you could add more tests to cover different valid and invalid card numbers would be great.

raisiqueira:
@ariskemper Done. Check if the new assertions are enough.

ariskemper:
@raisiqueira for now should be fine, we could add extra tests in next releases for missing prefixes. @fabian-hiller what do you think?

fabian-hiller:
Looks good on the first view. I will focus on other parts of the project for this and maybe the next week. After that I will review and merge everything.

fabian-hiller:
Unfortunately, due to our rewrite a few months ago, this PR is no longer up to date with our source code. I still plan to add this action, but other things have a higher priority for me at the moment. If you or anyone else in the community is interested in creating a new PR, that would be great.

raisiqueira:
Hi @fabian-hiller, I'll open another merge request with the update, I believe that we can close this one.

Issue #287, opened 2023-12-07
Help Wanted: Completing the API Reference on Website


fabian-hiller:
> [!IMPORTANT]
> The following description is outdated. All synchronous APIs are already fully documented. We are just missing a few asyncronous APIs. If you still want to help, feel free to comment below.

This issue is used to coordinate the expansion of our API reference. The first reference was created by @kazizi55 in #220. Please use the already added references as a template. You can also use this issue if you encounter problems or have questions.

If you are interested in helping us extend the API reference, please add a comment to this issue with the API you want to take over. I will summarize the current state in the tables below.

<details>
  <summary>Schemas</summary><br />

| API | Author | Pull Request | Merged |
| -- | -- | -- | -- |
| any | @fabian-hiller | - | ✅ |
| array | @kazizi55 | #330 | ✅ |
| bigint | @fabian-hiller | - | ✅ |
| blob | @fabian-hiller | - | ✅ |
| boolean | @wout-junius | #316 | ✅ |
| date | @wout-junius | #315 | ✅ |
| enum\_ | @fabian-hiller | - | ✅ |
| instance | @fabian-hiller | - | ✅ |
| intersect | @fabian-hiller | - | ✅ |
| literal | @fabian-hiller | - | ✅ |
| map | @fabian-hiller | - | ✅ |
| nan | @fabian-hiller | - | ✅ |
| never | @fabian-hiller | - | ✅ |
| nonNullable | @fabian-hiller | - | ✅ |
| nonNullish | @fabian-hiller | - | ✅ |
| nonOptional | @fabian-hiller | - | ✅ |
| null\_ | @fabian-hiller | - | ✅ |
| nullable | @sqmasep | #328 | ✅ |
| nullish | @sqmasep | #329 | ✅ |
| number | @kazizi55 | #331 | ✅ |
| object | @kazizi55 | #357 | ✅ |
| optional | @sqmasep | #321 | ✅ |
| picklist | @sqmasep | #312 | ✅ |
| record | @fabian-hiller | - | ✅ |
| recursive | @fabian-hiller | - | ✅ |
| set | @fabian-hiller | - | ✅ |
| special | @fabian-hiller | - | ✅ |
| string | @kazizi55 | #220 | ✅ |
| symbol | @wout-junius | #317 | ✅ |
| tuple | @fabian-hiller | - | ✅ |
| undefined\_ | @fabian-hiller | - | ✅ |
| union | @fabian-hiller | - | ✅ |
[...]

PR #330, opened 2023-12-29
docs: add array related documentations

kazizi55:
Implemented array related docs as discussed in https://github.com/fabian-hiller/valibot/issues/287


fabian-hiller:
Thank you! I will review it in the next days!

PR #345, opened 2024-01-03
refactor: comparable function

ariskemper:


fabian-hiller:
Why do you think this is necessary? `comparableArray` and `comparableObject` are only used once and the previous  `comparable` functions seams not to complicated to me.

fabian-hiller:
Maybe there is a way to avoid `comparable` altogether. At least that is what I would expect from a testing framework.

ariskemper:
@fabian-hiller it could be solved with JSON.stringify, because vitest currently doesn't provide toDeepEqual. It has deep.equal, but it doesn't work as expected.
This works for example expect(JSON.stringify(schema)).toEqual(JSON.stringify(enumType(['key1', 'key2'])));

fabian-hiller:
`JSON.stringify()` is a good idea. Let's keep that in mind when rewriting the tests. I will give you feedback on the email tests today.

fabian-hiller:
One thing to note is that if the properties are in a different order, the `JSON. stringify()` solution breaks. This was perhaps the reason I implemented `comparable`.

ariskemper:
@fabian-hiller then we keep it in the same order.

fabian-hiller:
The order should not matter. That's why we should probably stick with `comparable` if there is no other solution.

ariskemper:
@fabian-hiller we could use extend matchers and add extra method toCompare? [](https://vitest.dev/guide/extending-matchers)

fabian-hiller:
Can you send some sample code showing how you would implement it?

ariskemper:
@fabian-hiller this adds additional code/types and i would keep it as i refactor in first commit.

fabian-hiller:
Thank you! This looks great. I will review it once I made some more progress in other areas of the library.

fabian-hiller:
Since we removed this function is the previous rewrite, I will close this PR.

PR #346, opened 2024-01-03
feat: video url validation

ariskemper:


fabian-hiller:
I can see the value of this validation function, but also see the problem that it is impossible to support all video URLs. Maybe it's too specific and we should recommend to custom implement such validations individually (e.g. using `startsWith`). I prefer that Valibot only covers clear and common cases.

ariskemper:
We could support 3 biggest and we could make it that it is possible to pass Record with custom mappings service => video url regex. Then with services array you could choose only the ones you need?

fabian-hiller:
Yes, that could work. But on the other hand, I don't want to overcomplicate our API. In special cases it is better to use `custom`, `regex` or write a custom validation action.

Issue #351, opened 2024-01-04
Document minimum typescript version


johanndev:
Hi!

The documentation is missing a note about the minimum typescript version that is required for valibot to work (5.0.2?)
This should probably be documented on the installation page.

If you could confirm the minimal typescript version, I can provide a PR for the website - thanks!

fabian-hiller:
Do you know if there is a way to figure out the minimum TypeScript version required? I think the most advanced feature we are using ist the [`const` modifier](https://github.com/microsoft/TypeScript/pull/51865) that was shipped with v5.

johanndev:
I'm not aware of a programmatic way of determining the required TS version - I found 5.0.2 as the min version just by trial and error.

fabian-hiller:
We could add the following to the "General" section of the [installation guide](https://valibot.dev/guides/installation/) below the `tsconfig.json` text.

> The minimum required TypeScript version is v5.0.2.

If you are interested, feel free to create a draft PR for an integration test.

johanndev:
> We could add the following to the "General" section of the [installation guide](https://valibot.dev/guides/installation/) below the tsconfig.json text.

Yes, that sounds perfect 👍

> If you are interested, feel free to create a draft PR for an integration test.

I'm gonna do some research first, I'll get back to you :)

fabian-hiller:
I added the min version info to the docs.

johanndev:
I created a draft PR: #354

As the original issue was addressed, I'm gonna close this discussion.

fabian-hiller:
Thank you!

PR #354, opened 2024-01-07
test: Add PoC for typescript version integration test

johanndev:
Addresses #351 

I included 4.9.5 to show the cutoff version.

I would suggest to exclude this version test from regular test runs and instead only run it periodically (say, once a day) in CI.


fabian-hiller:
Thank you for creating this PR. I am currently focused on other areas of the library. I will probably get back to this PR at the end of this month or in February.

fabian-hiller:
Right now these tests do not seem too important and I am cautious about adding more complexity to the repo. We will come back to this PR at a later point in time if necessary.

Issue #352, opened 2024-01-04
Nextjs build  Failed


mahersalhani:
Checking validity of types  .Failed to compile.

./node_modules/valibot/dist/index.d.ts:1980:27
Type error: Type parameter declaration expected.

  1978 |  * @returns A picklist schema.
  1979 |  */
> 1980 | declare function picklist<const TOptions extends PicklistOptions>(options: TOptions, message?: ErrorMessage): PicklistSchema<TOptions>;
       |                           ^
  1981 | /**
  1982 |  * See {@link picklist}
  1983 |  *
error Command failed with exit code 1.

I use 
"valibot": "^0.25.0",

fabian-hiller:
Can you send me your schema? But it seems like this is an external problem.

ZilvinasAbr:
I've got similar issue when trying to use valibot inside of a server action for validating form submission with Next.js app router.

The issue happens when trying to build the next.js application.

Code looks something like this:
```ts
'use server'

import { object, parse, string } from 'valibot'

const schema = object({
  name: string(),
})

export const submit = async (formData: FormData) => {
  const parsedFormData = parse(schema, {
    name: formData.get('name'),
  })

  // Rest of the code
}

```

I get the error when building:
```
n [ValiError]: Invalid type

...

issues: [
    {
      reason: 'type',
      validation: 'enum',
      origin: 'value',
      message: 'Invalid type',
      input: undefined,
      issues: undefined,
      abortEarly: undefined,
      abortPipeEarly: undefined,
      skipPipe: undefined,
      path: [Array]
    }
  ]

```

fabian-hiller:
Something seems to be wrong here. A `ValiError` should only be thrown when validating data, not when bundling your application. Valibot is just a simple JavaScript library that exports some functions. We have no dependencies. So I assume there is a problem in your code, configuration or build process. Feel free to share a minimal reproduction via Stackblitz so I can debug the problem.

mahersalhani:
> Can you send me your schema? But it seems like this is an external problem.

import { object, string, minLength, toTrimmed, type Output } from 'valibot';

export const DesignFormSchema = object({
  name: string(undefined, [toTrimmed(), minLength(3, 'design.name_required')]),
});

export type DesignFormData = Output<typeof DesignFormSchema>;


When I used "valibot": "^0.19.0" it's worked

fabian-hiller:
[...]

PR #359, opened 2024-01-08
Add peerDependency warning if using < typescript 5

EliBates:
I discovered this issue using nextjs when when using typescript < 5. 
https://github.com/fabian-hiller/valibot/issues/352

This adds a peerDependency of typescript 5 so that anyone installing valibot will be warned if they don't have the appropriate version of TypeScript installed in their project.


fabian-hiller:
The idea is great, but there is one problem. If someone is not using TypeScript, this change will erroneously display an error when installing Valibot.

EliBates:
@fabian-hiller Good point. I think we can add:

```json
  "peerDependenciesMeta": {
    "typescript": {
      "optional": true
    }
  }
```

This should make it warn only users who are using typescript. I think this will be helpful because I still see a lot of projects on <5 typescript.

fabian-hiller:
Thank you! I will review this PR in the next few days.

package.json
```diff
@@ -10,6 +10,14 @@
     "url": "https://github.com/fabian-hiller/valibot"
   },
   "private": true,
+  "peerDependencies": {
+    "typescript": ">=5"
+  },
```
mxdvl, line 15: One thing to note about the TypeScript project is that [it does not follow SemVer](https://www.semver-ts.org/1-background.html#typescripts-versioning-policy), and breaking changes can be introduced in minor versions.

You might want to be more explicit about versions that are compatible with the tilde syntax:

```suggestion
  "peerDependencies": {
    "typescript": "~5.0||~5.1||~5.2||~5.3"
  },
```
Or even just a more specific range:

```suggestion
  "peerDependencies": {
    "typescript": "=5"
  },
```

fabian-hiller:
Sorry for not picking up this PR yet. Other things had a higher priority. I expect to review it this month.

fabian-hiller:
Unfortunately Git screwed something up. I will close this PR and add the changes to the source myself. I will add you as a co-author to the commit.

PR #360, opened 2024-01-09
refactor: flatten

ariskemper:


fabian-hiller:
Thanks for improving the code. I think I don't like to split small functions into even smaller ones when it's not necessary. It increases the bundle size and makes the code more complex in my eyes.

ariskemper:
@fabian-hiller would be good to know how much this change affects the size?

fabian-hiller:
Thank you for the research. I think readability is subjective here and I prefer not to split up small functions unnecessarily.

fabian-hiller:
Since we rewrote this function is the previous rewrite, I will close this PR.

PR #365, opened 2024-01-14
doc(api): add `unknown()` related doc

mwskwong:
Parent issue: https://github.com/fabian-hiller/valibot/issues/287

Created by referencing the `string()` doc. See if this is in the right direction. And whether I have missed anything on the related API part.


fabian-hiller:
Thank you!

PR #366, opened 2024-01-14
refactor: string schema tests

ariskemper:


ariskemper:
@fabian-hiller ready for review and merge?

library/src/schemas/string/stringAsync.test.ts
```diff
@@ -83,17 +83,17 @@ describe('stringAsync', () => {
   describe('schema pipeline', () => {
     const schema1 = stringAsync([minLength(2), maxLength(3)]);
 
-    test('should contain invalid length message, type and requirements', () => {
+    test('should expose the pipeline', () => {
```
fabian-hiller, line 86: I would refactor that and test the entire schema object for the different arguments.

ariskemper: @fabian-hiller so you would like to add more tests for using different arguments, validations ( minLength, maxLength, bic, creditCard .... ) or entire schema properties?

fabian-hiller: I would not distinguish between different pipeline actions. As you can see in the [API reference](https://valibot.dev/api/string/), `string` allows you to pass no argument, only a pipeline or a message and a pipeline. I would just test each combination.

PR #369, opened 2024-01-14
fix #206 : add ipv6 validation method

### LINKED ISSUE #206: ipv6 validation allows invalid ipv6 addresses
The current regexp used for checking ipv6 in valibot might not be correct, as it allows, e.g.:
`fe80::3:bEFf:5b:%3NG` which is not valid.

https://github.com/fabian-hiller/valibot/blob/982045e5597f2906714a9f38674880d8ee649645/library/src/validations/ipv6/ipv6.ts#L13

As per https://github.com/garycourt/uri-js/issues/40#issuecomment-532204167, it looks like other regexps can not be used to fully make sure an ipv6 address is valid and the regexp used in uri-js seems quite expensive (although the issue is from 2019, so it might have changed nowadays).
In any case, the Fastify team is now using a different approach to parse ipv6 addresses: [fastify/fast-uri@main/lib/utils.js#L27](https://github.com/fastify/fast-uri/blob/main/lib/utils.js?rgh-link-date=2023-10-07T17%3A52%3A23Z#L27) (the mentioned "read buffer" approach is used there).

The buffer code is faster, but also increases bundle size a lot more compared to the regex. What do we prefer?

ariskemper:
@fabian-hiller this is potential fix for #206, i added additional ipv6 validation method.


ariskemper:
@fabian-hiller what do you think about PR, will you merge it, you have any questions or should i add the docs to the same PR?

fabian-hiller:
I plan to review it on the weekend

fabian-hiller:
I want to let you know that I have rescheduled the review for next week.

fabian-hiller:
I will close this PR. However, I will keep it in mind when improving our IPv6 validation in the future.

PR #374, opened 2024-01-16
docs: removed comma

shairez:
The comma is unnecessary here and makes the sentence harder to read / understand


fabian-hiller:
Thank you! Isn't this a subordinate clause that must be separated by a comma?

shairez:
Not sure about the scientific side of it 😅
But for me it broke the flow of the sentence, and hence, it's meaning. 
So maybe it's correct to do it that way (I'm no English expert), but in reality for me it didn't achieve its goal IMO, so I would remove it if it were my docs.

PR #383, opened 2024-01-23
Refused promise input to `parse`

genki:
I have faced to a problem to `parse` the promise value mistakenly several times.
And I came up with that the `parse` must not accept any promise values as input.
There must be no needs to `parse` the promise.

### DIFF
--- a/library/src/methods/parse/parse.ts
+++ b/library/src/methods/parse/parse.ts
@@ -10,9 +10,9 @@ import type { BaseSchema, Output, ParseInfo } from '../../types/index.ts';
  *
  * @returns The parsed output.
  */
-export function parse<TSchema extends BaseSchema>(
+export function parse<TSchema extends BaseSchema, TInput>(
   schema: TSchema,
-  input: unknown,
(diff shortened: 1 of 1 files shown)


fabian-hiller:
I am not sure I want to change this. What if we add a `promise` schema function in the future that allows promises to be passed? I understand why you like this change but on the other side I don't want to limit the library unnecessary.

fabian-hiller:
Also, a `Promise' schema can already be created with `instance` and `special`, so this change could also be a breaking change for some users.

genki:
I see, I will respect your design.
For further studies, I would like to know about it.
I though the `parseAsync` was prepared for such the promise values to parse.
I couldn't came up with use cases that we want to parse Promise object it self.
All `instance(Promise)` and `special<Promise>` can't be replaced with `anyAsync`?

fabian-hiller:
The idea of `parseAsync` is that you can use async validation functions like `customAsync` in your pipeline. `any` and `anyAsync` allow any value, not just promises. Read more about async validation here: https://valibot.dev/guides/async-validation/

PR #393, opened 2024-01-26
docs: Fix variable name

morinokami:
I think the name of the variable `value` should not be changed here.


fabian-hiller:
Thank you!

PR #410, opened 2024-02-04
docs: Add Valibot validater for Hono to ecosystem document

chimame:


fabian-hiller:
Thank you for creating this PR. Why would you change it? If there is an official integration, I think it is better to link to the main project or package instead of the integration.

chimame:
I forgot to mention the reason.

I understand that valibot can be used with Hono, but I have clearly stated what library can be used to use valibot. If it is different from the image described in this ecosystem, there is no problem in closing this pull request. 🆗

chimame:
```
- [Hono](https://hono.dev/): Ultrafast web framework for the Edges
  - [@hono/valibot-validator](https://github.com/honojs/middleware/tree/main/packages/valibot-validator): The validator middleware using Valibot for [Hono](https://hono.dev/) applications. You can write a schema with Valibot and validate the incoming values.
```

This description might have been better. However, if this PR differs from the policy, please close it.

chimame:
I misunderstood your intentions. You're pointing out the URL being linked to. change.

ZerNico:
I don't think it's necessary to explicitly link to the Valibot validator. It's an official package for Hono and is actually linked inside the Hono docs in the third-party middleware page.

chimame:
Thanks for your input. So let's close it since it's not necessary.

Issue #373, opened 2024-01-16
Support Metadata like description


fabian-hiller:
### Discussed in https://github.com/fabian-hiller/valibot/discussions/368

<div type='discussions-op-text'>

<sup>Originally posted by **xcfox** January 14, 2024</sup>
Valibot is a super cool Schema Builder! I am attracted by Valibot's modular design.

I plan to use Valibot as an Entity Schema Builder for [MikroORM](https://github.com/mikro-orm/mikro-orm).

So far, I am using `TypeScript`'s `class` and `Decorators` to define MikroORM's [Entity](https://mikro-orm.io/docs/defining-entities) like this:

```typescript
@Entity()
export class BaseEntity {
  @PrimaryKey()
  id!: number

  @Property()
  createdAt = new Date()

  @Property({ onUpdate: () => new Date() })
  updatedAt = new Date()
}

interface WithName {
  name: string
}

@Entity()
export class User extends BaseEntity implements WithName {
  @Property({ unique: true, comment: "user's name" })
  name!: string

  @index()
  @Property({ nullable: true })
  email: string

  @Property({ type: 'text' })
  password!: string
}
```

Because TypeScript's `class` doesn't support multiple inheritance, I have to define the `name` field repeatedly.

Maybe I can use Valibot and custom pipelines to solve this problem:

```typescript
import { intersect, Output, nullable, date, string, object } from 'valibot'
import { primaryKey, property, index, unique, description } from './custom'

const baseEntity = object({
  id: string([primaryKey()]),
  createdAt: date([property()]),
  updatedAt: date([property()]),
})

const withName = object({
  name: string([property(), index(), description("user's name")]),
})

export const userEntity = intersect([
  object({
    email: nullable(string([property(), unique()])),
    password: string([property()]),
  }),
  baseEntity,
  withName,
])

export interface IUser extends Output<typeof userEntity> {}
```

Here I use many custom pipelines: `primaryKey`, `property`, `index`, `unique`, `description`. However, the current version of Valibot's pipeline only supports `Validation` and `Transformation`. I don't think these custom pipes belong to either `Validation` or `Transformation`.

[...]

PR #411, opened 2024-02-04
feat: support metadata

xcfox:
Implements https://github.com/fabian-hiller/valibot/issues/373 

- All schema functions accept the metadata argument:
  - Most schemas accept this argument using the location of the `message`.
  - Additional metadata argument added for some special schemas. `any()`, `unknown()`, `recursive()`
- The metadata argument has been added to the following method:

  - `merge()`
  - `omit()`
  - `partial()`
  - `pick()`
  - `required()`

- Use `isSchema` in `restAndDefaultArgs` to distinguish whether an object is a valid schema or metadata.

- Currently only the `description` field has been added to the metadata.
  We can either follow the [json-schema](https://json-schema.org/understanding-json-schema/reference/generic) completely, or not pre-populate any fields like [yup](https://github.com/jquense/yup?tab=readme-ov-file#schemametametadata-schemametadata-schema), or just a built-in `description` field like [zod](https://zod.dev/?id=describe).


fabian-hiller:
Thank you for creating this PR. I will review it in the next weeks.

xcfox:
@fabian-hiller Thank you! 

I think metadata is an essential feature, and both [yup](https://github.com/jquense/yup?tab=readme-ov-file#schemametametadata-schemametadata-schema) and [zod](https://zod.dev/?id=describe) support it.

Do you think metadata needs to be discussed in more depth in terms of implementation? Do we need to consider the placement of metadata settings or the built-in metadata field?

fabian-hiller:
I am currently focused on other stuff. I will get back to you on the weekend or next week.

jeremyjacob:
Metadata would be super nice for exporting schemas to other formats like OpenAPI.

fabian-hiller:
Can you make an example with dummy code?

vova-white:
> Metadata would be super nice for exporting schemas to other formats like OpenAPI.
> Can you make an example with dummy code?

Maybe something like:

```js
const ProductSchema = pipe(
  object({
    productId: pipe(
      number(),
      integer(),
      minValue(1),
      withDescription("The unique identifier for a product"),
    ),
    productName: pipe(
      string(),
      nonEmpty(),
      withDescription("Name of the product"),
    ),
  }),
  withDescription("A product from Acme's catalog"),
);
```

Will be converted to: 

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/product.schema.json",
  "title": "ProductSchema",
  "description": "A product from Acme's catalog",
  "type": "object",
  "properties": {
    "productId": {
      "description": "The unique identifier for a product",
      "type": "integer"
    },
    "productName": {
      "description": "Name of the product",
      "type": "string"
    }
  }
}
```

I think this will be very useful for those who develop a public API and use, for example, [valibot-json-schema](https://github.com/gcornut/valibot-json-schema) to generate documentation.

PR #416, opened 2024-02-05
feat: add min length for number

ariskemper:


fabian-hiller:
It seems like `minValue` does the job and it is not required to support numbers in `minLength`.

ariskemper:
@fabian-hiller i was not going  in detail, but probably maxValue, minValue solve this and we keep minLength, maxLength for strings, arrays. I could rename the PRs and we keep the refactored tests and i add the docs in the same PR?

fabian-hiller:
Yes, sounds good! Thank you!

fabian-hiller:
I am closing this PR as it has been added recently. Sorry for not building on top of this PR.

Issue #72, opened 2023-08-07
Recursive schema


ZerNico:
Hello there!
At the moment there seems to be no way to use a recursive schema.

Zod for example does it like this:
```typescript
const baseCategorySchema = z.object({
  name: z.string(),
});

type Category = z.infer<typeof baseCategorySchema> & {
  subcategories: Category[];
};

const categorySchema: z.ZodType<Category> = baseCategorySchema.extend({
  subcategories: z.lazy(() => categorySchema.array()),
});
```

Maybe there even is a way to do this without having to manually define a type, I know some validation lib had that feature, I just can't remember which one it was. It might not be possible with the way Valibot works, though.

fabian-hiller:
This should work with [`recursive`](https://github.com/fabian-hiller/valibot/blob/main/library/src/schemas/recursive/recursive.ts).

ZerNico:
Awesome, I even looked through the code, must have totally overlooked it then, thanks!

Is this the right approach to use it then?

```typescript
import { BaseSchema, object, recursive, string, parse, Output, merge, array } from "valibot";

const baseCategorySchema = object({
  name: string(),
});

type Category = Output<typeof baseCategorySchema> & {
  subcategories?: Category[];
};

const categorySchema: BaseSchema<Category> = merge([baseCategorySchema, object({ subcategories: recursive(() => array(categorySchema)) })]);

const result = parse(categorySchema, {
  name: "People",
  subcategories: [
    {
      name: "Politicians",
      subcategories: [
        {
          name: "Presidents",
          subcategories: [],
        },
      ],
    },
  ],
});
```

fabian-hiller:
Looks good. Does it work as expected?

ZerNico:
Jup seems to work fine. An example like this could probably be added to the docs then.
This is exactly the example from Zod but converted to Valibot.

fabian-hiller:
As soon as I find the time, I'll probably add advanced guides to the docs and include that. Thanks for your contribution!

firxworx:
Random (related) question/confirmation as I check out valibot for the first time :) ...

I was curious to see if the following would be the right approach to parsing a general JSON value with Valibot?

```ts
import {
  type BaseSchema,
  type Output,
  union,
  boolean,
  number,
  string,
  nullType,
  recursive,
  array,
  record,
} from 'valibot'

export type JsonPrimitive = Output<typeof vJsonPrimitive>
[...]

PR #424, opened 2024-02-06
docs(recursive): add category example + include imports for easy copy…

niconiahi:
in this PR I add a new example for `recursive`. Imports are included to allow quick copy/paste, otherwise the user has to include one by one. Also it gives a better idea of the real amount of imports it's going to be required

the example is the one mentioned in #72 

does this sound interesting to be added? in the end I think _it is_ because it's a more "feets on the ground" example comparing it to something like a binary tree example (the only current one)


fabian-hiller:
Thank you! I will probably review it on the weekend or next week.

PR #441, opened 2024-02-16
Add support for `input` as value in `v.recursive()` getter

### LINKED ISSUE #440: Enable `v.recursive()` getter to accept a `value` to improving usage flexibility
We sometimes need to adjust schemas based on input values. 

[`yup.lazy()`](https://github.com/jquense/yup?tab=readme-ov-file#lazyvalue-any--schema-lazy) lets us do this by using a getter function to choose different schemas depending on the input type, as shown below:

```typescript
const node = yup.object({
  id: yup.number(),
  child: yup.lazy(() => node.default(undefined)),
});

let renderable = yup.lazy((value) => {
  switch (typeof value) {
    case 'number':
      return yup.number();
    case 'string':
      return yup.string();
    default:
      return yup.mixed();
  }
});

const renderables = yup.array().of(renderable);
```

This is great for flexible and dynamic schemas. I'm wondering if `valibot` is thinking about adding something similar for `recursive` methods, like this:

```typescript
const schema = v.recursive((value) => {
  switch (typeof value) {
    case 'number':
      return v.number();
    case 'string':
      return v.string();
    default:
      return v.any();
  }
});
```

This would make schemas even more flexible for complex data. If `valibot` is open to it, I'd love to help by contributing this feature.

Mini-ghost:
### Linked issue

Resolve #440 

### Description

Currently, when we aim to create a valibot schema that depends on real-time data, the following code snippet is a standard approach:

```ts
const input = {
  // Your input here
};

const schema = v.recursive(() => {
  if(input.type === 'A') {
    return v.object({
      foo: v.string()
    });
  }

  return v.object({
    Column: v.number()
  });
});

const output = v.parse(schema, input);
```

However, this method presents challenges in abstracting the schema effectively.

If the `v.recursive()` function could accept a parameter—specifically, the second parameter passed to `parse()`—we would be able to abstract the schema with greater ease, as shown below:

```ts
// schema.ts
const schema = v.recursive((input) => {
  if(input.type === 'A') {
    return v.object({
      foo: v.string()
    });
  }

  return v.object({
    Column: v.number()
  });
});

// validate.ts
const input = {
  // Your input here
};

const output = v.parse(schema, input);
```

### DIFF
--- a/library/CHANGELOG.md
+++ b/library/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the library will be documented in this file.
 
+## vX.X.X (Month DD, YYYY)
+
+- Add `input` of schema to `getter` function of `recursive` and `recursiveAsync` schema (pull request #441)
+
 ## v0.28.1 (February 06, 2024)
 
 - Fix bug in `union` and `unionAsync` schema for transformed inputs (issue #420)
--- a/library/src/schemas/recursive/recursive.test.ts
+++ b/library/src/schemas/recursive/recursive.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, test } from 'vitest';
+import { describe, expect, test, vi } from 'vitest';
 import { parse } from '../../methods/index.ts';
 import { minLength } from '../../validations/index.ts';
 import { string } from '../string/index.ts';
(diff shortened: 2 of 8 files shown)


fabian-hiller:
Thank you!

PR #445, opened 2024-02-20
docs: fixed command k prevent default

### LINKED ISSUE #444: [docs] Command + K doesn't prevent default, opening the URL bar in Firefox
In Firefox, doing Command + K will focus the URL bar to search things on Google. When using that same shortcut for site-wise search, the event needs to be prevent defaulted to avoid passing on the event to Firefox.

image

In a traditional JS framework such as React or Vue, this would be an easy fix but by the nature of the Qwik framework, I'm not exactly sure how to fix this myself for a PR.

CapitaineToinon:
Using the beta feature `sync$` to conditionally preventDefault on the command K search. Fixes #444


fabian-hiller:
Thanks for creating this issue! I need to investigate first to see if upgrading to the latest Qwik version breaks anything.

PR #462, opened 2024-02-29
fix: `minValue`/`maxValue` intput/output types

### LINKED ISSUE #266: Incorrect type inference in v0.21.0
Hello,

I was able to use this 👇 code to parse a string into a number in v0.20.1 and get the correct type for `n` (`number`):

```ts
const n = parse(
  transform(string(), (input) => parseInt(input, 10), [minValue(0)]),
  "1",
);
```

But when I updated to v0.21.0, the inferred type of `n` of the same code became `string | number | bigint | boolean | Date`.

I have to change it to this 👇 to fix the type.

```ts
const n = parse(
  transform(string(), (input) => parseInt(input, 10), [minValue<number, number>(0)]),
  "1",
);
```

Did I miss something?

Thanks

Sec-ant:
Before:
image

After:
image

Fixes #266.


Sec-ant:
I just noticed `value`, `notValue` and `equal` validations have similar type problems like this. I wonder if this fix is the right direction, so I can try to fix those validations, too. @fabian-hiller

fabian-hiller:
I will first check to see if the issue #266 still exists after we change our API to this proposal: https://github.com/fabian-hiller/valibot/discussions/463

fabian-hiller:
I think this was solved with #502

PR #465, opened 2024-03-01
Enforce interface consistency with ESLint

romeerez:
Add `@typescript-eslint/consistent-type-definitions` ESLint rule to enforce interfaces over types for better TS performance.

I didn't run `eslint --fix` because it makes changes to a lot of files, reducing the possibility for this PR to be merged.

Added `eslint:fix` script to package.json. After merging this PR, you can run `eslint:fix` to automatically apply the new rule.

Changes in `flatten.ts` are needed because otherwise, TS would complain about the types when `BaseSchema` becomes an interface.

This PR on it's own fixes 'type instantiation is too deep' error caused by this reproduction:

<details>
<summary>reproduction</summary>

```ts
import {
  array,
  boolean,
  merge,
  number,
  object,
  optional,
  parse,
  partial,
  pick,
  string,
} from './dist/index.js';
// this file is located in library/check.ts and depends on built output of valibot

const schema = merge([
  partial(
    merge([
      pick(
        object({
          id: number(),
          userId: number(),
          slug: string(),
          title: string(),
          body: string(),
          favoritesCount: number(),
          createdAt: string(),
          updatedAt: string(),
        }),
        ['slug', 'title', 'body']
      ),
      object({ tags: array(string()) }),
    ])
  ),
  object({ favorite: optional(boolean()) }),
]);

const result = parse(schema, {});
```
</details>

The rule only converts `type` syntax to `interface` when appropriate, but it cannot convert such code:

```ts
type X = Omit<Y> & { add: SomeType }
// to:
interface X extends Y { add: SomeType }
```

This conversion is valuable for TS performance.


fabian-hiller:
Thank you for creating this issue!

PR #468, opened 2024-03-04
[i18n] Allow valibot v0.29 in peer dependencies

siketyan:
valibot v0.29.0 has been released but installing with `@valibot/i18n` causes peer dependency failure because the package requires v0.28.x. This pull request adds support of valibot v0.29.x to the i18n package.

### DIFF
--- a/packages/i18n/package.json
+++ b/packages/i18n/package.json
@@ -34,7 +34,7 @@
     "valibot": "workspace:*"
   },
   "peerDependencies": {
-    "valibot": "^0.28.0"
+    "valibot": ">=0.29.0 <1"
   },
   "files": [
     "index.js",


fabian-hiller:
Thank you for creating this issue. Shouldn't the current specification already cover every 0.X.X version? Can you send me your package manager and error message?

siketyan:
@fabian-hiller

I am using pnpm and got the following error:

image


In SemVer, `^0.28.0` is defined as `>= 0.28.0, < 0.29.0` (For 1.x or later it is `>= 1.x.y, < 2.0.0`). Thus we need to add `^0.29.0` explicitly or allow entire 0.x `>= 0.28.0, < 1.0.0`. Personally I prefer the former style because the interface(s) for i18n may change during 0.x.

fabian-hiller:
Thanks for your research! I prefer to write `>=0.29.0 <1.0.0` (there was an important bug fix in `v0.29.0`), otherwise I have to release a new version for `@valibot/i18n` every time I release a new `valibot` version. I don't expect an API change for i18n.

siketyan:
@fabian-hiller
Changed the requirement to `>=0.29.0 <1` in 3baac9a.

fabian-hiller:
Thank you so much!

PR #496, opened 2024-03-28
feat: Validate ISO timestamps with offsets

### LINKED ISSUE #495: `isoTimestamp` only accepts UTC timestamps
It looks like `ISO_TIMESTAMP_REGEX` validates only UTC timestamps (ending with `Z`). But the standard also [defines the possibility of having offsets](https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC).

From my understanding, the following formats should all be considered as valid ISO timestamps:
- `<time>Z`
- `<time>±hh:mm`
- `<time>±hhmm`
- `<time>±hh`

If you consider it a bug, I'd happily open a PR 🙂

macarie:
I've added a bunch of tests with both valid and invalid timestamps and updated the JSDoc to list all the allowed formats.

The regex is relatively permissive as it allows for any number in the offset. If something stricter is required, I can work on it, but I'm afraid it would explode in complexity. For example, a `00:00` offset is valid only if preceded by a `+` and not by `-`; the maximum offset should be ±12:00, etc.

Closes #495.

### DIFF
--- a/library/src/regex.ts
+++ b/library/src/regex.ts
@@ -79,7 +79,7 @@ export const ISO_TIME_SECOND_REGEX = /^(?:0\d|1\d|2[0-3])(?::[0-5]\d){2}$/u;
  * [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp regex.
  */
 export const ISO_TIMESTAMP_REGEX =
-  /^\d{4}-(?:0[1-9]|1[0-2])-(?:[12]\d|0[1-9]|3[01])T(?:0\d|1\d|2[0-3])(?::[0-5]\d){2}(?:\.\d{1,9})?Z$/u;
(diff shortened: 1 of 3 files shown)


fabian-hiller:
Thank you for creating this PR. Since I am currently busy rewriting the library based on #463, it may take some time before I release the next version. So I don't recommend to wait and use the `regex` function instead or add a custom `isoTimestamp` action to your own codebase.

macarie:
Sure, no problem! 😄

brandonpittman (approved):
Let's get this released.

mxdvl:
Would you like me to port this changes as part of #502?

This would technically be a breaking change, as inputs that would previously throw an error may now pass, so bundling it with other breaking changes might be appropriate.

fabian-hiller:
Feel free to contribute to #502, but don't go crazy, because I don't want to spend too much time on new features. I want to get the next version out as soon as possible. Until v1, breaking changes are ok.

fabian-hiller:
Thank you! Merged in #530.

PR #501, opened 2024-04-02
Fix isoTimestamp regex

### LINKED ISSUE #500: Timestamp regex doesn't support timezone other than Zulu
https://github.com/fabian-hiller/valibot/blob/6ca70b2a24eadf0e7e00923fa3b75f5657d0b8f5/library/src/regex.ts#L81-L82

Should support a time like `2024-04-02T08:52:31+07:00`.

brandonpittman:
To allow timezones

Closes #500


fabian-hiller:
This is a duplicate of #496. Feel free to add a review to this PR.

PR #502, opened 2024-04-03
Rewrite Valibot with the `pipe` function

fabian-hiller:
One month after I published #463 to discuss Valibot's API, it is finally time to share my first draft with you. I have not only implemented the `pipe` functions. In the end, I basically rewrote the whole library. I improved many things while keeping performance, bundle size and developer experience in mind.

> Except for the `pipe` function, the basic API remains the same. However, to make the migration as smooth as possible for you, we plan to partner with [Codemod](https://codemod.com/) to automatically migrate most of the changes with a single CLI command.

## `pipe` function

First of all, thank you for your feedback! It was amazing to see how the Valibot community came together to discuss this change. In total, my post on X got more than 15,000 impressions and more than 70 developers participated on GitHub.

### Better mental model

The `pipe` function reduces the mental model of Valibot to schemas, methods and actions. Schemas validate data types like strings, numbers and objects. Methods are small utilities that help you use or modify a schema. For example, the `pipe` function is a "method" that adds a pipeline to a schema. Actions are used exclusively within a pipeline to further validate or transform a particular data type.

### Simplified source code

By moving the pipeline into its own function, we made Valibot even more modular. This has several advantages. The most obvious is that it reduces the bundle size if your schemas do not require this feature. But even if you use the pipe function, you can expect an average 15-20% reduction in bundle size with the new implementation.

Another great benefit is that we were able to simplify the schema functions by removing the `pipe` argument. Schema functions are now more independent and only cover the validation of their data type. This allows us to simplify the unit tests, making the library even safer to use.

Furthermore, for any primitive data type schema such as `string`, we could remove its async `stringAsync` implementation, since this was only necessary to support async validation within the `pipe` argument. This will further reduce the overall bundle size of the library.

### More flexibility and control

The `pipe` function also adds flexibility to the library and gives you more control over your schemas. Previously, it was very difficult to extend the pipeline of a schema with additional validations or transformations. This changes with the `pipe` functions because they can be nested, making your code more composable.

```ts
import * as v from 'valibot';

const EmailSchema = v.pipe(v.string(), v.email());
const GmailSchema = v.pipe(EmailSchema, v.endsWith('@gmail.com'));
```

Another great benefit is that pipelines now support data type transformations. There is no longer any confusion between a transformation method and pipeline transformations. You can even add another schema function after a transformation to validate its output.

```ts
import * as v from 'valibot';

const PixelSchema = v.pipe(
  v.string(),
  v.regex(/^\d+px$/),
  v.transform(parseInt),
  v.number(),
  v.maxValue(100)
);
```

When building forms, this also gives you more control over the input and output type of a schema. It is now possible for a field to have an input type of `string | null` but an output type of `string` without specifying a default value.

```ts
import * as v from 'valibot';

// This schema
const ProfileSchema = v.object({
  name: v.string(),
  age: v.pipe(v.nullable(v.number()), v.number()),
  bio: v.pipe(v.nullable(v.string()), v.string()),
});

// Has this input type
type ProfileInput = {
  name: string;
  age: number | null;
  bio: string | null;
};

// But this output type
type ProfileOutput = {
  name: string;
  age: number;
  bio: string;
};
```

## `object` schema

Another major change is that I have removed the `rest` argument from the `object` schema. This reduces the bundle size by up to 150 bytes if you don't need this functionality. If you do need it, you can use the newly added `objectWithRest` schema, and to allow or not allow unknown entries, there is now a special `looseObject` and `strictObject` schema for a better developer experience.

```ts
import * as v from 'valibot';

const NormalObjectSchema = v.object({ ... });
const ObjectWithRestSchema = v.objectWithRest({ ... }, v.string());
const LooseObjectSchema = v.looseObject({ ... });
const StrictObjectSchema = v.strictObject({ ... });
```

> I also plan to remove the `rest` argument from the `tuple` schema and provide a `tupleWithRest`, `looseTuple` and `strictTuple` schema.

## Type safety

I am happy to announce that I have been able to drastically improve the type safety of the library. Previously, like many other libraries out there, less important parts were only generally typed.

A concrete example is the issues that Valibot returns when the data does not match your schema. Even though each issue contains very specific data depending on the schema and validation function, they were all previously typed with a generic `SchemaIssue` type.

With this rewrite, each schema and validation function brings its own issues type. This way each schema knows exactly what kind of issues can occur. We also added the `isOfKind` and `isOfType` util functions to filter and handle specific objects in a type-safe way.

valibot

## Unit tests

Valibot was kind of my first project that came with unit testing. When I implemented the first tests, I had very little experience with testing. Even though the current tests cover 100% of the library, I don't feel comfortable releasing v1 before rewriting them and making them perfect. Also, we have had some type issues in the past. That's why I'm also planning to add type testing to ensure the developer experience when using TypeScript.

It is important to me to ensure the functionality of the library and to make sure that no unexpected bugs occur when changes are made to the code in the future. Developers should be able to fully rely on Valibot. If you are a testing expert, please have a look and review the `string` and `object` schema tests.

## Next steps

I created this draft PR for two reasons. One is to get feedback from the community before rewriting the entire library. I have already discussed a lot of details with @Demivan and think we are on a good path, but maybe we missed something. Feel free to investigate the source code and reach out via the comments.

With the following commands you can clone, bundle and play with the new source code in advance:

```bash
# Clone repository
git clone git@github.com:fabian-hiller/valibot.git

# Switch branch
git switch rewrite-with-pipe

# Install dependencies
cd ./valibot && pnpm install

# Bundle library
cd ./library && pnpm build

# Modify `playground.ts`

# Run playground
pnpm play
```

On the other hand, I created this PR also to ask you to help me implement the missing parts. If you are passionate about open source and want to join an exponentially growing open source project, now might be the time.

Starting April 8, everyone is welcome to take over the implementation of a schema or action function. I recommend starting with a simple function like `minBytes`, where you only have to copy and modify the `minLength` action, and maybe look at the previous implementation of the same function.

To participate, just add a comment to this PR with the part you want to take over. Please only add a comment if you are able to do the implementation within 3 days, to not block the rewrite. After you get my thumbs up, you can create a PR that merges your changes into the `rewrite-with-pipe` branch.

The goal is to have everything implemented this month. ⚡️


Afsoon:
I want to contribute, but I have a question. You say the following:

>  I recommend starting with a simple function like minBytes, where you only have to copy and modify the minLength action, and maybe look at the previous implementation of the same function.

But I have noticed the action has missing tests. Is it intentional?

fabian-hiller:
> But I have noticed the action has missing tests. Is it intentional?

I plan to add any missing tests over the weekend. We should be ready to go on April 8.

IlyaSemenov:
I noticed that you're using snake_case for validation type, for example:

```ts
export function minLength(...) { // <------- camelCase function name
  return {
    kind: 'validation',
    type: 'min_length', // <----- snake_case rephrase of the same name
    ...,
  };
}
```

What do we lose if we have a rule to keep `validation.type` to be always exactly the same as the function name?

fabian-hiller:
Thank you for your review! I personally prefer snake_case for string values because it is easier to read and looks more professional (just personal opinion) then camelCase. In terms of bundle size, the difference is only about 2 bytes. Why would you prefer to use the same name as the function name? To make it "uniform"?

IlyaSemenov:
Personally, I am also a big fan of snake_case (and tabs for indentation, for that matter). I have to suffer this love for consistency within Node.js ecosphere.

However, in this case my point was different: to reduce the number of 'different' things/names a developer has to invent / to come up with in order to implement a new method. Currently, it's 4 elements:
1. `maxLength` for the actual method name
2. `MaxLengthAction` for the action interface (essentially a 1, capitalized and suffixed).
3. `MaxLengthIssue` for the issue interface (essentially a 1, capitalized and suffixed).
4. `max_length` for the type key

In my thinking, the last one is easily avoidable as (1) could be used for (4), presumably without downsides. That's not a big deal, just a heads up to give you some thinking. If you still think it's good to have it different, so be it!

fabian-hiller:
Thank you for your detailed reply. I will think about it. It would be nice to get feedback from more developers, but it probably does not matter to most developers.

Saeris:
> I noticed that you're using snake_case for validation type, for example:
> 
> ```ts
> export function minLength(...) { // <------- camelCase function name
>   return {
>     kind: 'validation',
>     type: 'min_length', // <----- snake_case rephrase of the same name
>     ...,
>   };
> }
> ```
> 
> What do we lose if we have a rule to keep `validation.type` to be always exactly the same as the function name?

Another way to further reduce the overall bundle size, ironically, is to change the return values of these functions to class instances rather than plain objects for the simple fact that we can then replace both `kind` and `type` properties by using `instanceof` to check for heritage.

Okay, that may sound dumb, but hear me out! Here's an example from some exploratory work I did prior to #259 getting merged:

`src/types/schema.ts`
```ts
/**
 * Base schema type.
 */
export abstract class Schema<TInput = any, TOutput = TInput> {
  /**
   * Input and output type.
   *
   * @internal
   */
  _types?: { input: TInput; output: TOutput };
  /**
   * Parses unknown input based on its schema.
   *
   * @param input The input to be parsed.
   * @param info The parse info.
   *
   * @returns The parse result.
   *
   * @internal
   */
  abstract _parse(input: unknown, info?: ParseInfo): SchemaResult<TOutput>;
}
```
`src/schemas/boolean/boolean.ts`
```ts
export class BooleanSchema<TOutput = boolean> extends Schema<boolean, TOutput> {
  /**
   * The error message.
   */
  message: ErrorMessage;

  constructor(message?: ErrorMessage) {
    super();
    this.message = message;
  }

  _parse(input: unknown, info?: ParseInfo) {
    // ... schema parsing logic
  }
}

export const boolean = (message?: ErrorMessage) => new BooleanSchema(message);
```

With the above, we can use `instanceof` to both check whether the return value of `boolean()` is a `Schema` or more specifically, a `BooleanSchema`. Additionally, it enables us to consolidate the `interface` types we had before into the class declaration itself, reducing the overall amount of code that needs to be maintained in the library.

You can then use the returned class instance the exact same way you are now with the object these functions return. We only keep the original function for the convenience of not needing to use `new` all of the time.

Anyways, food for thought! Might not be the best way to go about things, but I think it's worth considering to further reduce the codebase and bundle size.

fabian-hiller:
I have looked at your code and currently think that it is more complicated when using classes. Also, the bundle size has increased by about 20% in my investigations.

`boolean` with class instance (147 bytes GZIP):

```ts
export class BaseSchema {}

export class BooleanSchema extends BaseSchema {
  message;
  expects = 'boolean';
  constructor(message) {
    this.message = message;
  }
  _run(dataset, config) {
    // ...
  }
}

export const boolean = (message) => new BooleanSchema(message);
```

`boolean` with plain object (120 bytes GZIP):

```ts
export function boolean(message) {
  return {
    kind: 'schema',
    type: 'boolean',
    expects: 'boolean',
    async: false,
    message,
    _run(dataset, config) {
      // ...
    },
  };
}
```

fabian-hiller:
I need 1 more day to add each missing unit and type test.

fabian-hiller:
I will add the missing tests for the methods and schemas tomorrow, but anyone who is interested can already start helping me port the previous validations and transformations to the new `/actions` directory. I recommend to copy a similar existing action (e.g. `decimal` for regex based validations like `email`), rename everything and if necessary port over the validation or transformation logic from the previous code. I am happy to help and guide you if you have any questions.

xcfox:
@fabian-hiller  I would like to help with this refactoring. I've completed `trimEnd` and `trimStart` at https://github.com/fabian-hiller/valibot/pull/511 .

EltonLobo07:
@fabian-hiller I would like to work on `excludes`.

IlyaSemenov:
I can take `integer`.

xcfox:
I have completed `upperCase` and `lowerCase` at https://github.com/fabian-hiller/valibot/pull/513

mxdvl:
Could I take the `tuple` and related methods?

fabian-hiller:
At the moment I think it is better if I take care of the schemas and more complex functions and the community helps to create the actions to not block the rewrite. But if I am too slow I might change my mind next week.

EltonLobo07:
@fabian-hiller I would like to work on `bytes` and `notBytes` (validations).

EltonLobo07:
@fabian-hiller I have completed `bytes` and `notBytes` at #517 . I would like to work on `minSize` and `maxSize` validations next.

fabian-hiller:
I have merged each PR. 🧩 Thank you so much for your help! Feel free to port over more actions!

Afsoon:
I have just created a new PR migrating Credit Card action https://github.com/fabian-hiller/valibot/pull/519

EltonLobo07:
@fabian-hiller I have completed `minSize` and `maxSize` at #521 . I would like to work on `email` validation next.

Afsoon:
This PR reimplement `cuid2` https://github.com/fabian-hiller/valibot/pull/522

mxdvl:
Implementation of BIC (ISO 9362) in #524

library/src/actions/creditCard/creditCard.ts
```diff
@@ -0,0 +1,143 @@
...
+    message,
+    requirement(input) {
+      let sanitized: string | undefined;
+      return (CREDIT_CARD_REGEX.test(input) &&
+        // Remove any hyphens and blanks
+        (sanitized = input.replace(SANITIZE_REGEX, '')) &&
+        // Check if it matches a provider
+        PROVIDER_REGEX_LIST.some((regex) => regex.test(sanitized)) &&
+        // Check if passes luhn algorithm
+        _isLuhnAlgo(sanitized)) as boolean;
```
mxdvl, line 130: If you coerce the string value to a boolean with the [double NOT operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT#double_not_!!) to prevent the `as boolean` casting

```suggestion
        !!(sanitized = input.replace(SANITIZE_REGEX, '')) &&
        // Check if it matches a provider
        PROVIDER_REGEX_LIST.some((regex) => regex.test(sanitized)) &&
        // Check if passes luhn algorithm
        _isLuhnAlgo(sanitized));
```

fabian-hiller: Thank you for your review! I considered that, but this code always returns a boolean. So `as boolean` feels right in this case and adds no runtime code.

mxdvl: It actually returns `'' | boolean` if the [input or sanitized is an empty string](https://www.typescriptlang.org/play?target=9#code/MYewdgzgLgBAlmADgV1gXhgIgIYCNgAmApgGaYBQANkbBNmHFHAF5EEwaYXmiSwBu2SsiIcYACjoMmrdhgQooAOgBORRJWzAiAQUqVxAeiWGA5gBoYAcisBKWzABkjmAFlsUABar6BEAFtxBwA+GAAGJQBWch5wCBBqJUoQU3EAbxhBYVEAX1sAbiA)

fabian-hiller: The previous `CREDIT_CARD_REGEX.test(input)` check ensures that `sanitized` can't be an empty string. Therefore, I think it is technically impossible for a string to be returned.

devcaeg:
@fabian-hiller I am very interested in the new types functionality to be able to accurately get the errors that a schema may have. This is very useful for end-to-end type safety frameworks. Is it possible to test this functionality in some way at the moment?

fabian-hiller:
@devcaeg yes, with the following commands you can clone, bundle and play with the new source code in advance:

```bash
# Clone repository
git clone git@github.com:fabian-hiller/valibot.git

# Switch branch
git switch rewrite-with-pipe

# Install dependencies
cd ./valibot && pnpm install

# Bundle library
cd ./library && pnpm build

# Modify `playground.ts`

# Run playground
pnpm play
```

EltonLobo07:
@fabian-hiller I have completed `email` at #528 . I would like to work on `hexColor` and `hexdecimal` validations next.

Afsoon:
Here is `ip`, `ipv4` and `ipv6` https://github.com/fabian-hiller/valibot/pull/534. I would like to work next on `notValue`, `notLength` and `notSize`

fabian-hiller:
Thank you all for your help! I will try to merge all current PR related to this draft PR this weekend.

EltonLobo07:
@fabian-hiller I have completed `hexColor` and `hexadecimal` at #533 and #535. I would like to work on `some`, `every` and `custom`

fabian-hiller:
`custom` is renamed to `check`. This better matches what this action does.

EltonLobo07:
@fabian-hiller Can you check #540 ? Its related to `every`, `some` and `check` rewrite.

fabian-hiller:
I will try to merge all open PRs related to this draft PR in the next 24 hours.

fabian-hiller:
Sorry, some other stuff came up and I wasn't able to review and merge the PRs. I will try to work on it as soon as possible.

fabian-hiller:
All PRs that ported previous actions are merged! 🎉 Thanks a lot for your help! 🙏 Feel free to continue porting the missing actions.

EltonLobo07:
@fabian-hiller I would like to work on `startsWith`, `endsWith` and `multipleOf`.

fabian-hiller:
I expect to finish the rewrite in the next two weeks. After that, we need to update the docs and create a codemod together with codemod.com for an easy migration to the new version. I will keep you posted.

EltonLobo07:
@fabian-hiller I would like to work on `emoji`, `hash`, `length` & `octal`.

EltonLobo07:
Based on my comparison, here are the remaining actions that need to be rewritten: `imei`, `isoDate`, `isoDateTime`, `isoTimeSecond`, `isoWeek`, `mimeType`, `regex`, `size`, `ulid`, `uuid` and `value`. @fabian-hiller Can you confirm - is the list and the name of each action correct? If not, can you reply with an updated list? After that, maybe contributors can reply by listing down the actions they want to work on and update the list of remaining actions. This will make sure everyone is aware, making it easy to pick actions.

fabian-hiller:
Thank you! The names are correct. I will check later if anything else is missing.

fabian-hiller:
Your list looks good. Thank you for your contributions to the rewrite!

fabian-hiller:
Thanks to everyone who contributed to the rewrite. We are now in the home stretch. 🏎️ I expect to finish the rewrite next week. 🏁

mxdvl:
Could I contribute the iso* date/time functions?

EltonLobo07:
@fabian-hiller I would like to work on `size` action.

EltonLobo07:
Remaining actions: `imei`, `mimeType`, `regex`, `ulid`, `uuid` and `value`.

AndreyYolkin:
I'm going to bake uuid and ulid today

fabian-hiller:
With `intersect` I added the last missing schema today. 🎉 Tomorrow or Saturday I will add the last missing actions and continue with the methods. Early adopters can already start bundling and using the next version of Valibot. I am happy to help with any questions! 🧩

EltonLobo07:
Hi @fabian-hiller, I can help rewrite `mimeType`, `regex`, and `vaue` actions if you haven't started working on any of the mentioned actions so that you can focus on rewriting methods and quickly complete the entire Valibot rewrite. Let me know.

MrGeniusProgrammer:
@LogicalAlpha7Programmer and I would like to help rewrite the `imei` action.

IlyaSemenov:
I believe the new API should include some explanatory document alongside it, or at the very least better JSDoc coming with `BaseValidation`, `TypedDataset` and friends. Currently, there is a steep learning curve on how to come up with own actions, and the code of existing actions is somewhat cryptic (thus it's not even clear whether it's actually correct).

Consider the code of two supposedly trivial validators:

https://github.com/fabian-hiller/valibot/blob/07c15f611112cecd1799e3f9e61933185d559b71/library/src/actions/minLength/minLength.ts#L109-L124

https://github.com/fabian-hiller/valibot/blob/07c15f611112cecd1799e3f9e61933185d559b71/library/src/actions/ip/ip.ts#L87-L101

So what strikes immediately is:
- Both are 10+-liners doing the supposedly trivial validation (checking length/running regexp).
- Looks like they're mostly copy/paste of each other.
- Half of the fields don't make immediate sense, and looking up JSDoc doesn't really help:
  - What's `expects`? Based on JSDoc, it's _"The expected property."_ — seriously, this is not helping at all. The documentation should make it clear when a developer is supposed to put null and when a string, and where does this string go (i.e. what's the proper way to format it).
  - What's `dataset.typed`? Based on JSDoc, it's `_"The typed dataset type. -> Whether is's typed."_ This doesn't bring any value added. What's dataset and what does it mean that it's typed?
    - Both actions don't do `_addIssue` if dataset is not typed. Should the developer always copy/paste that condition?
  - What's `requirement`? It's not present in the base interface (`BaseValidation`), so...should the developer always use it, or it's up to the particular action? What's the recommended flow? How's it different from the code which is actually executed in `_run`? In `url.ts`, the `requirement` is a function which is then called as `this.requirement`, and in `minLength` it's the actual value. How's that publicly exported function possibly helping?
- Why does `minLength` add ```{ received: `${dataset.value.length}` }``` and `url` doesn't? What's that at all? Based on JSDoc, it's the parameter named `other` described as _"The optional props,"_ which isn't really helping. What's the guidance on when to use it?
 
Besides better documentation/developer guide, what if we add some kind of helpers/wrappers for the basic/most popular types of actions, reducing/wrapping the boilerplate? For example, it could be (just a quick attempt — not much thinking given into it):

```ts
export function minLength(
  requirement: number,
  message?: ErrorMessage<MinLengthIssue<LengthInput, number>>
): MinLengthAction<
  LengthInput,
  number,
  ErrorMessage<MinLengthIssue<LengthInput, number>> | undefined
> {
  return validator({
    kind: 'validation',
    type: 'min_length',
    async: false,
    expects: `>=${requirement}`,
    requirement,
    message,
    run(value) {
      if (value.length < this.requirement) {
        return [{ type: 'length', received: `${dataset.value.length}` }]
      }
    },
  })
}
```

fabian-hiller:
> I believe the new API should include some explanatory document alongside it. Currently, there is a steep learning curve on how to come up with own actions, and the code of existing actions is somewhat cryptic.

I will soon be documented in detail on our website.

> or at the very least better JSDoc

With that I am careful until we are sure about everything to not blew up the maintenance work.

> What's `expects`? Based on JSDoc, it's _"The expected property."

`expects`, `expected` and `received` are language-neutral strings that can be used to dynamically generate rich error messages. The language neutrality allows you to easily display error messages for different languages. `expects` and `expected` is usually `null` for regex based validations. `null` basically means "do not include this parameter in the error message". Check out the code for [`_addIssue`](https://github.com/fabian-hiller/valibot/blob/rewrite-with-pipe/library/src/utils/_addIssue/_addIssue.ts#L69-L71).

For example, `email` will output `Invalid email: Received "jane@example..com"` but `minLength` will return `Invalid length: Expected >=1 but received 0`.

> What's `dataset.typed`? Based on JSDoc, it's `_"The typed dataset type. -> Whether is's typed."

`typed` means that the data matches the type definition of the schema. This information brings several benefits to the library. For example, it improves the returned issues of `union` and `variant` and allows us to skip actions that require the data to be typed. For example, we cannot validate the length of an array if we are not sure that the input is actually an array. In the future, there will be actions that can validate untyped data if some fields you specify are typed. This will be necessary for complex form validation.

> What's `requirement`? It's not present in the base interface (`BaseValidation`)

Valibot tries to expose all information to make the schemas "readable" from the outside. The `requirement` property is used to validate the data that is passed to an action. A third party library could extract this information to generate HTML input attributes such as `type`, `min`, `max`, and `pattern`.

I removed `requirement` from `BaseValidation` for some reason. Unfortunately, I can't remember. I will check again.

> Why does `minLength` add ```{ received: `${dataset.value.length}` }``` and `url` doesn't?

The optional properties of the `other` argument allow us to override certain properties of an issue. ```{ received: `${dataset.value.length}` }``` overrides the default value of `received` which is generated based on the input value.

> Besides better documentation/developer guide, what if we add some kind of helpers/wrappers for the basic/most popular types of actions, reducing/wrapping the boilerplate?

In my investigations, this only increases the initial bundle size of the library. That's why I decided against it.

Feel free to reach out if you have further questions, but keep in mind that the library is still under active development and the documentation will be updated soon.

fabian-hiller:
I think the `notLength`, `notValue` and `notSize` actions are also still missing

EltonLobo07:
Yes, they are still missing. @Afsoon is rewriting those actions. Mentioned here: https://github.com/fabian-hiller/valibot/pull/502#issuecomment-2063168781

Afsoon:
Sorry for not saying it before, I have been busy planning my wedding and I don't have enough bandwidth to do it. They are free to pick by anyone @EltonLobo07

devcaeg:
@fabian-hiller 

`Valibot` in all `schemas` contains an internal function `_parse` which could be executed with `schema._parse(value)` but I have seen that in this rewrite of the library, `_parse` is no longer found.

`_parse` is very useful for libraries that want to use `Valibot` optionally without having to import internally `parse` or `safeParse` from `Valibot`.

fabian-hiller:
@devcaeg `._parse` has been renamed to `._run`, but the functionality remains similar. Feel free to reach out if you have further questions.

devcaeg:
@fabian-hiller 

`_run` receives two parameters, the second parameter is an object that has all its properties optional, but it is not initialized by default, so currently it is mandatory to pass the second parameter even if it is an empty object.

In the first parameter that receives `_run` is an object that has the property `typed` which is a `boolean`, what does it affect if it is `true` or `false`?

znycheporuk:
@devcaeg, from comment above:

`typed` means that the data matches the type definition of the schema. This information brings several benefits to the library. For example, it improves the returned issues of union and variant and allows us to skip actions that require the data to be typed. For example, we cannot validate the length of an array if we are not sure that the input is actually an array. In the future, there will be actions that can validate untyped data if some fields you specify are typed. This will be necessary for complex form validation.

devcaeg:
@znycheporuk

Thank you for your response. I have a new question. I am developing a library that allows developers to use `Valibot`, `Zod`, or `TypeBox`. The developer provides their schemas, and my library analyzes them (based on the properties) to determine which of the three libraries is being used. Once the library is identified, if it is `Zod`, it executes `safeParse`, and if it is `Valibot`, it executes `_parse`. With the recent changes in `Valibot`, it should now execute `_run`. However, how do I know whether to pass `true` or `false` to `typed`?

fabian-hiller:
> However, how do I know whether to pass `true` or `false` to `typed`?

If the data you pass to `._run` is not type safe, you pass `false`. On validation, `._run` will change it to `true` if the data matches the type definition of the schema.

> the second parameter is an object that has all its properties optional, but it is not initialized by default

Yes, that's correct. It's not optional because it's always present in our internal implementation. Changing it to optional would slightly increase the bundle size and error proneness, because TypeScript would no longer alert us if we forgot it.

devcaeg:
Thank you very much for the information.

EltonLobo07:
@fabian-hiller I would like to rewrite the remaining actions - `notLength`, `notValue` and `notSize`. With those actions taken, I think there are no more remaining actions. If I am allowed to rewrite the remaining actions mentioned above, expect pull requests no later than Wednesday of the next week.

AndreyYolkin:
> @fabian-hiller I would like to rewrite the remaining actions - `notLength`, `notValue` and `notSize`. With those actions taken, I think there are no more remaining actions. If I am allowed to rewrite the remaining actions mentioned above, expect pull requests no later than Wednesday of the next week.

I had a look into implementation, it's almost identical to the length, value and size you greatly wrote before. I think, I can submit PR today and you can also make a review, so together we make it faster

fabian-hiller:
I will try to finish the rewrite this weekend and release a pre-release next week. The official release will follow as soon as our documentation is updated and I have looked into a codemode for migration. I expect to release the pre-release on Monday or Tuesday and the final release sometime in June.

fabian-hiller:
All schemas, methods and actions are now implemented. 🎉 Thank you all for your help in porting the actions! Together we ported more than 60 actions! 🙏 Next I will add the missing async implementation of some functions. Once that's done, I'll start preparing our first prerelease. 🚀

fabian-hiller:
Done 🥹

fabian-hiller:
v0.31.0-rc.0 is now available 🙌

saturnonearth:
What is the equivalent way of writing with "merge" with the new pipes? 

Having trouble with converting some thing like this:

```ts
const requestSchema = v.merge([
	v.pick(testSchema, ['name', 'id', 'description']),
	v.object({
		id: v.optional(testSchema.entries.id)
	})
]);
```

fabian-hiller:
`merge` spreaded the `.entries` into a new `object` schema. This is no longer possible because there are now 4 different object schemas. So the new way to achieve the same is to do the spreading yourself. I welcome feedback on this.

```ts
const MergedSchema = v.object({ ...Schema1.entries, ...Schema2.entries });
```

PR #513, opened 2024-04-10
feat(rewrite-with-pipe): add `upperCase` and `lowerCase`

xcfox:


xcfox:
@fabian-hiller Next I'm going to work on [toMaxValue](https://github.com/fabian-hiller/valibot/tree/main/library/src/transformations/toMaxValue) and [toMinValue](https://github.com/fabian-hiller/valibot/tree/main/library/src/transformations/toMinValue). 

I see that the old `toTrim` is named `trim` in this refactoring, so I am using `upperCase` and `lowerCase`.
The validation actions `maxValue` and `minValue` already exist in the code now, should I continue to use `toMaxValue` and `toMinValue` as the new transformation action names?

library/src/actions/lowerCase/lowerCase.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Creates a lower case transformation action.
+ *
+ * @returns A lower case action.
+ */
+export function lowerCase(): LowerCaseAction {
+  return {
+    kind: 'transformation',
+    type: 'lower_case',
+    async: false,
```
Demivan, line 23: @fabian-hiller Do transforms need own type? Do transformations need introspection support?

If it is not needed, all transforms can be converted to simple wrappers:

```
export const lowerCase = transform(value => value.toLowerCase())
```

This will reduce library bundle size even more.

fabian-hiller:
@xcfox my idea was to follow the native JS naming when possible. That's the only reason why I changed `toTrimmed` to `trim`. For `toLowerCase` and the other functions I would stick with the current names.

library/src/actions/lowerCase/lowerCase.ts
fabian-hiller: I don't really know. For now, I would stick with the current approach and keep introspection support. Due to GZIP compression, I do not expect big difference in bundle size.

Demivan: Makes sense. Another question: is LowerCaseAction needed? Looks like it does not provide any benefit over using `BaseTransformation<string, string, never>`

fabian-hiller: > Another question: is LowerCaseAction needed? Looks like it does not provide any benefit over using `BaseTransformation<string, string, never>`

I don't quite understand what you mean. Can you explain it in more detail?

Demivan: I mean, do transformation actions need a separate interface declared for them? 
`LowerCaseAction` is declared on line 6 of this file.
It adds a little bit more typing than just using `BaseTransformation<string, string, never>` directly, but I don't think it provides any useful info.

fabian-hiller: Now I understand. Thanks for the explanation. I see your point, but for now I would continue this way so that the implementation is uniform to validation actions and to schemas.

fabian-hiller:
Thank you! I will review and merge this PR in the next 3 days.

PR #514, opened 2024-04-11
feat(rewrite-with-pipe): add min-bytes validation

Afsoon:


library/src/actions/minBytes/minBytes.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    test('for UTF-8 characters', () => {
+      const value3 = 'あい'; // in UTF-8, 'あい' is 6 bytes
+      expectNoActionIssue(action, [value3]);
+    });
+  });
+
+  describe('should return dataset with issues', () => {
+    const action = minBytes(5, 'message');
+    const getBytesLength = (value: string) => new TextEncoder().encode(value).length;
```
Afsoon, line 67: I avoid exporting these internal implementation functions to use in tests; for me, it's a leak that final users don't need to know, but I'm up to changing it.

fabian-hiller: Wouldn't it be better to enter the number we expect directly? The calculation is not really necessary, is it?

Afsoon: You are right, I have justd changed for a map.

PR #518, opened 2024-04-12
feat(rewrite-with-pipe): add integer, safeInteger and finite

IlyaSemenov:
Add `integer`, `safeInteger` and `finite` for #502.

Please review the untyped inputs test/implementation, I copy/pasted this from `url` action without really understanding the purpose of it.


library/src/actions/finite/finite.test-d.ts
```diff
@@ -24,19 +24,18 @@ describe('finite', () => {
   });
 
   describe('should infer correct types', () => {
-    type Input = 1 | 2 | 3;
-    type Action = FiniteAction<Input, undefined>;
+    type Action = FiniteAction<number, undefined>;
```
IlyaSemenov, line 27: How so? This loses the point of the test — it's now not testing that the correct type (possibly a subtype of number) is inferred.

fabian-hiller: Sorry for the change. I am in a bit of a rush to finish the rewrite. I am not sure anymore why I changed it. Maybe you are right. Feel free to port more actions and later we will have another look at it and maybe change it back to your version.

PR #524, opened 2024-04-15
feat(rewrite-with-pipe): add `bic` action

mxdvl:
To help with #502

### DIFF
--- a/library/src/actions/bic/bic.ts
+++ b/library/src/actions/bic/bic.ts
@@ -1,0 +1,106 @@
+import type {
+  BaseIssue,
+  BaseValidation,
+  ErrorMessage,
+} from '../../types/index.ts';
+import { _validationDataset } from '../../utils/index.ts';
+
+/**
+ * BIC issue type.
+ */
+export interface BicIssue<TInput extends string> extends BaseIssue<TInput> {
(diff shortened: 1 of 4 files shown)


library/src/actions/bic/bic.ts
```diff
@@ -0,0 +1,106 @@
...
+  /**
+   * The error message.
+   */
+  readonly message: TMessage;
+}
+
+/**
+ * [BIC] (https://en.wikipedia.org/wiki/ISO_9362) regex.
+ */
+export const BIC_REGEX = /^[A-Z]{6}(?!00)[A-Z\d]{2}(?:[A-Z\d]{3})?$/u;
```
mxdvl, line 62: Not sure if this should be imported from `../../regex.ts` or defined here… I can see both patterns in the rewrite @fabian-hiller

mxdvl: [Import it is](https://github.com/fabian-hiller/valibot/pull/524/commits/894506ecdf747d4089f13d94c31df5d84f9afcc9#diff-6691fd7eac3e8dc67ce122eb420b34b623c6744d4b976633c4c9e4f04d857ca3R1)

PR #526, opened 2024-04-15
feat(rewrite-with-pipe): add `isoTime` action

mxdvl:
Help with #502 

May be a good starting point for https://github.com/fabian-hiller/valibot/pull/496#issuecomment-2057146443

PR #530, opened 2024-04-16
feat(rewrite-with-pipe): add `isoTimestamp` action

### LINKED ISSUE #495: `isoTimestamp` only accepts UTC timestamps

macarie:
> [!NOTE]  
> Part of the effort started in #502.
>
> This PR is based on the code in #526.

---

As already stated in #496:

> The regex is relatively permissive as it allows for any number in the offset. If something stricter is required, I can work on it, but I'm afraid it would explode in complexity. For example, a `00:00` offset is valid only if preceded by a `+` and not by `-`; the maximum offset should be ±12:00, etc.

Closes #495.


macarie:
I split this into two commits as I'm not sure what will happen with #496.

If that PR ends up being merged, I can rebase and drop the regex change in this one 👍🏼

fabian-hiller:
`yyyy-mm-ddThh:mm:ss.sss±hh` does not seem to be supported by the [Date constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date), even though it is part of the spec. So I am not sure if we should support it. `new Date('2024-04-23T16:25:56.792+00')` results in `Invalid Date`.

macarie:
> `yyyy-mm-ddThh:mm:ss.sss±hh` does not seem to be supported by the [Date constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date), even though it is part of the spec.

To be honest, I've never seen that one used. I just wanted to follow the spec to avoid potential issues in the future.

I see the tests are failing now; I think something was changed upstream. I'll look into that.

fabian-hiller:
I also discovered that `±23:59` is the maximum value. `new Date('2024-04-23T16:25:56.792+99:99')` results in `Invalid Date`.

macarie:
Yes, I mentioned that in the first message.

> The regex is relatively permissive as it allows for any number in the offset. If something stricter is required, I can work on it, but I'm afraid it would explode in complexity. For example, a 00:00 offset is valid only if preceded by a + and not by -; the maximum offset should be ±12:00, etc.

I think the maximum value should be `±12:00`, tho. I'm not sure how the parsing works with `new Date`, but `+13:00` becomes `-01:00`.

macarie:
Actually, it looks like the standard does not specify a minimum/maximum offset at all; it only specifies _how_ it should be formatted. The limits are left to the implementation.

@fabian-hiller, if you want something stricter, I can work on it. It's just that the standard seems to be really open about most stuff. I'm not sure that mirroring what `new Date()` does would be a good choice. Users might use custom parsers that are more/less permissive than `new Date()`, and I think this validation should work for them too.

fabian-hiller:
> I see the tests are failing now; I think something was changed upstream. I'll look into that.

Please don't make any changes to the code. I am still working on it. You can improve the regex later if you have time.

fabian-hiller:
> I'm not sure that mirroring what `new Date()` does would be a good choice. Users might use custom parsers that are more/less permissive than `new Date()`, and I think this validation should work for them too.

I am just worried that people will put it in `new Date(...)` and complain about `Invalid Date`, but I understand your point. I will make some changes to the regex. Feel free to review them and improve it further in another PR.

fabian-hiller:
I will merge it already, but feel free to check my changes anyway.

Issue #81, opened 2023-08-08
feature: add `formData` schema


Yovach:
Hi,
I don't know how this would affect bundle size, but I think it would be interesting to have an API like :

```typescript
const reportSchema = formData({
  title: string(),
  content: string(),
  photo: file(),
});
```

It could be useful in `server actions` (Next.js) and client-side validations in `<form>`.

fabian-hiller:
I think the idea is cool, but I don't know yet how much sense it makes, because as far as I know `FormData` can't be typed. Probably the better approach would be to use `object` and convert `FormData` to an object before validation.

fabian-hiller:
Here is a code example:

```ts
import { email, minLength, object, parse, string } from "valibot"; // 0.75 kB
import { loginUser } from "~/api";

const LoginSchema = object({
  email: string([
    minLength(1, "Please enter your email."),
    email("The email address is badly formatted."),
  ]),
  password: string([
    minLength(1, "Please enter your password."),
    minLength(8, "You password must have 8 characters or more."),
  ]),
});

export default function LoginRoute() {
  async function login(formData: FormData) {
    "use server";
    try {
      const { email, password } = parse(
        LoginSchema,
        Object.fromEntries(formData.entries())
      );
      await loginUser({ email, password });
    } catch (error) {
      // Handle errors
    }
  }

  return (
    <form action={login}>
      <input name="email" type="email" required />
      <input name="password" type="password" required minLength={8} />
      <button type="submit">Login</button>
    </form>
  );
}
```

Yovach:
> Here is a code example:
> 
> ```ts
> import { email, minLength, object, parse, string } from "valibot"; // 0.75 kB
> import { loginUser } from "~/api";
> 
> const LoginSchema = object({
>   email: string([
>     minLength(1, "Please enter your email."),
>     email("The email address is badly formatted."),
>   ]),
>   password: string([
>     minLength(1, "Please enter your password."),
>     minLength(8, "You password must have 8 characters or more."),
>   ]),
> });
> 
> export default function LoginRoute() {
>   async function login(formData: FormData) {
>     "use server";
>     try {
>       const { email, password } = parse(
>         LoginSchema,
>         Object.fromEntries(formData.entries())
>       );
>       await loginUser({ email, password });
>     } catch (error) {
>       // Handle errors
>     }
>   }
> 
>   return (
>     <form action={login}>
>       <input name="email" type="email" required />
[...]

PR #539, opened 2024-04-19
feat: add formData schema

frenzzy:
Fixes https://github.com/fabian-hiller/valibot/issues/81, implements the suggestion from https://github.com/fabian-hiller/decode-formdata/issues/10.

This PR introduces a new formData schema that enables the extraction and validation of required FormData directly from the server. Here’s how it works:

For example, if you send the following FormData to the server:
```html
<form enctype="multipart/form-data" method="post">
  <input name="title" type="text" value="Red apple" />
  <input name="price" type="number" value="0.89" />
  <input name="created" type="date" value="2024-04-12T07:00:00.000Z" />
  <input name="active" type="checkbox" value="true" />
  <input name="tags" type="text" value="fruit" />
  <input name="tags" type="text" value="healthy" />
  <input name="tags" type="text" value="sweet" />
  <input name="images.0.title" type="text" value="Close up of an apple" />
  <input name="images.0.created" type="date" value="2024-04-12T07:01:00.000Z" />
  <input name="images.0.file" type="file" value="a.jpg" />
  <input name="images.0.tags" type="text" value="foo" />
  <input name="images.0.tags" type="text" value="bar" />
  <input name="images.1.title" type="text" value="Our fruit fields at Lake Constance" />
  <input name="images.1.created" type="date" value="2024-04-12T07:02:00.000Z" />
  <input name="images.1.file" type="file" value="b.jpg" />
  <input name="images.1.tags" type="text" value="baz" />
</form>
```
You can now define a validation schema like this:
```ts
import * as v from 'valibot';

const MySchema = v.formData({
  title: v.string(),
  price: v.number(),
  created: v.date(),
  active: v.boolean(),
  tags: v.array(v.string()),
  images: v.array(
    v.object({
      title: v.string(),
      created: v.date(),
      file: v.blob(),
      tags: v.array(v.string())
    })
  ),
});
```
You can use this schema to extract and validate the required info:
```ts
import { parse } from 'valibot'

async function server(formData: FormData) {
  const data = parse(MySchema, formData)
}
```
This will give you validated and typed form data like this:
```ts
const data = {
  title: 'Red apple',
  price: 0.89,
  created: Date,
  active: true,
  tags: ['fruit', 'healthy', 'sweet'],
  images: [
    {
      title: 'Close up of an apple',
      created: Date,
      file: File,
      tags: ['foo', 'bar'],
    },
    {
      title: 'Our fruit fields at Lake Constance',
      created: Date,
      file: File,
      tags: ['baz'],
    },
  ],
}
```

This PR streamlines the way we handle and validate `FormData` from the client, ensuring the data is correctly typed and validated upon reception.

### DIFF
--- a/library/src/schemas/formData/formData.test.ts
+++ b/library/src/schemas/formData/formData.test.ts
@@ -1,0 +1,483 @@
+import { describe, expect, test } from 'vitest';
+import { type ValiError } from '../../error/index.ts';
+import { parse } from '../../methods/index.ts';
+import type {
+  Output,
+  TypedSchemaResult,
+  UntypedSchemaResult,
+} from '../../types/index.ts';
+import { custom, minLength } from '../../validations/index.ts';
+import { array } from '../array/array.ts';
+import { boolean } from '../boolean/boolean.ts';
+import { date } from '../date/date.ts';
+import { instance } from '../instance/instance.ts';
+import { nullable } from '../nullable/nullable.ts';
+import { number } from '../number/index.ts';
+import { object } from '../object/object.ts';
+import { optional } from '../optional/optional.ts';
+import { string } from '../string/index.ts';
+import { formData } from './formData.ts';
+
+describe('formData', () => {
+  test('should pass only FormData', () => {
+    const schema = formData({ foo: string() });
+    const input = new FormData();
+    input.append('foo', 'bar');
+    const output = parse(schema, input);
+    expect(output).toEqual({ foo: 'bar' });
+
+    expect(() => parse(schema, {})).toThrowError();
+    expect(() => parse(schema, 123)).toThrowError();
+  });
+
+  test('should throw custom error', () => {
+    const error = 'Value is not a FormData!';
+    const schema = formData({ n: number() }, error);
+    expect(() => parse(schema, 123)).toThrowError(error);
+  });
+
+  test('should throw every issue', () => {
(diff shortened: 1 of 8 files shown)


fabian-hiller:
Thanks for the PR and your contribution to the project. At the moment I am not sure if we should add `formData` as a schema to Valibot. The reason is that schemas do not transform or change the input. `date()` returns a `Date`, `blob()` returns a `Blob` and people probably expect `formData(...)` to return a `FormData` instance.

On the other hand, I see value in adding functions like `filterArray` and `filterRecord` in the long run, and `decodeFormData` might fit in perfectly in this category. But right now I have no time to think about it in more detail. Another idea could be to implement this function based on `decode` of my `decode-formdata` package and publish it via `@decode-formdata/valibot` on npm. Feel free to share your thoughts.

fabian-hiller:
I think it might be the right approach to implement `decodeFormData` as a schema function. Similar to the idea of `filterArray` and `filterRecord`. But before we do that, we have to think about the mental model for Valibot as well as the naming. This process will take a while due to the ongoing rewrite of the library.

Should we add these advanced schemas to the core package? Should they be added to `/schemas` or do we confuse people with such advanced functions next to schemas that "just" validate data types? These are just some of the questions I have in mind.

frenzzy:
Closing in favor of https://github.com/fabian-hiller/valibot/pull/672

PR #568, opened 2024-05-07
feat(rewrite-with-pipe): add `isoWeek` action

mxdvl:
Part of #502

Note that some years, [like 2012, can have 54 weeks](https://www.timeanddate.com/calendar/custom.html?year=2012&country=9&cdt=1&typ=0&display=3&df=1).

### DIFF
--- a/library/src/actions/index.ts
+++ b/library/src/actions/index.ts
@@ -23,6 +23,7 @@ export * from './isoDate/index.ts';
 export * from './isoDateTime/index.ts';
 export * from './isoTime/index.ts';
 export * from './isoTimestamp/index.ts';
+export * from './isoWeek/index.ts';
 export * from './length/index.ts';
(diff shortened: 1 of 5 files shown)


fabian-hiller:
> Note that some years, like 2012, can have 54 weeks.

Are you sure about that? According to [Wikipedia](https://en.wikipedia.org/wiki/ISO_week_date#Last_week), 53 is the maximum number for leap years.

mxdvl:
> > Note that some years, like 2012, can have 54 weeks.
> 
> Are you sure about that? According to [Wikipedia](https://en.wikipedia.org/wiki/ISO_week_date#Last_week), 53 is the maximum number for leap years.

Ah of course ISO thought about that! I was simply going by the fact that 2012 had 54 different weeks in it, but in that case 1 January is W52 & 31 December is W01.

PR #570, opened 2024-05-08
docs(installation): add jsr

Shyam-Chen:
> [Migrating Deno modules from /x to JSR](https://jsr.io/docs/migrate-x-to-jsr)

--- 

Updated:

> **There are no plans to discontinue or shut down deno.land/x.** Module authors can continue to publish and update modules there, and Deno code using it will continue to function.


fabian-hiller:
Thanks for creating this PR and expanding the guide for JSR. I think instead of removing Deno, we should keep it until they deprecate it in favor of JSR, since Valibot is still a very popular package on /x.

fabian-hiller:
Thank you!

PR #578, opened 2024-05-15
feat(rewrite-with-pipe): add `regex` action

EltonLobo07:


fabian-hiller:
I am sorry, but I may simplify the unit tests. I don't think it's necessary to test the native regex functionality since it's not controlled by our source code. I prefer to just test that the passed regex is used to validate the input.

PR #580, opened 2024-05-18
feat(rewrite-with-pipe): `notLength`, `notValue` and `notSize`

AndreyYolkin:
This PR adds last missing actions to #502


AndreyYolkin:
I'll mark PR as ready, but consider review the tests for `notValue`, as some type conversions are not trivial and probably we need a better coverage for them

fabian-hiller:
That's fine. I will review all the code before merging anyway. Thank you very much!

PR #585, opened 2024-05-21
docs: fixed trim links

Balastrong:
Just noticed these links were not working :)


fabian-hiller:
I am in the process of updating the documentation for v0.31.0 and some names like these have changed. The API reference is outdated, not the guides. 🙃

PR #591, opened 2024-05-23
Add Grit migration script

morgante:
You can try this out by running

```
grit apply valibot_v03
```

Grit also includes built-in testing for the markdown file. Just run `grit patterns test` in this repo.

### DIFF
--- a/.grit/patterns/valibot_v03.md
+++ b/.grit/patterns/valibot_v03.md
@@ -33,7 +33,8 @@ pattern vb_schemas_with_mandatory_arg() {
     `union`,
     `intersect`,
     `tuple`,
-    `picklist`
+    `picklist`,
+    `variant`
   }
 }
 
@@ -50,8 +51,11 @@ pattern vb_pipe_schema_args($schema, $new_schema_args, $piped_args) {
         // If it is a union, then there must be a second argument
         and {
           $schema <: vb_schemas_with_mandatory_arg(),
-          $one_arg <: not undefined
-        }
+          $one_arg <: not undefined,
+          if ($schema <: `variant`) {
+            $two_arg <: not undefined
+          }
+        },
       }
     },
     $new_schema_args = join($other_args, `,`)
@@ -138,14 +142,15 @@ const Schema = v.pipe(v.map(v.number(),v.pipe(v.string(), v.url(), v.endsWith('@
 
 ## Union without pipe
 
-`intersect` and `union` are not necessarily piped.
+`intersect` and `union` are not necessarily piped. `variant` also takes a non-piped array as the second argument.
 
 ```js
 const Schema = v.object({
   email: v.string([v.email(), v.endsWith('@gmail.com')]),
   password: v.string([v.minLength(8)]),
   other: v.union([v.string([v.decimal()]), v.number()]),
-  intersection: v.intersect([v.string(), v.number()])
+  intersection: v.intersect([v.string(), v.number()]),
+  variant: v.variant("type", [v.string(), v.number()])
 });
 ```
 
@@ -154,7 +159,8 @@ const Schema = v.object({
   email: v.pipe(v.string(), v.email(), v.endsWith('@gmail.com')),
   password: v.pipe(v.string(), v.minLength(8)),
   other: v.union([v.pipe(v.string(), v.decimal()), v.number()]),
-  intersection: v.intersect([v.string(), v.number()])
+  intersection: v.intersect([v.string(), v.number()]),
+  variant: v.variant("type", [v.string(), v.number()])
 });
 ```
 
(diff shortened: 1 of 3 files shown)


.grit/patterns/valibot_v03.md
```diff
@@ -0,0 +1,105 @@
...
+    `toCustom` => `transform`,
+    `toTrimmed` => `trim`,
+    `toTrimmedEnd` => `trimEnd`,
+    `toTrimmedStart` => `trimStart`
+  }
+}
+
+pattern vb_pipe() {
+  or {
+    `v.string([$args])` => `v.pipe(v.string(), $args)`,
```
fabian-hiller, line 24: This should work for any schema function, not just `string`. It is also important that it works with any import alias as well as with direct imports (same applies to the other steps). You can find a list of all schema functions in our [API reference](https://valibot.dev/api/).

.grit/patterns/valibot_v03.md
```diff
@@ -0,0 +1,105 @@
...
+  }
+}
+
+pattern vb_coerce() {
```
fabian-hiller, line 28: I would change it to something like this:

```
pattern vb_coerce() {
  or {
    `v.coerce($schema, $fn)` => `v.pipe(v.unknown(), v.transform($fn))`
  }
}
```

fabian-hiller (changes requested):
Thank you very much! This looks really nice if it works as expected!

fabian-hiller:
You can always copy the output to our [playground](https://valibot.dev/playground/) to see if the transformed code is valid.

.grit/patterns/valibot_v03.md
morgante: Sounds good!

.grit/patterns/valibot_v03.md
morgante: Sorry I don't really understand the distinction in your first argument - the API is not intuitive enough to me to understand what you're going for / why `v.string([v.email()])` becomes `v.pipe(v.string(), v.email()` but `v.map(v.number(), v.string())` becomes `v.pipe(
  v.map(
    v.number(),
    v.pipe(v.string(), v.url(), v.endsWith('@example.com')), // <-- nested pipe
  )`.
  
I was happy to provide a sample, but don't understand your API intimately.

morgante: I also think you need to calibrate your expectations. Writing a codemod that handles 100% of all edge cases is usually not worth it - people still need to review the code. The main value is in handling the 90% common cases.

fabian-hiller: Thank you! I will try to have a look later today or in the next few days.

fabian-hiller:
If I run `grit apply valibot_v03` I get the following error: `Error: Permission denied (os error 13)`

fabian-hiller:
Once I can confirm that the codemod works, I will add instructions on how to run it to our migration guide. Do people need to manually add the `.grit/patterns/valibot_v03.md` file to their repo to run the command? Does it only works on Linux and MacOS?

morgante:
> If I run `grit apply valibot_v03` I get the following error: `Error: Permission denied (os error 13)`

Hmm, what steps did you take to install it / can you provide reproduction instructions? This isn't expected.

> Do people need to manually add the .grit/patterns/valibot_v03.md file to their repo to run the command?

Yes, currently they will need to download it. I'm planning to add a [feature](https://github.com/getgrit/gritql/issues/356) for applying remote patterns directly too.

> Does it only works on Linux and MacOS?

We also have Windows [releases](https://github.com/getgrit/gritql/releases) available.

fabian-hiller:
> Hmm, what steps did you take to install it / can you provide reproduction instructions? This isn't expected.

I followed the [Quickstart](https://docs.grit.io/cli/quickstart#installation) guide and executed: `npm install --location=global @getgrit/launcher`. I can run `grit`. So the installation was successful.

I got it to work with `sudo` but it should work without it. I will test the codemod in the next few days and get back to you. Thanks again!

> Yes, currently they will need to download it. I'm planning to add a https://github.com/getgrit/gritql/issues/356 for applying remote patterns directly too.
> We also have Windows [releases](https://github.com/getgrit/gritql/releases) available.

👍

morgante:
> I got it to work with sudo but it should work without it.

Agreed, sudo definitely shouldn't be required. Maybe you can try running `grit doctor` without sudo to figure out why. I'm also happy to help debug on Discord.

fabian-hiller:
I get the same error when I type `grit doctor`. With `sudo` it prints out environment and configuration info. I will get back to you on Discord tonight or in the next few days. In the meantime, if you have time, you can try to get the following codemod to work.

> Hint: `transform` and `brand` can be deeply nested and should now be placed next to each other in the pipeline of the wrapped schema.

```ts
// Before
const Schema = v.transform(
  v.brand(v.string(), "Name"),
  (input) => input.length
);

// After
const Schema = v.pipe(
  v.string(),
  v.brand('Name'),
  v.transform((input) => input.length)
);
```

morgante:
This now handles `brand` and `transform`.

PR #612, opened 2024-05-28
fix: exclude unrelated names from codemod

morgante:
We need to make sure the codemod doesn't rename methods that don't come from valibot.


fabian-hiller:
Thank you!

Issue #621, opened 2024-06-03
Automatic upgrade to v0.31.0 with Grit only works for certain patterns


morinokami:
Hi, I tried the automatic upgrade command for v0.31.0 as described [here](https://valibot.dev/guides/migrate-to-v0.31.0/), but it seems to only work well with specific patterns when using Grit. When I run the command `npx @getgrit/cli apply github.com/fabian-hiller/valibot#migrate_to_v0_31_0` on the following code, only the first schema is converted, and the rest are not.

```js
import * as v from 'valibot';

const Schema = v.pipe(v.string(), v.email());

const ObjectSchema = v.object({ key: v.string() }, v.null_());
const TupleSchema = v.tuple([v.string()], v.null_());

const LooseObjectSchema = v.object({ key: v.string() }, v.unknown());
const LooseTupleSchema = v.tuple([v.string()], v.unknown());
const StrictObjectSchema = v.object({ key: v.string() }, v.never());
const StrictTupleSchema = v.tuple([v.string()], v.never());
```

The command output is as follows:

```sh
$ npx @getgrit/cli apply github.com/fabian-hiller/valibot#migrate_to_v0_31_0
./index.js
     import * as v from 'valibot';
     
    -const Schema = v.string([v.email()]);
    +const Schema = v.pipe(v.string(), v.email());
     
     const ObjectSchema = v.object({ key: v.string() }, v.null_());
     const TupleSchema = v.tuple([v.string()], v.null_());
```

fabian-hiller:
Thank you for creating this issue. Can you send me the before, after and expected code? Here is a template:

```ts
// Before
...

// After
...

// Expected
...
```

morinokami:
@fabian-hiller Sure! The schemas in my code are excerpts from [the migration document](https://valibot.dev/guides/migrate-to-v0.31.0/), so the expected code is the same as described there. Here is the filled-in template:

```js
// Before
import * as v from 'valibot';

const Schema = v.string([v.email()]);

const ObjectSchema = v.object({ key: v.string() }, v.null_());
const TupleSchema = v.tuple([v.string()], v.null_());

const LooseObjectSchema = v.object({ key: v.string() }, v.unknown());
const LooseTupleSchema = v.tuple([v.string()], v.unknown());
const StrictObjectSchema = v.object({ key: v.string() }, v.never());
const StrictTupleSchema = v.tuple([v.string()], v.never());

// After
import * as v from 'valibot';

[...]

PR #625, opened 2024-06-04
fix: migrate object and tuples with grit

seren5240:
fixes https://github.com/fabian-hiller/valibot/issues/621

### DIFF
--- a/.grit/patterns/migrate_to_v0_31_0.md
+++ b/.grit/patterns/migrate_to_v0_31_0.md
@@ -143,6 +143,16 @@ pattern rewrite_object_and_tuple($v) {
   }
 }
 
+pattern rewrite_merge($v) {
+  `$v.merge([$schemas])` where {
+    $entries = [],
+    $schemas <: some bubble($entries) $schema where {
+      $entries += `...$schema.entries`,
+    },
+    $entries = join($entries, `, `),
+  } => `$v.object({ $entries })`
+}
+
 pattern is_valibot() {
   `$v` where {
     $v <: or {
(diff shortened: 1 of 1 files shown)


fabian-hiller:
Thank you very much! I will merge it, but I think there are a few cases your changes do not cover yet. `object` can take additional arguments besides the entries and the rest schema. The same goes for `merge`. We just need to copy those arguments over. Also, I suggest applying the `merge` pattern before the `object` pattern, as the `object` pattern may transform the output of the `merge` pattern if `merge` contains a rest schema. Also, I would run both patterns before anything else, as the `rewrite_pipes` pattern may transform the output of the other patterns.

```ts
// Before
const Schema = merge([ObjectSchema1, ObjectSchema1], v.string(), 'Error message', [v.custom(() => true)]);

// After
const Sschema = v.pipe(v.objectWithRest({ ...ObjectSchema1.entries, ...ObjectSchema1.entries }, v.string(), 'Error message'), v.check(() => true));
```

seren5240:
@fabian-hiller I believe the instance where `object` or `rewrite_pipes` transforms the output of other rewrites is handled by the top-level `sequential` pattern, which runs `rewrite_functions` twice. The order of the disjunction in `rewrite_functions` doesn't actually make a difference to the final output.

I'm having trouble finding documentation of the old `merge` function. Would you be able to point me to it, or else explain how the various arguments of

```
const Schema = merge([ObjectSchema1, ObjectSchema1], v.string(), 'Error message', [v.custom(() => true)]);
```

map to the new `pipe` syntax?

fabian-hiller:
`merge` returned an `object` and had the exact same arguments with exception of the first argument to merge the object entries. [Here](https://github.com/fabian-hiller/valibot/blob/v0.30.0/website/src/routes/api/(methods)/merge/index.mdx) is the old API reference. `pipe` is necessary in this example because of last argument. The merge pattern should only transform it to `object`. The object pattern should pick it up if any "rest" is specified and the pipe pattern should transform the pipelines.

PR #626, opened 2024-06-04
fix: transform schemas with additional arguments

seren5240:
Per discussion: https://github.com/fabian-hiller/valibot/pull/625#issuecomment-2147168690

### DIFF
--- a/.grit/patterns/migrate_to_v0_31_0.md
+++ b/.grit/patterns/migrate_to_v0_31_0.md
@@ -154,11 +154,23 @@ pattern rewrite_nested_pipes($v, $args) {
 
 pattern rewrite_object_and_tuple($v) {
   or {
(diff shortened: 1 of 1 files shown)


.grit/patterns/migrate_to_v0_31_0.md
```diff
@@ -37,7 +37,7 @@ pattern rewritten_names() {
     `toCustom` => `transform`,
     `toTrimmed` => `trim`,
     `toTrimmedEnd` => `trimEnd`,
-    `toTrimmedStart` => `trimStart`
+    `toTrimmedStart` => `trimStart`,
```
seren5240, line 40: Might be good to run `grit patterns test` before merges, since this missing comma breaks the pattern from running at all.

PR #637, opened 2024-06-08
fix(deno): Add missing file extensions

Omochice:
## Summary

Fixled some files that were imported without extensions.

## Details

When using `https://deno.land/x/valibot` with deno, the type check fails after updating to `v0.31.0`.

So, I fixed some files has `import` without extensions.

I checked this fixes by `deno check src/**/*.ts` on `Library` dir.


fabian-hiller:
Sorry! And thank you!

Issue #499, opened 2024-03-30
Search fails to work


scarf005:
![image](https://github.com/fabian-hiller/valibot/assets/54838975/67b98040-e1d4-47ab-9a1f-01586d7fc88d)

## To reproduce

1. visit https://valibot.dev
2. search for `_types`
3. search fails to work.

## Additional context

![image](https://github.com/fabian-hiller/valibot/assets/54838975/d534e5f7-2ebe-4882-a294-c609008e9dfd)

there was no errors on console or network tab.

fabian-hiller:
Thank you for reporting this issue. I will look into it next month.

PR #651, opened 2024-06-13
Fix doc search error when the first char is a special char

TristanShz:
Hi, I fixed the problem related to issue #499, let me know if you think there is a better way to do it.


fabian-hiller:
Thank you so much! I will review it tomorrow.

fabian-hiller:
Thank you for this PR. It helped me figure out what was wrong. Your code fixed but did not solve the problem. The main problem was that Algolia was returning search results that did not make sense when special characters were entered, and that we were not resetting the error state when the input changed. I solved both problems.

PR #663, opened 2024-06-19
Use type exports where appropriate

allezxandre:
Some type exports are erased by TypeScript and break when using Deno as the runtime:

```
error: Uncaught SyntaxError: The requested module './types.ts' does not provide an export named 'PartialCheckIssue'
export { PartialCheckIssue } from './types.ts';
         ^
    at <anonymous> (https://jsr.io/@valibot/valibot/0.33.1/src/actions/partialCheck/index.ts:3:10)
```

This PR fixes those exports


fabian-hiller:
Congratulations on becoming our 100th contributor! 🍾

PR #666, opened 2024-06-20
Improve regexes

slevithan:
I didn't audit all the regexes. Just the specific ones described below.

EMOJI_REGEX

- Previous version matched lots of false positives like digits (`0`, etc.), `*`, bare U+200D (ZWJ), and some symbols like `👁`, `✈`, `🏳`, and `♂` even when they're not followed by U+FE0F (none of which should match). So I fixed it.
- See some examples I put together at https://regex101.com/r/WKFYNE/1
- One open question I have is whether all enclosing marks are allowed where I used `\p{Me}`, or if it's a more limited set like just U+20E3. The segment in question (`\p{Emoji}\uFE0F\p{Me}?`) is used to match emoji like `2️⃣` which is made up of 3 code points: U+32 U+FE0F U+20E3.

HEXADECIMAL_REGEX

- Replaced `0h|0x` with `0[hx]`.

IPV4_REGEX

- Replaced each byte segment `(?:(?:[1-9]|1\d|2[0-4])?\d|25[0-5])` with `(?:2[0-4]\d|25[0-5]|1\d\d|[1-9]?\d)`. IMO it's easier to read without the nested grouping, and it's the same length. Then replaced `\d\d` with `\d{2}` to work around an eslint error that I don't agree with.
- Used interpolation to not repeat the byte subpattern.

IPV6_REGEX

- No changes except replacing the two IPv4 subpatterns with interpolation. There's probably still room for improvement, at least if you used lookahead to help remove some of the redundancy.

IP_REGEX

- 654 characters of a regex literal to 41 chars using a dynamic regex with simple composition.


slevithan:
> > One open question I have is whether all enclosing marks are allowed where...
> 
> What do you recommend?

I recommend using it as written, with `\p{Me}` instead of `\u20E3`. It won't create *problems* and might allow for edge-case emoji I can't think of. @mathiasbynens would likely know the precise answer.

> > Used interpolation to not repeat the byte subpattern.
> 
> I had implemented this the same way before, but some bundlers had trouble tree-shaking it and including the regex in every bundle, whether used or not. So I repeated it. Due to compression it is probably also better in terms of bundle size.

Fair enough, if bundle size trumps readability. Another solution would be to use the [`regex`](https://github.com/slevithan/regex) package since its *subroutines* syntax allows subpattern composition without interpolation. But I'm guessing you won't think the dependency (and the reliance on flag `v` from Node.js 20) is the right tradeoff.

fabian-hiller:
Thanks you for your feedback! It may take some time to finish this PR as other things have a higher priority at the moment.

slevithan:
I've done a lot more research on the emoji regex and published an improved/fixed version as [`emoji-regex-xs`](https://github.com/slevithan/emoji-regex-xs). I'll update the PR here shortly, and also remove all the use of composition in the regexes since it sounds like you'd prefer to avoid that.

fabian-hiller:
Thank you very much! I will probably review and merge this PR next week.

slevithan:
OK, updated. Passes tests and lint.

slevithan:
In the last-added commit, I updated the comment about `RGI_Emoji`, pointing out that it's not a better replacement, even in the future when targeting Node.js 20+. The reasons are discussed in the [`emoji-regex-xs`](https://github.com/slevithan/emoji-regex-xs) readme, but I'll also describe them here.

To clarify what the emoji regex in this PR matches:

- It matches everything matched by ES2024's `/\p{RGI_Emoji}/v`.
- It excludes non-emoji symbols, dingbats, "emoji components" (like digits, *, #, and ZWJ), etc.
  - This is where the emoji regex prior to this PR fails.
- It intentionally matches additional non-RGI emoji.

On that last point, unfortunately, some common-sense and broadly-supported emoji are not officially in the "RGI" list. Even the Unicode org provides [emoji-test.txt](https://unicode.org/Public/emoji/latest/emoji-test.txt) that mixes in non-RGI emoji strings to help identify real-world emoji. And some emoji are commonly used in an underqualified or overqualified way (by including or excluding certain invisible Unicode markers) that prevents them from being matched by `RGI_Emoji`. For example, the iOS emoji keyboard overqualifies certain emoji. So we need something that matches everything in `RGI_Emoji`, and more.

The regex here allows overqualified and underqualified emoji using a general pattern that matches all Unicode sequences that follow the structure of valid emoji.

fabian-hiller:
Thank you for your research! Is the new emoji regex more strict or accurate than the old one?

slevithan:
Yes. If by "the old one" you mean my own iterations, this is the final version based on research in depth, and I've now therefore also published it as its own library ([`emoji-regex-xs`](https://github.com/slevithan/emoji-regex-xs)). If there are changes to `emoji-regex-xs` in the future (e.g. if new versions of Unicode modify the general patterns for emoji), it can easily be updated by following along with future versions of `emoji-regex-xs`.

`emoji-regex-xs` shares the API and 3000+ tests with [`emoji-regex`](https://github.com/mathiasbynens/emoji-regex). `emoji-regex` is very large (13 KB uncompressed), but it is authoritative (its author helped add things like `\p{}`, `RGI_Emoji`, and `/v` to JS regexes), well established, and extremely broadly used. `emoji-regex` now also recommends and crosslinks `emoji-regex-xs` in its readme, for people who want a *general* pattern (not tied to a specific version of Unicode) that can therefore be much more lightweight.

If by "the old one" you mean the regex currently used in the code that this PR replaces, then yes, this PR is both more strict and more accurate.

Here is the current regex that is being replaced by this PR:

```js
/^[\p{Extended_Pictographic}\p{Emoji_Component}]+$/u
```

This is *extremely* wrong and should never have been used. I'm assuming you picked it up from Zod, which uses the same thing, and I can find other people posting it online, which is where Zod probably got it from. It presumably has spread virally because because:

1. It indeed matches any sequence of one or more emoji.
2. Emoji are *complicated* and most people aren't aware of what specific Unicode properties like `\p{Emoji_Component}` match, or the full variation of different Unicode pattern sequences needed to match the full set of emoji, much less to also deal with with real-world data (where some additional variation is added).
3. Until now (with `emoji-regex-xs`, which came about based on my work on this PR 😊), there simply has not been a lightweight JS regex pattern available online that is this robust. So people have settled for incomplete/broken patterns or, if they knew what they were doing, they used the giant `emoji-regex` pattern that hard codes thousands of code points. Or, in the last year, they started using ES2024's `\p{RGI_Emoji}`, but that has problems as well which is why neither `emoji-regex` nor `emoji-regex-xs` follows it strictly.

There are two big problems with the regex that this PR replaces:

1. It matches many things that are not emoji.
2. It cannot be used (or modified) to match *individual* emoji, because its reliance on `\p{Emoji_Component}` means it cannot possibly identify where where one emoji ends and the next starts. It only works at all because Valibot's requirement for this regex is to match *one or more* back-to-back emoji.

Regarding the first problem, I already mentioned some of its false positives in this PR's initial comment, but here are some additional details:

- Incorrectly matches `*` (U+002A) and `#` (U+0023).
- Incorrectly matches ASCII digits (`0`, `1`, `234`, etc.).
- Incorrectly matches a large number of symbols and dingbats that are not emoji (most of which predate emoji). Examples include `✈` (U+2708) and `🏳` (U+1F3F3).
  - Some of these characters also have versions that *are* emoji and *should* be matched. Ex: `✈️` (U+2708 U+FE0F) and `🏳️` (U+1F3F3 U+FE0F).
- Incorrectly matches a variety of Unicode format characters, enclosing marks, nonspacing marks, and other modifiers that can be used to help construct emoji but are not emoji on their own. Examples include [U+200D](https://codepoints.net/U+200D), [U+20E3](https://codepoints.net/U+20E3), [U+FE0F](https://codepoints.net/U+FE0F), and [U+E007F](https://codepoints.net/U+E007F).
- Incorrectly matches all *regional indicator symbol letters* A-Z, *tag digits* 0-9, and *tag Latin small letters* a-z (collectively matched by `/[\u{1F1E6}-\u{1F1FF}\u{E0030}-\u{E0039}\u{E0061}-\u{E007A}]/u`). These can be used as components of emoji flags and [subdivision flags](https://emojipedia.org/emoji-tag-sequence#list), but are not emoji on their own.

fabian-hiller:
Thanks again for your research and detailed answer! I thought it would be the best DX if `emoji` could be combined with `minLength`, `maxLength` and `length` to control the number of emoji. Unfortunately, this does not work because emoji can have different lengths. An alternative would be to control this via the arguments of `emoji` and dynamically bake it into the regex. What do you think is the best solution?

slevithan:
I'm not familiar with Valibot's APIs. Could `minLength`, `maxLength`, and `length` not be sensibly used to refer to the number of discrete emoji to match, rather than the number of code points (returned by `[...str].length`) or code units (`str.length`) in the matched string? If they could, then both of the APIs you described sound reasonable to me, but I might be more opinionated if you showed code examples of how the two approaches would be used.

Certainly, with the new emoji regex, something like this could easily be done. You'd just need to change the `+` quantifier in the `^(?:…)+$` wrapper to whatever you want (e.g., `{1}`, `{5}`, `{2,17}`, or `{1,}`). It cannot be done with the current (pre-PR) emoji regex, for reasons explained in my last comment.

But this seems like something for a follow-up issue. I'd prefer to land this PR as is and for new functionality to be added afterward.

fabian-hiller:
> But this seems like something for a follow-up issue. I'd prefer to land this PR as is and for new functionality to be added afterward.

I agree.

> [...] be sensibly used to refer to the number of discrete emoji to match

Any ideas on how to implement this? Maybe we could add something like a `minChar`, `maxChar` and `char` action.

slevithan:
Okay, I looked at [valibot.dev/api/string/](https://valibot.dev/api/string/) and [valibot.dev/api/emoji/](https://valibot.dev/api/emoji/) to understand a bit more what you're referring to.

I agree that `minLength`, `maxLength`, and `length` must continue to refer to JavaScript's UTF-16 code units.

The term "character" is very overloaded so I'd advise against using `char`, etc. I think the concept you're looking for is "[grapheme clusters](https://unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries)" or "extended grapheme clusters", which Unicode also describes as "user-perceived characters". A concrete example is the emoji '👩🏻‍👩🏻‍👦🏻‍👦🏻':

```js
// Code unit length
'👩🏻‍👩🏻‍👦🏻‍👦🏻'.length;
// 19

// Code point length
[...'👩🏻‍👩🏻‍👦🏻‍👦🏻'].length;
// 11

// Grapheme cluster length
[...new Intl.Segmenter().segment('👩🏻‍👩🏻‍👦🏻‍👦🏻')].length;
// 1
```

[`Intl.Segmenter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter/Segmenter) is in Node.js 16+ ([browser support](https://caniuse.com/mdn-javascript_builtins_intl_segmenter)). The JS library [orling/grapheme-splitter](https://github.com/orling/grapheme-splitter) extends support backward.

Since `Intl.Segmenter`'s documentation uses "grapheme" to refer to grapheme clusters, I think it's okay to use `grapheme` the same way (even though it actually segments clusters of graphemes), especially if `Intl.Segmenter` is used as the implementation.

I think `minGraphemeLength`, `maxGraphemeLength`, and `graphemeLength` would be significantly preferable to controlling this via the arguments of `emoji`, since counting grapheme clusters is broadly useful. E.g. the Spanish letter "ñ" can be represented as either `'\u00F1'` or `'\u006E\u0303'`. So either 1 or 2 code points, but 1 grapheme cluster.

slevithan:
Just realized that my example emoji '👩🏻‍👩🏻‍👦🏻‍👦🏻' currently only renders as a single user-perceived character on Microsoft and Facebook platforms. Since other platforms don't currently include a unique emoji design for it, they might render it as four discrete emoji characters in a row, possibly while still selecting it as a single character.

slevithan:
Edited my last comment to use '👩🏻‍🏫' (a grapheme cluster with two graphemes) instead of '👩🏻‍👩🏻‍👦🏻‍👦🏻' (a grapheme cluster with four graphemes), since the latter currently only [renders](https://emojipedia.org/family-woman-light-skin-tone-woman-light-skin-tone-boy-light-skin-tone-boy-light-skin-tone#designs) as a single user-perceived character on Microsoft and Facebook platforms. The emoji sets from Apple and Google don't include a unique design for it, so at least on iOS it renders as four discrete emoji characters, while the cluster is still *selected* as a single character.

fabian-hiller:
Sorry for the late reply. I took a week off. Thanks for your research. I like the idea of adding a `minGraphemes` and `maxGraphemes` and `graphemes` action using `Intl.Segmenter` to count emojis. We can work on it after merging this PR.

Would you rename `emoji` to `emojis` to reflect that multiple emoji are allowed? Should we update the unit tests of `emoji` to reflect the changes in this PR?

slevithan:
> Would you rename `emoji` to `emojis` to reflect that multiple emoji are allowed?

The word *emoji* is already both singular and plural, so the current name is good.

> Should we update the unit tests of `emoji` to reflect the changes in this PR?

Done.

fabian-hiller:
This PR looks good to me now. Can I merge it?

slevithan:
LGTM!

fabian-hiller:
v0.37.0 is available

Movsar-Khalakhoev:
Due to changes in `EMOJI_REGEX`, version 0.37.0 does not work on React Native. 
[Hermes had problems](https://github.com/facebook/hermes/pull/1295) working with regular expressions. Now the problem is fixed, but the update has not yet reached the RN

fabian-hiller:
What do you recommend? Wait for the fix to be released or make changes to Valibot?

slevithan:
@Movsar-Khalakhoev following a chain of issues from the one you linked to, I see that Hermes' previous lack of support for Unicode properties was leading to Zod not working with it due to the same regex that Valibot was using prior to 0.37.0. See: https://github.com/StefanTerdell/zod-to-json-schema/issues/129

So I'm not sure why the previous version of Valibot would have worked with Hermes if this is the issue. It might be helpful to provide more details about the issue you're seeing.

slevithan:
Also, although I haven't verified that the Hermes/RN fix is working, I'm assuming the Hermes fix (https://github.com/facebook/hermes/pull/1295) was released in Hermes [0.13.0](https://github.com/facebook/hermes/releases/tag/v0.13.0) a few days ago, and that this was then included in the React Native [0.75.1](https://github.com/facebook/react-native/releases/tag/v0.75.1) release.

PR #712, opened 2024-07-11
add 'extends' to `CustomAsyncSchema` def of the API ref

EltonLobo07:
Also, is the "Special schema async type." description intentional? I think it should be "Custom schema async type."


fabian-hiller:
Yes, it should be custom. Feel free to update the description.

PR #726, opened 2024-07-18
add `awaitAsync` API reference

EltonLobo07:
Since `awaitAsync` did not have an equivalent sync action (which is obvious but mentioned for context), I had to fill in the "related" schemas, methods, and utils. So I might have made a mistake, especially relating some methods to `awaitAsync`. I'll share my thoughts:

- For schemas, it is straightforward, as I can use any schema after awaiting the `promise` schema. That means, every schema can be listed.
- For methods, I copied the list from the `promise` schema and separated the sync and async methods properly. I think if a user needs to check the resolved value of a promise, the user will have to use the `promise` schema along with the `awaitAsync` action always. Then can proceed using any combination of schemas, methods and actions. 

Let me know if the reasoning is flawed.

### DIFF
--- a/website/src/routes/api/(async)/awaitAsync/index.mdx
+++ b/website/src/routes/api/(async)/awaitAsync/index.mdx
@@ -101,10 +101,6 @@ The following APIs can be combined with `awaitAsync`.
   ]}
 />
 
-### Methods
-
-<ApiList items={['config', 'getDefault', 'getFallback']} />
-
 ### Utils
 
 <ApiList items={['isOfKind', 'isOfType']} />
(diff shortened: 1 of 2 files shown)


fabian-hiller:
> For schemas, it is straightforward, as I can use any schema after awaiting the promise schema. That means, every schema can be listed.

👍

> For methods, I copied the list from the` promise` schema and separated the sync and async methods properly...

I would only list the `pipe` method. All other methods are not directly related to `awaitAsync`.

fabian-hiller:
Thank you!

PR #727, opened 2024-07-18
rename "/MaxSizeIssue " API ref folder to "MaxSizeIssue"

EltonLobo07:
I wasn't able to pull remote changes from this GitHub repository because `MaxSizeIssue`'s API reference folder had a trailing space, and trailing spaces [don't work as expected](https://stackoverflow.com/questions/68133091/git-commands-fail-with-spaces-in-path) on Windows. I assume the folder name "MaxSizeIssue " instead of "MaxSizeIssue" was a mistake. I have checked all of the references to the folder after the change, and everything works as expected. I guess the trailing space was ignored by the router.


fabian-hiller:
Great catch! Thank you!

Issue #728, opened 2024-07-18
Proposal to Make Prettify Type in Valibot Overridable for Better Developer Control


devcaeg:
The `Prettify` type is an internal type used by `Valibot`. It acts as a “trick” in `TypeScript` to “unroll” the types, making them easier for the end user to understand. However, the `Prettify` type can cause issues because it “decomposes” the types, potentially leading to uncommon errors in `TypeScript`. Therefore, I propose that this type be declared in a `declare module 'valibot'` block. This would allow developers to override the type in their projects if, for example, they want to negate its effects.

```ts
declare module 'valibot/types' {
    export type Prettify<TObject> = {
        [TKey in keyof TObject]: TObject[TKey];
    } & {};
}
```

fabian-hiller:
Thanks for bringing this up! I agree and will look into it, but it may take some time as other things have a higher priority for me right now.

devcaeg:
@fabian-hiller 

I just created a pull request, feel free to review it when you have time. Thank you very much!

PR #729, opened 2024-07-18
Make Prettify Type in Valibot Overridable for Better Developer Control

devcaeg:
#728

### DIFF
--- a/library/src/methods/flatten/flatten.ts
+++ b/library/src/methods/flatten/flatten.ts
@@ -4,9 +4,9 @@ import type {
   BaseSchemaAsync,
   InferIssue,
   IssueDotPath,
-  Prettify,
 } from '../../types/index.ts';
 import { getDotPath } from '../../utils/index.ts';
+import type { Prettify } from 'valibot/types';
 
 // TODO: Add unit and type tests for flatten method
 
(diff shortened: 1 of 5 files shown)


fabian-hiller:
Thanks for the implementation! Should we call it `valibot` instead of `valibot/types`? Are there any downsides to this change? I need to do some research on my own before merging.

colinhacks:
Are you sure you're able to override `type` declarations in this way? In my experiments I get a "Duplicate identifier" error.

devcaeg:
> Thanks for the implementation! Should we call it `valibot` instead of `valibot/types`? Are there any downsides to this change? I need to do some research on my own before merging.

In reality there should be no difference.

devcaeg:
> Are you sure you're able to override `type` declarations in this way? In my experiments I get a "Duplicate identifier" error.

I use this method on several projects and it works well. I do the override in `.d.ts` files.

devcaeg:
For example, the `adonisjs` framework uses this method.

https://docs.adonisjs.com/guides/digging-deeper/emitter#making-events-type-safe

fabian-hiller:
Thanks for the info and the link. I will try to review it next week.

fabian-hiller:
> For example, the `adonisjs` framework uses this method.

Do they also use a `*.d.ts` file internally that can be overwritten by a custom `*.d.ts` file?

PR #732, opened 2024-07-20
fix `fallbackAsync` for async arg & add relevant tests

EltonLobo07:
This PR fixes the issue related to `fallbackAsync` given any async schema.

Issue:

When the validation of an async schema fails, `fallbackAsync` still returns the default value instead of the supplied fallback value. The code block given below exposes this bug.
```
import * as v from 'valibot';

const Schema = v.fallbackAsync(
  v.pipeAsync(
    v.string(),
    v.checkAsync(async (value) => false)
  ),
  "fallback"
);

// logs: default
console.log(await v.safeParseAsync(Schema, "default"));
```
Fix:

Inside the `_run` method of the object returned by `fallbackAsync`, wait for the schema to run completely before checking if the dataset has any issues. This is needed as `fallbackAsync` can be given an async schema that might execute some async code.

Additional information:

Simply waiting for the async code to complete fixes the issue, but I was worried about the rejected promise case. I checked some of the other methods and found out that there is no logic to catch rejected promises. I assume this is intentional, and the thought process behind this approach would be to let the user handle rejected promises. Let me know if my assumption is incorrect.

### DIFF
--- a/library/CHANGELOG.md
+++ b/library/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to the library will be documented in this file.
 - Add `base64` action to validate Base64 strings (pull request #644)
 - Refactor `HEXADECIMAL_REGEX` (pull request #666)
 - Change `EMOJI_REGEX` to be more accurate and strict (pull request #666)
+- Fix bug in `fallbackAsync` method for async schemas (pull request #732)
 
 ## v0.36.0 (July 05, 2024)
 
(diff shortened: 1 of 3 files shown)


fabian-hiller:
Thanks for this fix! Yes, at the moment we expect the user to catch async errors.

fabian-hiller:
v0.37.0 is available

PR #744, opened 2024-07-25
add `getFallbacksAsync` API reference

EltonLobo07:
I have used the same example that was added to `fallbackAsync` as I was unable to think of a better example. I'll try to come up with better examples once most of the API references are added. Feel free to update or add examples if you have something better.


fabian-hiller:
Thank you very much! I will review it later, but in general it is perfectly fine to use the same example. I did the same thing with some references.

PR #759, opened 2024-08-01
add `objectWithRestAsync` API reference

EltonLobo07:


fabian-hiller:
I may remove the `NewUserSchema` example as I don't think we should "recommend"  `never` as the rest schema.

EltonLobo07:
The reason I added it is because it is mentioned on this webpage, and this is a good webpage for an example of such a schema. But I agree - if we don't recommend it, then it's better not to add it as an example.

PR #773, opened 2024-08-09
Add babel-plugin-transform-regex to devDependencies

### LINKED ISSUE #704: Improve DX with `babel-plugin-transform-regex`
Thoughts on adding [babel-plugin-transform-regex](https://github.com/slevithan/babel-plugin-transform-regex) to `devDependencies`?

This would allow changing e.g. the unreadable/unmaintainable `IPV4_REGEX`:

```js
/^(?:(?:[1-9]|1\d|2[0-4])?\d|25[0-5])(?:\.(?:(?:[1-9]|1\d|2[0-4])?\d|25[0-5])){3}$/u
```

To the much nicer:

```js
regex`^
  (?<byte> 2[0-4]\d | 25[0-5] | 1\d\d | [1-9]?\d )
  ( \. \g<byte> ){3}
$`
```

This would then get transpiled into a native regex (you can try it [here](https://slevithan.github.io/babel-plugin-transform-regex/demo/)), without the need for any runtime dependency, any added run-time cost for users, or any interpolation when constructing the regex (which could cause issues with tree-shaking).

This is just an example, and some of the other regexes in `library/src/regex.ts` could benefit more significantly.

(This is a follow up from #666, where interpolation was removed from regex changes due to tree-shaking concerns.)

slevithan:
Closes #704.

Per previous discussion with @fabian-hiller in #704, this adds package [babel-plugin-transform-regex](https://github.com/slevithan/babel-plugin-transform-regex) to `devDependencies` for improved DX.

- Tests, lint, prettier, and build all pass.
- Renamed `library/src/regex.ts` as `library/src/regex.src.ts`.
  - `regex.ts` is now the transpiled output.
- In `regex.src.ts`, I updated only the first two regexes (`BASE64_REGEX` and `BIC_REGEX`) to use the `regex` tag with more readable syntax (using free spacing, subroutines, a subroutine definition group, and noncapturing `(…)`).
  - The emitted source in `regex.ts` for these two regexes is identical to before, which helps demonstrate that the Babel plugin is doing a great job with producing efficient and readable native regexes (since the regexes being replaced were hand-tuned regex literals). As a result, this PR's only change to `regex.ts` is an added header comment.
  - Other regexes and the rest of the code in `regex.src.ts` (apart from a hint comment at the top) is untouched.
  - I'll update the remaining regexes (many of which will benefit much more than these two, which are quite simple) after this PR lands.
- Also added package `regex` to `devDependencies` so the types work in `regex.src.ts`.
  - The `import { regex } from 'regex';` line in `regex.src.ts` is stripped from transpiled source by the Babel plugin (via its `removeImport` option which removes any import declarations with module name `'regex'`).
  - To make the types for `regex` work, I had to update `library/tsconfig.json`, changing `"moduleResolution": "node"` to [`"Bundler"`](https://www.typescriptlang.org/docs/handbook/modules/reference.html#bundler). Values `"NodeNext"` and `"Node16"` would also work for `regex.src.ts`, but those values conflict with other, current `tsconfig.json` settings.
    - This change doesn't seem to cause any issues elsewhere, but I'm not deeply knowledgeable about TS/CJS/bundling so it's worth examining.


fabian-hiller:
Thank you for creating this PR. Since I have less time at the moment and other things have a higher priority, I will try to get back to you next week.

fabian-hiller:
Is it not possible to do the transformation as part of our build step? I don't like that we need a `regex.ts` and `regex.src.ts`.

slevithan:
That sounds good, with the slight downside that you won't be able to review the emitted regex literals as easily (without looking at the output in `dist`).

I have a working solution that I just updated the PR with. It generates a `regex.gen.ts` file to use for the build, and then deletes it afterward. tsup (and esbuild which it relies on) doesn't have a built-in way to alias local file import paths, but I created a tiny esbuild plugin in `tsup.config.ts` that I'm using for this.

If this seems a little too magical, we could also use a build process like the following:

1. Copy (or rename) `regex.ts` to a temp file (lets say `regex.orig.ts`).
2. Run Babel on the temp file, with the output overwriting `regex.ts`.
3. Run the regular build.
4. Delete the generated `regex.ts`, then rename the `regex.orig.ts` copy back to `regex.ts`.

But my initial impression is that moving files around like that would not be preferable.

fabian-hiller:
Thanks for the update! I am pretty busy at the moment and mainly focused on getting our v1 RC and after that our v1 release ready. I will come back to this PR when I have more time again or this feature becomes more important.

slevithan:
No problem, take your time. If you have a moment, though, I'm curious whether you have concerns with the current approach (or need time to think about it), or if you think the current PR is on the right track to pick back up from when you have more time.

library/package.json
```diff
@@ -49,22 +49,31 @@
     "eslint.fix": "eslint \"src/**/*.ts*\" --fix",
     "format": "prettier --write ./src",
     "format.check": "prettier --check ./src",
+    "babel.regex": "babel src/regex.ts --out-file transpiled/regex.ts",
+    "prebuild": "pnpm babel.regex",
     "build": "tsup",
+    "postbuild": "rimraf transpiled",
```
fabian-hiller, line 55: Is `rimraf` needed here? What is the disadvantage of `rm -rf`?

library/tsup.config.ts
```diff
@@ -1,4 +1,26 @@
...
+const localAliases: {[key: string]: string} = {
+  // Use only the filename as the key; it's handled for imports with any relative path
+  'regex.ts': './transpiled/regex.ts',
+};
+
+const localAliasesPlugin = {
+  name: 'localAliases',
+  setup(build: {onResolve: Function}) {
+    // Handle `./file.ext`, `../../file.ext`, etc.
+    const re = new RegExp(`^(?:\./|(?:\.\./)+)(?:${Object.keys(localAliases).map(
```
fabian-hiller, line 13: Is such a complex regex really necessary?

fabian-hiller:
In general, I don't like to complicate the setup because I like things simple, but of course I see the benefits of this PR.

Is it possible to run our tests with the compiled output? For security reasons this seems to be necessary in the long run.

library/package.json
slevithan: I added `rimraf` because of concerns about `rm -rf` working cross-platform, but I'm not sure where this would actually be a problem so I'll go ahead and replace it with `rm -rf`.

slevithan:
> In general, I don't like to complicate the setup because I like things simple, but of course I see the benefits of this PR.

I share your concern. This was actually the primary reason for my initial approach of having a separate "source" file, with `regex.ts` being the transpiled output. But I see the downside of that as well.

I've simplified things somewhat in the latest update to the PR, including getting rid of `rimraf`, and narrowing the aliasing plugin (which can always be generalized later if for whatever reason more paths need to be aliased).

> Is it possible to run our tests with the compiled output? For security reasons this seems to be necessary in the long run.

I've updated the PR to do so. Vite/Rollup share the [resolve.alias](https://vitejs.dev/config/shared-options#resolve-alias) option, so this was easy to do. Their local aliasing support seems very robust (and allows regexes or functions, which I didn't use here).

fabian-hiller:
Thank you for the changes! This PR is now going in a good direction.

slevithan:
Cool. In the meantime, I've made a couple more tweaks and I've gone ahead and updated the PR to cover the first 12 of 24 regexes in `regex.ts`. Hopefully this shows more substantively what the end result will be (I think it makes sense to still wait on the remainder until after this PR lands). Of the first 12 regexes, a couple (`CUID2_REGEX` and `DECIMAL_REGEX`) were so simple that they didn't benefit from `regex`, so I left them alone. I also included the following comment above `IP_REGEX`:

```js
// Q: Would it be tree-shakeable if this used interpolation without adding any new variables, as
// ``new RegExp(`^(?:${IPV4_REGEX.source.slice(1, -1)}|${IPV6_REGEX.source.slice(1, -1)})$`, 'iu')``?
```

If the answer is yes, that would remove a lot of redundancy. That redundancy is much easier to verify now that I've rewritten the regexes in a readable way.

**Important:** None of the rewritten regexes have changed what they match in any way (existing bugs in `IPV6_REGEX` and `IP_REGEX` have been preserved).

slevithan:
I've rebased on top of recent conflicting Valibot configuration changes, and additionally bumped `regex` to v4.1.3. The latest version of `regex` avoids the need to update Valibot's `library/tsconfig.json` to use `"moduleResolution": "Bundler"`. So I've undone that change.

slevithan:
Congrats on the v1 launch. 😊 Does it make sense to pick this back up now that v1.0.0 has been out for a while?

fabian-hiller:
Hey @slevithan 👋 I'm very sorry for my late reply, but I don't have much time at the moment. So I have to be really careful with prioritizing things. Right now I should probably focus on adding new actions, improving our docs, and educating people on why Valibot is better than alternative libraries in various use cases. I think I would keep our regexes as they are and maybe test our library if we add more complex regexes at some point in the future.

slevithan:
No worries, mate. Should we close this PR?

Issue #776, opened 2024-08-12
Preserve `@ts-expect-error` directives


alecmev:
As mentioned in https://github.com/fabian-hiller/valibot/issues/659#issuecomment-2183564529, some prefer to keep `skipLibCheck` off. It's not about being pedantic, it can help explain otherwise cryptic errors (speaking from experience), and puts pressure on the ecosystem to keep their types clean.

However, in this case it isn't an easy fix at all (looks like `QuestionMarkSchema` being recursive pushes the limits of inference, https://github.com/microsoft/TypeScript/issues/30134, https://github.com/microsoft/TypeScript/issues/47599). Hit my head for half an hour, to no avail.

Thankfully, this repo already has `// @ts-expect-error` peppered throughout:

https://github.com/fabian-hiller/valibot/blob/094d9845e4d5c90659ac61d2627dba811568d07a/library/src/types/other.ts#L90-L99

So all that needs to be done is just preserve these in the emitted `.d.ts`. Easier said than done, as I haven't been able to find if `tsc` can even do it, but worst case, at least it could be a post-build script just adding these back in above `QuestionMarkSchema` and `QuestionMarkSchemaAsync`, and inside of `TuplePath`.

fabian-hiller:
A better solution would be to figure out how to avoid the recursive inference problem. If we can't find a solution, we could try to limit the recursiveness to a certain depth. Are you interested in exploring this problem?

alecmev:
Done 😉 https://github.com/fabian-hiller/valibot/pull/778

fabian-hiller:
Related issue at the TS repo: https://github.com/microsoft/TypeScript/issues/38628

alecmev:
Good find! Looking at that issue, how about `/** @ts-expect-error */`? TypeScript preserves doc comments.

fabian-hiller:
This may work, but it is a dirty hack. I would prefer an official solution, or that we rethink our implementation. But I will consider it.

fabian-hiller:
This will probably be fixed in the next version.

fabian-hiller:
v0.38.0 is available 🚀

Issue #659, opened 2024-06-17
TypeScript compile fails


gflohr:
I can no longer build a package depending on valibot version 0.32.0 or 0.31.0. Downgrading to 0.30.0 fixes the issue.

`index.ts`:

```typescript
import * as v from 'valibot';
console.log(v.object({}));
```

`package.json`:
```
{
  "name": "valibot-build",
  "scripts": {
    "build": "rm -rf dist; tsc"
  },
  "dependencies": {
    "valibot": "^0.32.0"
  },
  "devDependencies": {
    "typescript": "^5.4.5"
  }
}
```

`tsconfig.json`:
```json
{
	"compilerOptions": {
		"allowSyntheticDefaultImports": true,
		"noImplicitAny": true,
		"noImplicitReturns": true,
		"noImplicitThis": true,
		"removeComments": true,
		"declaration": true,
		"target": "ES2017",
		"lib": ["ES2017", "dom"],
		"module": "CommonJS",
		"sourceMap": true,
		"typeRoots": ["node_modules/@types"],
		"esModuleInterop": true,
		"moduleResolution": "node",
		"strictNullChecks": true,
		"strictBindCallApply": true,
		"forceConsistentCasingInFileNames": true,
		"noFallthroughCasesInSwitch": true,
		"outDir": "dist",
		"resolveJsonModule": true
	},
	"exclude": ["node_modules"],
}
```

I have played around with the `tsconfig.json` but to no avail. The build fails:

```sh
$ npm run build

> build
> rm -rf dist; tsc

node_modules/valibot/dist/index.d.ts:5678:189 - error TS2344: Type 'QuestionMarkSchema' does not satisfy the constraint 'BaseSchema<unknown, unknown, BaseIssue<unknown>>'.
  Type 'LazySchema<QuestionMarkSchema>' is not assignable to type 'BaseSchema<unknown, unknown, BaseIssue<unknown>>'.
    The types returned by '_run(...)' are incompatible between these types.
      Type 'Dataset<unknown, unknown>' is not assignable to type 'Dataset<unknown, BaseIssue<unknown>>'.
        Type 'TypedDataset<unknown, unknown>' is not assignable to type 'Dataset<unknown, BaseIssue<unknown>>'.
          Type 'TypedDataset<unknown, unknown>' is not assignable to type 'TypedDataset<unknown, BaseIssue<unknown>>'.
            Type 'unknown' is not assignable to type 'BaseIssue<unknown>'.

[...]

PR #778, opened 2024-08-13
Fix public API errors

alecmev:
Fix https://github.com/fabian-hiller/valibot/issues/776
Fix https://github.com/fabian-hiller/valibot/issues/659

The `getDotPath` test seems alright, but I'm not sure if the `lazy` one actually hits the changes, it's a lot of layers to unravel. Can you investigate, please? Also, what's the reason these types were recursive in the first place? Don't mean to pull a Chesterton’s Fence here.

### DIFF
--- a/library/src/schemas/lazy/lazy.test-d.ts
+++ b/library/src/schemas/lazy/lazy.test-d.ts
@@ -1,5 +1,12 @@
 import { describe, expectTypeOf, test } from 'vitest';
-import type { InferInput, InferIssue, InferOutput } from '../../types/index.ts';
+import type {
+  GenericSchema,
+  InferInput,
+  InferIssue,
+  InferOutput,
+} from '../../types/index.ts';
+import { object } from '../object/object.ts';
+import { optional } from '../optional/optional.ts';
 import {
   string,
   type StringIssue,
@@ -28,5 +35,25 @@ describe('lazy', () => {
     test('of issue', () => {
       expectTypeOf<InferIssue<Schema>>().toEqualTypeOf<StringIssue>();
     });
+  });
+
+  test('should handle recursion', () => {
+    interface BinaryTree {
+      element: string;
+      left?: BinaryTree;
+      right?: BinaryTree;
+    }
+
+    const BinaryTreeSchema: GenericSchema<BinaryTree> = object({
+      element: string(),
+      left: optional(lazy(() => BinaryTreeSchema)),
+      right: optional(lazy(() => BinaryTreeSchema)),
+    });
+
+    expectTypeOf<
+      NonNullable<
+        NonNullable<InferInput<typeof BinaryTreeSchema>['left']>['right']
+      >
+    >().toEqualTypeOf<BinaryTree>();
   });
 });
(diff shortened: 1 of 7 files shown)


alecmev:
Can you add a failing test to this PR, please?

fabian-hiller:
`key6` and `key7` should be `undefined` but without a question mark. Since every `LazySchema` and `NonNullableSchema` now matches `QuestionMarkSchema`, a question mark is added.

alecmev:
Thank you! Pushed a fix. I'll be honest, I still don't know how to work around the recursion issue, so this is pure brute-force, but it seems to work. Any potential problems?

fabian-hiller:
I think you fixed it for exactly one layer, but if we nest multiple `lazy` schemas it still breaks. That's why I think the recursive implementation is necessary for it to work properly.

fabian-hiller:
The only and probably simplest way to solve this problem is to define the rule that `object` only adds a question mark if `optional` or `nullish` is the outermost schema of an entry. This way we can remove the recursive implementation.

```ts
// With question mark
v.object({ key: v.optional(v.lazy(() => v.string())) }); // { key?: string | undefined } 

// Without question mark
v.object({ key: v.lazy(() => v.optional(v.string())) }); // { key: string | undefined } 
```

fabian-hiller:
This would introduce a breaking change, but since we are still v0, this is ok. I like that this rule simplifies the implementation and makes the library usable for more users. But I don't know if there are any major drawbacks yet.

alecmev:
Tried every possible hack that I could come up with, no luck. Can't dedicate more time to this, sorry. How about this:

- Keep the new tests, for the future.
- Drop the `QuestionMarkSchema` changes.
- Replace the ignores with `/** @ts-expect-error */` for now, don't have to commit.
- Report this in the TypeScript repo, maybe they'll help.

I wouldn't change the API, doesn't feel right to remove something that ultimately works.

fabian-hiller:
> Tried every possible hack that I could come up with, no luck. Can't dedicate more time to this, sorry.

Don't worry about it. I appreciate every contribution! So thank you for the time you have already invested in Valibot. 🙏

> I wouldn't change the API, doesn't feel right to remove something that ultimately works.

I see what you mean, but someone could also argue why there is a question mark when `optional` is nested very deep. It actually makes sense to allow users to control this behavior by intentionally using `optional` and `nullish` as the outermost schema within `object`.

I think that Valibot could become a big and important project in the long run. Therefore, it would be nice to solve the TS error problem and introduce clear rules for such behavior.

alecmev:
Thank you for caring 😉 

Re the API change - what you say makes sense to me, but I think I shouldn't participate in this decision, as I haven't even used `.lazy` or `.nonNullable`.

fabian-hiller:
Another thing you could try is to install Valibot via JSR, as it should ship our source code including the comments: https://jsr.io/@valibot/valibot

fabian-hiller:
I fixed the TS errors. Thanks for your cooperation!

alecmev:
Great, thanks! 👍

PR #786, opened 2024-08-16
add `parserAsync` API reference

EltonLobo07:
I noticed the 'Explanation' section is missing under 'Parameters'. I assume this is because `parserAsync` returns a function instead of directly using the arguments to generate a final output. But I think an explanation like the one on [this](https://valibot.dev/api/parseAsync/) webpage is needed. Maybe we can add a similar explanation under the 'Returns' section. Good idea?


fabian-hiller:
Yes, feel free to improve this in another PR 🙏

PR #802, opened 2024-08-26
Initialize @valibot/to-json-schema package

gcornut:
Following the discussions on https://github.com/gcornut/valibot-json-schema/issues/84, here is my proposal on a new package for converting Valibot to JSON schema to be moved here next to Valibot core library code :) 

## The setup

1. New packages/to-json-schema NPM package named `@valibot/to-json-schema`
2. Some basic schema & validation conversions
    - `any` schema
    - `null` schema
    - `string` schema
    - `object` schema
    - `email` validation
3. Lint, Types and Tests (100% coverage) 
 
I tried to match the code style of the core Valibot library.

## What is next ?

- More conversions: all the supported schemas, validations and metadata already [supported by `@gcornut/valibot-json-schema`](https://github.com/gcornut/valibot-json-schema?tab=readme-ov-file#supported-features)
- More conversion options
  - Convert schemas into JSON schema definitions
  - Convert non JSON types to similar, yet different types (for example: `Date` into `format: 'date-time'` or `format: 'unix-time'`, `BigInt` into `format: 'int64'` or string, explicit `undefined` into `null`, etc.)
- Customize which JSON schema features to use ?
   => Some tools (like [OpenAI structured output](https://platform.openai.com/docs/guides/structured-outputs/supported-schemas)) does not support all the features of JSON schema. Maybe we could add a configuration to skip those features ?

I'll deprecate and archive `@gcornut/valibot-json-schema` once this lib gets to feature parity. Although I will keep the CLI in a new repo because it's a use case I still need.


.github/workflows/ci.yml
```diff
@@ -79,3 +79,42 @@ jobs:
...
+    needs: install
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup pnpm
+        uses: ./.github/actions/pnpm
+      - name: Vitest tests
+        run: pnpm test
+        working-directory: packages/to-json-schema
```
gcornut, line 120: Could be done cleaner with GH job matrix 🤔

packages/to-json-schema/src/schema-converters/convertSchema.ts
```diff
@@ -0,0 +1,92 @@
...
+  const converted = converter(schema, convertSchema);
+
+  // Convert pipe items
+  if (hasPipe(schema)) {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [_, ...pipeItems] = schema.pipe;
+    for (const pipeItem of pipeItems) {
+      if (pipeItem.kind === 'validation') {
+        // Merge converted validation into it's base schema
+        Object.assign(converted, convertValidation(pipeItem, options));
```
gcornut, line 84: It might be a better idea to give the base schema to the validation converter let it decide how to modify the base schema. 
I'm thinking that `pipe(string(), nullable())` could be converted to `{ type: ['string', 'null'] }` but only if the base schema (here string) has a type. Some schemas (like `anyOf`) do not need a type in JSON schema

packages/to-json-schema/src/types.ts
```diff
@@ -0,0 +1,5 @@
+/** Conversion options */
+export interface ConversionOptions {
+  /** Force conversion to JSON Schema even on incompatible schemas. */
```
gcornut, line 3:
```suggestion
  /** Force conversion to JSON Schema on unsupported Valibot features (schema, validation, transformation or metadata). */
```

.github/workflows/ci.yml
fabian-hiller: I don't know what GH job matrix is but its probably an improvement for another PR

fabian-hiller:
Thank you! I will probably give you feedback on everything by tomorrow.

fabian-hiller:
Thanks again for the PR. I have reviewed everything and also investigated several alternative implementations. I even tried a fully typesafe implementation where the output type was the actual JSON schema output. Unfortunately, the typesafe implementation got quite complex and probably too hard to maintain at this point. Since Valibot is not a JSON-schema-first library, I don't expect many users to expect such an implementation. However, I have added an alternative implementation to `/src2` with a focus on simplicity, bundle size and performance. Can you take a look at it and let me know what you think? Once we have decided on the basic structure of the source code, we can continue to add support for more schemas and actions.

packages/to-json-schema/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "@typescript-eslint/eslint-plugin": "^7.15.0",
+    "@typescript-eslint/parser": "^7.15.0",
+    "@vitest/coverage-v8": "^1.6.0",
+    "eslint": "^8.57.0",
+    "eslint-plugin-import": "^2.29.1",
+    "eslint-plugin-jsdoc": "^48.5.2",
+    "eslint-plugin-security": "^2.1.1",
+    "tsup": "^8.1.0",
+    "typescript": "^5.5.3",
+    "valibot": "workspace:*",
```
gcornut, line 67: Depending on the workspace version of valibot here is convenient but it actually produces errors (in IDE, during type check, test, build) when the valibot does not have a built `dist` folder (the default state when cloning the repo)

I initially added `paths` in the tsconfig to tell TS & vite to use the raw source files instead.

One other solution could be to always build valibot before running anything in this package. 

Or we could just use valibot from NPM instead of the workspace but it would need to be updated at each new valibot release.

What do you think would be the best solution here ? 🤔

fabian-hiller: Sorry, that was my fault. Please undo my changes. I will ask next time.

fabian-hiller:
Thank you! Would you change anything? Otherwise, feel free to clean up the code and add support for more schemas and actions after we answer my next question.

I saw that [TypeBox's](https://github.com/sinclairzx81/typebox) `Type.Undefined()` returns `{ type: 'undefined' }`, but this does not seem to be supported by JSON Schema Draft 7 (according to the `JSONSchema7` type). Should we upgrade to a newer one? Should we even specify the version we are using? Are you aware of any breaking changes in newer versions of JSON Schema that would result in different output?

gcornut:
> saw that [TypeBox's](https://github.com/sinclairzx81/typebox) Type.Undefined() returns { type: 'undefined' }, but this does not seem to be supported by JSON Schema Draft 7 (according to the JSONSchema7 type)

I don't see how JSON schema could support `undefined` since JSON does not support `undefined` (only `null`) 🤷

> Should we upgrade to a newer one? Should we even specify the version we are using? Are you aware of any breaking changes in newer versions of JSON Schema that would result in different output?

Yes their are newer versions of the JSON schema specs (which some major changes 😬)  but most tools use the draft-07 version (OpenAI, Vercel AI, etc.)

And actually if we explicitly want to support OpenAI structured output we would need to have a subset of JSON schema [since it does not support every feature](https://platform.openai.com/docs/guides/structured-outputs/some-type-specific-keywords-are-not-yet-supported). For Vercel AI I can't see document that suggest they don't support all JSON schema features 🤷.

We could add a `target?: 'json-schema-draft-07' | 'openai-structured-outputs'` that could change the output and fail on unsupported feature (with still the possibility of forcing the conversion).

gcornut:
Some people JSON schema only for documentation purpose so maybe for them we could add another output "target" that could output things like `type: 'undefined'` which are unusable to validation a JSON payload but maybe useful for documentation.

fabian-hiller:
Thanks for your detailed answer. For now, I think it is best to optimize everything for JSON Schema Draft 7 and thread everything else as not supported (`force` can still be used). In the long run, we will evaluate how people use this package and change or add functionality based on that. Feel free to clean up the code and add support for more schemas and actions.

gcornut:
How strict do you think this lib should be on the input schema ? 

I was thinking that the default value in optional/null/nullish schemas and the value in the literal schema should only be JSON compatible values. And actually these values needs to be strictly compatible, excluding types like `Date` (that does serialize to JSON but as a string) or `Infinity` (that serializes to `null` in JSON).

Maybe the actions we convert should check on which type they are applied (`email` is only applicable to `string`, `integer` on `number`, etc.) ? Valibot does mostly guard against action/schema missmatch, but only via TS type ?

It's probably overkill but we could validate the input schema via a valibot meta schema 😄

fabian-hiller:
> I was thinking that the default value in optional/null/nullish schemas and the value in the literal schema should only be JSON compatible values...

I think that `toJsonSchema` will throw anyway if the wrapped schema is not JSON schema compatible, or am I wrong? If I'm right, I think we don't need to worry about the type of the default value. `Infinity` and `NaN` are probably the only exceptions, but it is a very very rare case.

> Maybe the actions we convert should check on which type they are applied...

I think we should not worry about that. Our API is meant to be used in a certain way, and we should not have to deal with people using it incorrectly.

gcornut:
Alright so I've rebased from the branch, I've added almost all possible schema & action conversion. If it's too big, I can split the PR. I already tried to split this in commits.

I need to add optional conversions for date and bigint since they are not technically supported in JSON but can be converted to number or string depending on the use case.

fabian-hiller:
Thank you so much! It is ok for me if this PR is "big". I will try to review everything in the next days with the goal to release this package next week.

gcornut:
Ok I've added optional date and bigint schema conversion (requiring "strategy" configs)

fabian-hiller:
I am not sure if we should add the strategy configs from the beginning. I would prefer to wait until people come up with good use cases and then decide on the implementation. But feel free to share your opinion since your conversion library has a longer history.

gcornut:
Yes it might be a bit overkill.

I know sveltekit-superforms use date schema conversion to integer (https://github.com/ciscoheat/sveltekit-superforms/blob/main/src/lib/adapters/valibot.ts#L30). But I think they use JSON schema for a documentation/metadata use case. They don't need a strict schema but just a way to describes schema no matter the implementation (zod, joi, yup, valibot, etc.)

fabian-hiller:
Instead of offering such configurations, in the long run we could offer a general way to modify the output of `toJsonSchema`, similar to how it works with `JSON.stringify`.

fabian-hiller (commented):
At first glance, everything looks great. Many thanks for your contribution! If you are interested, please feel free to add your name to `LICENSE.md`.

Next, I will refactor a few code lines and then check the implementation and tests in detail.

packages/to-json-schema/src/type.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export interface JsonSchemaConfig {
+  /**
+   * Whether to force conversion to JSON Schema even for incompatible schemas and actions.
+   */
+  force?: boolean;
+  /**
+   * The name of the definitions property used. Defaults to "$defs".
+   */
+  definitionPath?: '$defs' | 'definitions';
```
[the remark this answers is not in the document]
gcornut: ´definitions´ is the old name but JSON schema draft07 supports both and a lot of tooling tend to use the old name.

I don't have concrete example so we can remove this config and wait for users to file issues on that !

gcornut:
> If you are interested, please feel free to add your name to LICENSE.md.

Thanks but no need to do that 🙂

> Next, I will refactor a few code lines and then check the implementation and tests in detail.

Ok great 👍

packages/to-json-schema/src/type.ts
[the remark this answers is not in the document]
fabian-hiller: I am not familiar enough with JSON schema to know what is best. If you have a strong opinion on this, feel free to share it. Otherwise we can add it in the future if needed.

packages/to-json-schema/src/convertSchema.ts
```diff
@@ -0,0 +1,348 @@
...
+    case 'nullable':
+    case 'nullish': {
+      json.anyOf = [
+        convertSchema({}, schema.wrapped as SchemaOrPipe, context),
+        { type: 'null' },
+      ];
+      const defaultValue = v.getDefault(schema);
+      if (
+        defaultValue !== undefined &&
+        assertJSON(
```
fabian-hiller, line 173: Is it even necessary to check if the default value is JSON compatible? Valibot already forces the user to match the value to the input type of the schema, and any invalid schema will result in an error anyway. The only expectation is values like `NaN` or `Infinity`, but that's a rare case I think we can ignore for now.

fabian-hiller:
I started refactoring some parts and will commit my changes later. We will probably have a few implementation interactions in the next days before releasing the first version of the package next week.

packages/to-json-schema/src/convertSchema.ts
gcornut: But the default value here can be anything right ? It could be a Map, a Set, a Date, a File, etc.
All of which would break in JSON 🤔

gcornut: Ah but the wrapped schema would be incompatible.
Then yes it would only be for fringe cases. We could remove 👍

fabian-hiller:
I refactored the code and made the following bigger changes:

**Definitions:** I removed the `definitions` property from `config` as I think we can create them automatically as needed. Basically, whenever `convertSchema` sees a lazy schema that is not already part of the definitions, it will add it with a generated reference counting up from 0. If users want to explicitly define the definitions and references, we can add that later as needed.

**Assert JSON**: I replaced or removed the `assertJSON` utility because I could not find a case where a correct use of Valibot could actually lead to an invalid case. I may be wrong, so feel free to add comments and discuss this with me.

Sorry for partially overwriting or changing your code without asking, but I think this will help us ship the first version of this package faster. Please let me know if you want to revert or change any of my modifications.

I did not touch the tests. We can do that after we agree on the final implementation.

gcornut:
> **Definitions:** I removed the `definitions` property from `config` as I think we can create them automatically as needed. Basically, whenever `convertSchema` sees a lazy schema that is not already part of the definitions, it will add it with a generated reference counting up from 0. If users want to explicitly define the definitions and references, we can add that later as needed.

Named definitions seems an essential feature to me. I know I would have to hack it back into my scripts that exports my valibot schemas into JSON schema.

But I won't be blocking just for that if you want to proceed 👍

fabian-hiller:
I agree that this could be an essential feature. Therefore, I added support for explicitly specified definitions with my last commit.

fabian-hiller:
Should I continue with the final review and unit tests, or do you have more feedback on the current implementation?

gcornut:
> Should I continue with the final review and unit tests, or do you have more feedback on the current implementation?

All good for me 👍

fabian-hiller:
I think that I will merge this PR later and ship our v0.1.0 release 🎉

fabian-hiller:
@gcornut do you think we should completely ignore unsupported schemas and actions, or is it better to try to get it right? A concrete example is `minLength`. Should we add nothing to the output (and loose the information) if the schema type is not supported or should we add both `minLength` and `minItems` to the JSON Schema output?

fabian-hiller:
If we choose the second strategy, we might consider removing the `force` configuration and just `console.warn` when something is not supported.

gcornut:
Both solutions sound ok to me as long as the conversion is strict by default. 

Like mentioned earlier there is definitely some people who just want to output schemas for documentation purpose and don't care about having 100% correct JSON schema definitions.

But maybe we could look into a "loose" output format later

fabian-hiller:
What do you think about removing `force` and only `console.warn` in such cases?

gcornut:
> What do you think about removing `force` and only `console.warn` in such cases?

Sounds ok. If some user need to be noticed that the conversion was not completely successful we can think of adding another option later 👍

fabian-hiller:
I think I will keep `force` and `throw` when it is `false` or `undefined` and `console.warn` otherwise.

fabian-hiller:
Thanks so much @gcornut for your contribution! We are ready to merge! 🍾

fabian-hiller:
v0.1.0 is available

PR #809, opened 2024-08-31
Improve performance and issues for nested variants

### LINKED ISSUE #794: Nested variants
Hey @fabian-hiller,
i am currently trying to implement a relatively complex form using Valibot for validation. 
Here is a simplified example of what I am trying to accomplish

```ts
import * as v from 'valibot';

const Schema = v.variant('type', [
  v.variant('subType', [
    v.object({
      type: v.literal('yes'),
      subType: v.literal('yes'),
      test1: v.string(),
    }),
    v.object({
      type: v.literal('yes'),
      subType: v.literal('no'),
      test2: v.string(),
    }),
  ]),

  v.variant('subType', [
    v.object({
      type: v.literal('no'),
      subType: v.literal('yes'),
      test3: v.string(),
    }),
    v.object({
      type: v.literal('no'),
      subType: v.literal('no'),
      test4: v.string(),
    }),
  ]),
]);

const result = v.safeParse(Schema, {
  type: 'no',
  subType: 'yes',
});

if (result.issues) {
  console.log(v.flatten(result.issues));
}
```

This works fine if the data is valid but if it is not it always goes into the first subType variant and just shows the issues of that. 
Using union also is not really an option because that would give me issues for all different paths but I only want the ones for the current selection.

Something like  
```ts
v.multiVariant(['type', 'subType'], [
  //...
]),
```
would probably work but that seems relatively hard to implement.

Is there currently a way to accomplish something like this?

Thanks in advance! :)

fabian-hiller:
Fix #794

I have implemented a completely new algorithm for `variant` that supports multiple nestes discriminator keys. The following describes how the algorithm works.

The algorithm searches for the correct object schema to use to validate the input based on the specified discriminator keys. It gathers information about the first deepest invalid discriminator, if any. Discriminators present in the input have priority over missing ones. An object schema is used for validation only if all specified discriminators are valid. If no object schema is found that matches the specified discriminators, `variant` will use the collected information about the first deepest invalid discriminator to return a useful issue.


fabian-hiller:
I will merge this now. However, you can still give me feedback.

ZerNico:
@fabian-hiller sounds like a very good solution. Also thanks for the quick fix, you are awesome :)

For allowing multiple discriminators I am not sure I'd like the breaking change to force setting the keys as an array. Even if it's still v0 it might not be super wise to annoy users with breaking changes for features most will probably not use. Also makes the usage a bit more verbose for everyone. 
Though I agree having both options way to write would make it more confusing. Might still be the best solution for this case. Would be nice if TypeScript would allow rest params for the first param too, would have been a nice solution too

fabian-hiller:
Thanks for your feedback. Let's keep it as is for now, and in the long run I will consider supporting both a string and an array of strings for the first argument if more users request this API design.

PR #819, opened 2024-09-06
add `requiredAsync` & `SchemaWithRequiredAsync` API refs

EltonLobo07:
The `requiredAsync` reference with this PR is a bit different from the `required` reference on the website. After this PR is merged, I'll make sure to update the `required` reference (if there are any changes needed). Here are the differences along with the reasons:

1. The `required` reference does not use a type parameter for the message parameter but the `requiredAsync` reference does. I think we should list `TMessage` so that: 
    a. the reference is as close to the code as possible. 
    b. we display the link between the `message` parameter and the output schema in the reference.

2. The naming of the output schemas. Instead of naming the output `Schema`, it is better to name the schemas `AllKeysSchema` and `SelectedKeysSchema` so that readers can identify the schemas in the 'Returns' section.

### DIFF
--- a/website/src/routes/api/(async)/partialAsync/index.mdx
+++ b/website/src/routes/api/(async)/partialAsync/index.mdx
@@ -68,9 +68,9 @@ const UpdateUserSchema = v.partialAsync(UserSchema);
 
 /*
   { 
-    email?: string,
-    username?: string, 
-    password?: string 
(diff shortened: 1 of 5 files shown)


fabian-hiller:
You are right. The `required` API reference is not fully updated. Feel free to fix it.

PR #830, opened 2024-09-14
Fix conversion of recursive schema

gcornut:
Hello @fabian-hiller 👋 

It seems that we forgot to test the conversion of recursive schema 😬. 
It currently results in an error `RangeError: Maximum call stack size exceeded`.

I think I have a solution, let me know what you think of this or if you find a better solution

### DIFF
--- a/packages/to-json-schema/src/convertSchema.ts
+++ b/packages/to-json-schema/src/convertSchema.ts
@@ -118,7 +118,7 @@ export function convertSchema(
 ): JSONSchema7 {
   // If schema is in reference map, use reference and skip conversion
   const referenceId = context.referenceMap.get(valibotSchema);
-  if (referenceId) {
+  if (referenceId && referenceId in context.definitions) {
     jsonSchema.$ref = `#/$defs/${referenceId}`;
     return jsonSchema;
   }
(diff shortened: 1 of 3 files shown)


packages/to-json-schema/src/toJsonSchema.test.ts
```diff
@@ -46,6 +46,55 @@ describe('toJsonSchema', () => {
         },
       });
     });
+
+    test('for recursive schema', () => {
+      const ul = v.object({
+        type: v.literal('ul'),
+        children: v.array(v.lazy(() => li)),
```
fabian-hiller, line 53: Changing this to `v.lazy(() => v.array(li))` still causes the error. I will investigate this now.

fabian-hiller: This `v.lazy(() => v.array(li))` cannot work because the reference of `v.array(li)` is always newly created and therefore cannot be associated with any reference in our reference map. We need to document that the return type of the getter must be a static reference for recursive schemas to work.

gcornut: Oh yeah I forgot that this was the reason why I explicitly required schemas wrapped in a lazy to be present in the definitions in @gcornut/valibot-json-schema

gcornut: Another solution could be to memoize unwrapping lazy schemas 🤔

packages/to-json-schema/src/convertSchema.ts
```diff
@@ -167,263 +167,270 @@ export function convertSchema(
         jsonSchema = convertAction(jsonSchema, valibotPipeItem, config);
       }
     }
+  } else {
```
fabian-hiller, line 170: Did you remove the `return` statement for a specific reason and replace it with `else`?

fabian-hiller: I was able to fix the bug without this change and the new code at the end.

fabian-hiller:
Thanks again for your contribution! I fixed both bugs based on your PR, but with a few changes in [this commit](https://github.com/fabian-hiller/valibot/commit/03a5f5816f3fede15ed007090574b7c40dfd0aed). I have added you as the co-author.

PR #843, opened 2024-09-20
improve `minValue` & `maxValue` actions and add missing test cases

EltonLobo07:
Both of the actions will add an issue when `dataset.value` is `NaN`.


fabian-hiller:
Thank you! I will try to review and merge it on the weekend.

library/src/actions/maxValue/maxValue.test.ts
```diff
@@ -52,11 +51,86 @@ describe('maxValue', () => {
       expectNoActionIssue(maxValue(10n), [-9999n, 0n, 10n]);
     });
 
+    test('for valid non-bigints', () => {
+      expectNoActionIssue(maxValue(123n), [
```
fabian-hiller, line 55: Instead of `123n`, `0n` and `-21n` I would test `10n`, `1n` and `0n` like we did it in `notValue.test.ts`. With `10n` we test general values and with `1n` and `0n` we test boolean-like values.

> The following code is just an example. Feel free to improve it if you have additional ideas. But do not overcomplicate the tests. 😁

```ts
expectNoActionIssue(maxValue(10n), [
  -10,
  0,
  10,
  -10.0,
  0.0,
  10.0,
  '-10',
  '0',
  '10',
  // Seems not to work with bigint
  // '-10.0',
  // '-0.0',
  // '10.0',
  '',
  ' ',
  new Date(-10),
  new Date(0),
  new Date(10),
  true,
  false,
]);
expectNoActionIssue(maxValue(1n), [
  0,
  0.0,
  '0',
  ' 0 ',
  '',
  ' ',
  true,
  false,
  new Date(0),
]);
expectNoActionIssue(maxValue(0n), [
  -1,
  -1.0,
  '-1',
  ' -1 ',
  false,
  new Date(-1),
]);
```

EltonLobo07: Oh, okay. Looks like I tested with different priorities (test with zero, one random positive and negative number). I'll update the tests using your recommend approach by tomorrow (under 24 hours).

fabian-hiller: Don't stress. In the end, these tests are not too important, but nice to have.

EltonLobo07: Updated. You can re-review anytime.

PR #846, opened 2024-09-23
Add maxGraphemeCount and minGraphemeCount validators

remonke:
I skipped the `graphemeCount` and `notGraphemeCount` actions because counting graphemes seems most useful for validating user-generated content (e.g., enforcing maximum/minimum post lengths). I believe these actions would have limited use cases.

I think it would make sense to add other `Intl.Segmenter`-based validators, such as `maxWordCount`. However, this would require a `locales` parameter for non-English speakers (counting graphemes seems to work with any locale). Currently, there are no actions that take three parameters, so I'm unsure about the implementation. I think the best approach would be to use a union type for the parameter, allowing both `number` and `{ count: number; locales: Intl.LocalesArgument; }`.

Let me know what you think about this.


fabian-hiller:
Thank you for implementing these actions. For completeness, I prefer to add `graphemeCount` and `notGraphemeCount` as well, even though they are probably rarely used. What do you think about renaming these actions to `minGraphemes` and `maxGraphemes` to make the names shorter and better match them with our other actions?

Word count actions are a great idea! Feel free to implement it after this PR is merged. Since `Intl.Segmenter` takes the language as its first argument, I would follow this approach in our API design as well. What do you think?

```ts
const Action = v.minWords('fr', 100, 'Error message');
```

remonke:
Okay, when it comes to the name, I'm not a huge fan of the name `graphemes`. It is not as explicit as `graphemeCount`, but since it requires a number argument, its purpose would likely be clear, so I'm fine with using it.

About the `minWords` action having 3 arguments, with `locales` being the first one, I think it's a clever solution. This allows the action to have 4 function signatures, so people who just want to use the default locale can simply skip the argument:

```ts
function minWords(requirement: number)
function minWords(requirement: number, message: string)
function minWords(locales: Intl.LocalesArgument, requirement: number)
function minWords(locales: Intl.LocalesArgument, requirement: number, message: string)
```

It could also be adopted for `graphemes`, but as I said, the locales parameter even doesn't seem to do anything. I think it's reasonable to skip it for now, as extra function signatures can always be added in the future without breaking existing code.

fabian-hiller:
Sounds good to me! I will try to review and merge this PR on the weekend or next week.

fabian-hiller:
Thank you so much! I am busy this weekend and not sure if I will find some time, but I will try to review it next week.

remonke:
Sorry, I messed something up with git while trying to rebase.

Issue #784, opened 2024-08-15
Add length validators based on the `Intl.Segmenter` API


remonke:
Actions `minLength` and `maxLength` use `String.prototype.length`, which is not very reliable as it relies on the number of character codes. This approach is not ideal for checking the number of characters the way humans perceive it, especially with emojis. For example, most emojis have a length of 2 (like 🙃), but some have a length of 7 (like 🧑🏻‍💻).

I suggest adding new actions like `minGraphemeCount` or `maxGraphemeCount`, which would use the `Intl.Segmenter` API instead of `String.prototype.length`. This would be particularly useful when dealing with user-generated content. As of April 2024, [the API is supported in all major browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter).

The code would look something like this:

```ts
import * as v from 'valibot';

const PostSchema = v.object({
  title: v.pipe(v.string(), v.maxGraphemeCount(300, /* optional language parameter */ 'en')),
});

fabian-hiller:
Thank you for your contribution. We have already discussed this in PR https://github.com/fabian-hiller/valibot/pull/666#issuecomment-2227393849. Feel free to create a PR and copy the source code of `length`, `notLength`, `minLength` and `maxLength` and implement `graphemes`, `notGraphemes`, `minGraphemes` and `minGraphemes` based on it.

2-NOW:
Fixes: #853

fabian-hiller:
This has been implemented and will be available in the next release.

PR #853, opened 2024-09-28
Add `*graphemes` validation actions.

remonke:


2-NOW:
Closes: #784

fabian-hiller:
Thank you! I will try to review and merge this PR this or next week.

fabian-hiller:
I started to work on this PR and will probably merge it tomorrow 🙌

fabian-hiller:
I also started to implement the words actions.

fabian-hiller:
v1.0.0-beta.1 with the grapheme actions is available

nikeee:
I hate to be _that_ guy but the naming might not be correct. A grapheme can be something like the dots over an ö. Composite characters like ö[^1] or emojis are actually grapheme **clusters** (something that is perceived as a single character by the user).

Also note that what counts as a grapheme cluster is dependent on the device + locale settings where the code runs and that it can change over time. 
For example, some browsers might render 🐱‍👤 as a single grapheme cluster (Edge on Windows 10+) but as two different grapheme clusters on other systems[^2] (like the device I'm typing this on). This may yield unexpected behavior due to different definitions of grapheme clusters on participating systems.

[^1]: In that case, dependent on the used Unicode normalization.
[^2]: See: https://emojipedia.org/ninja-cat

fabian-hiller:
Thanks for your feedback. I couldn't find a case where the locale actually made a difference, but I'm happy to improve the implementation if people actually run into problems.

PR #854, opened 2024-09-28
fix(to-json-schema): Error in JSR

nakasyou:
When I imported from `@valibot/to-json-schema` using JSR and Deno, I received an error.

スクリーンショット 2024-09-28 185110
Code:
```ts
import { toJsonSchema } from '@valibot/to-json-schema'
import { schema } from '../schema/schema.ts'

export const makeSchema = async () => {
  const jsonSchema = toJsonSchema(schema)
  
  await Deno.writeTextFile('schema.json', JSON.stringify(jsonSchema, null, 2))
}

if (import.meta.main) {
  await makeSchema()
}
```
The PR will fixes it.


fabian-hiller:
Thank you!

PR #855, opened 2024-09-28
Docs: add GQLoom to ecosystem

xcfox:
I'm excited to introduce you to GQLoom.

[GQLoom](https://gqloom.dev/) is a GraphQL weaver for TypeScript/JavaScript that weaves GraphQL Schema and resolvers using Valibot, Zod, or Yup, and supports sophisticated type inference to provide the best development experience.

## Hello World

```ts
import { resolver, query, weave } from "@gqloom/valibot"
import * as v from "valibot"

const HelloResolver = resolver({
  hello: query(v.string(), () => "world"),
})

export const schema = weave(HelloResolver)
```


fabian-hiller:
Thank you for supporting Valibot! 🙏

Issue #867, opened 2024-10-09
Type inference on findItem when using type predicates


cberthou:
For now, findItem returns the input type event when trying to narrow the typing using type predicates. For example :
```ts
import * as v from 'valibot';

type Animal = 'cat' | 'dog';

const isAnimal = (v: string): v is Animal => v === 'cat' || v === 'dog';

const schema = v.pipe(
  v.array(v.string()),
  v.findItem(isAnimal)
)

const animal = v.parse(schema, ['apple', 'cat', 'screw']);
// animal is type string | undefined
```

It would be cool to get better type narrowing. For now, you have to do this, which feels weird given that findItem exists :
```ts
const schema = v.pipe(
  v.array(v.string()),
  v.transform((array) => array.find(isAnimal)),
)
```

fabian-hiller:
Thank you for creating this issue and your PR! I am very interested in improving it. I will try to have a closer look in about two weeks.

fabian-hiller:
This is now implemented and will be available with the next version.

fabian-hiller:
v1.0.0-beta.4 is available

PR #868, opened 2024-10-09
improve type inference when using findItem with type predicates

cberthou:
#867 

Not a big fan of duplicating FindItemAction to allow for a type with a predicate, but I didn't want to break retrocompatibility, nor create complex types that would degrade readability or maintenability.


fabian-hiller:
It is ok to break things since we are not v1 yet. How would you implement it with breaking changes?

fabian-hiller:
This is now implemented. I have added you as a co-author to my commit. Thank you for this PR!

PR #870, opened 2024-10-10
Add base of uniqueItems action

nix6839:


fabian-hiller:
Without much thought I would use `dataset.value.length !== new Set(dataset.value).size` to check for duplicates and return a single issue. But it is true that this is not an easy decision. We basically have to decide whether we want one issue for the whole array or one issue for each duplicate item. We also have to decide whether to check for shallow equality or deep equality. For example, shallow equality will not work with `object` because the library parses the input and therefore copies the data into a a new object when the schema is executed. What do you think is best?

nix6839:
> We also have to decide whether to check for shallow equality or deep equality. For example, shallow equality will not work with object because the library parses the input and therefore copies the data into a a new object when the schema is executed.

I think it would be best to use deep equality to ensure it operates the same as JSON schema. However, I also believe there will be users who prefer shallow equality. To accommodate those users, how about using deep equality by default but allowing a customizable function in the form of `(a, b) => boolean`?
I am a bit concerned about the potential performance issues with deep equality. What are your thoughts on this?

> We basically have to decide whether we want one issue for the whole array or one issue for each duplicate item.

As for the issue handling, I think we should first agree on and implement the functionality above, and then look into how we can handle the issue format afterwards.

fabian-hiller:
How about we provide a `uniqueItems` action for a small bundle size and better performance and a `deepUniqueItems` action that uses a nested loop that deeply compares objects, arrays and dates?

nix6839:
I think that's a good idea. I'll try to implement `uniqueItems` first tomorrow.

fabian-hiller:
Would you return an issue for every duplicated item or just one for the entire array?

nix6839:
Since `checkItems` (which uses a similar naming pattern with `-Items` instead of `-Item`) operates by returning an issue for every duplicated item, how about making `uniqueItems` behave the same way for consistency?

nix6839:
I've committed it for now, but feel free to share any other opinions. If it's finalized, I'll start working on updating the documentation.

fabian-hiller:
Thank you very much! I will try to give you feedback at the end of next week as I have to focus on other things in the next few days.

fabian-hiller:
I just want to let you know that I am focusing on our v1 release first before reviewing this PR.

PR #874, opened 2024-10-16
Update index.mdx

jojojojojoj5564656465465:
it' didn't works with your command i think it must be the update of deno


fabian-hiller:
Thank you!

PR #877, opened 2024-10-17
Add "dependent default" example to `v.optional`

fartinmartin:
As I understand, this is the [conventional way](https://discord.com/channels/1252985447273992222/1295957546644930560/) to apply default values to optional props. I thought we could include it in the docs (until if/when a more concise API is added)!


fabian-hiller:
I think the API reference of `optional` is probably the wrong location but we could add such an example to the optionals guide: https://valibot.dev/guides/optionals/

fartinmartin:
Ah, totally—that makes sense to me! I've updated branch, let me know if I should tweak it further!

fabian-hiller:
Thank you! I will try to review it on the weekend.

Issue #894, opened 2024-10-23
improvement : add domain name validation


neyzth:
Hello, as a IT guy i have to work sometimes with domain name, not url. Something to check if the input like :
```txt
corebackbone.par.franceix.net
```
or
```txt
vl223.par-itx5-dist-2.cdn77.com
```
have a valid domain name syntax.

Is this possible to add a action for validating this kind of inputs please ?

fabian-hiller:
Yes, we can add a `domain` action. Is there any regex you would recommend? Would you be interested in creating a PR?

neyzth:
hi, this kind of regex i guess :
```REGEX
^((?!-)[A-Za-z0-9-]{1,63}(?<!-).)+[A-Za-z]{2,}$
```

fabian-hiller:
Are you interested in creating a PR that implements the actions and adds tests? This action is very similar to other actions like `email` or `decimal`. You could simply copy one of them and use it as a template.

JacKyDev:
If the questioner doesn’t want to, I’d do it. However, I’d only be able to manage it next week.

I wanted to dive deeper into the valibot code anyway. But only if you agree, as the questioner hasn’t responded yet.

paksa92:
I implemented it: #907

fabian-hiller:
I am focusing on our v1 release at the moment. In the meantime, I recommend using the regex with our `regex` action as a workaround.

JacKyDev:
@fabian-hiller : A quick general question about PRs since I’m looking at one right now. Should the `.mdx` files for the website, like `menu.md` and `properties.ts`, also be updated? Otherwise, it would mean the features are there but not visible anywhere (like in the docs).

fabian-hiller:
Yes, but I can also take care of the details because I know it can be hard to understand all the connections.

tats-u:
![image](https://github.com/user-attachments/assets/ae55e05e-b5bc-482b-8440-380570126849)
https://makenowjust-labs.github.io/recheck/playground/

```js
/^((?!-)[A-Za-z0-9-]{1,63}(?<!-).)+[A-Za-z]{2,}$/.test('0' + '0-'.repeat(30) + '\x00')
```

took about 10s to generate `false`.

paksa92:
@tats-u we agreed on this final regex pattern: `/^(?=.{1,253}$)(?!-)([a-z0-9-]{1,63}(?<!-)\.)+[a-z]{2,63}$/iu` in #907.

[...]

PR #907, opened 2024-11-01
feat: add domain name action

paksa92:
Because I love valibot so much I wanted to do something in return for all the headache she saved me from so here's a PR that adds domain name validation as requested by #894.

### DIFF
--- a/library/src/actions/domainName/domainName.test-d.ts
+++ b/library/src/actions/domainName/domainName.test-d.ts
@@ -1,0 +1,47 @@
+import { describe, expectTypeOf, test } from 'vitest';
+import type { InferInput, InferIssue, InferOutput } from '../../types/index.ts';
+import {
+  domainName,
+  type DomainNameAction,
+  type DomainNameIssue,
+} from './domainName.ts';
+
+describe('domainName', () => {
+  describe('should return action object', () => {
+    test('with undefined message', () => {
+      type Action = DomainNameAction<string, undefined>;
+      expectTypeOf(domainName()).toEqualTypeOf<Action>();
+      expectTypeOf(domainName(undefined)).toEqualTypeOf<Action>();
+    });
+
+    test('with string message', () => {
+      expectTypeOf(domainName('message')).toEqualTypeOf<
+        DomainNameAction<string, 'message'>
+      >();
+    });
+
+    test('with function message', () => {
+      expectTypeOf(domainName(() => 'message')).toEqualTypeOf<
+        DomainNameAction<string, () => string>
+      >();
+    });
+  });
+
+  describe('should infer correct types', () => {
+    type Action = DomainNameAction<string, undefined>;
+
+    test('of input', () => {
+      expectTypeOf<InferInput<Action>>().toEqualTypeOf<string>();
+    });
+
+    test('of output', () => {
+      expectTypeOf<InferOutput<Action>>().toEqualTypeOf<string>();
+    });
+
+    test('of issue', () => {
+      expectTypeOf<InferIssue<Action>>().toEqualTypeOf<
+        DomainNameIssue<string>
+      >();
+    });
+  });
+});
--- a/library/src/actions/domainName/domainName.test.ts
+++ b/library/src/actions/domainName/domainName.test.ts
@@ -1,0 +1,193 @@
+import { describe, expect, test } from 'vitest';
+import { DOMAIN_NAME_REGEX } from '../../regex.ts';
+import type { StringIssue } from '../../schemas/index.ts';
+import { expectActionIssue, expectNoActionIssue } from '../../vitest/index.ts';
+import {
+  domainName,
+  type DomainNameAction,
+  type DomainNameIssue,
+} from './domainName.ts';
+
+describe('domainName', () => {
+  describe('should return action object', () => {
+    const baseAction: Omit<DomainNameAction<string, never>, 'message'> = {
+      kind: 'validation',
+      type: 'domainName',
+      reference: domainName,
+      expects: null,
+      requirement: DOMAIN_NAME_REGEX,
+      async: false,
+      '~validate': expect.any(Function),
+    };
+
+    test('with undefined message', () => {
+      const action: DomainNameAction<string, undefined> = {
+        ...baseAction,
+        message: undefined,
+      };
+      expect(domainName()).toStrictEqual(action);
+      expect(domainName(undefined)).toStrictEqual(action);
+    });
+
+    test('with string message', () => {
+      expect(domainName('message')).toStrictEqual({
+        ...baseAction,
+        message: 'message',
+      } satisfies DomainNameAction<string, string>);
+    });
+
+    test('with function message', () => {
+      const message = () => 'message';
+      expect(domainName(message)).toStrictEqual({
+        ...baseAction,
+        message,
+      } satisfies DomainNameAction<string, typeof message>);
+    });
+  });
+
+  describe('should return dataset without issues', () => {
+    const action = domainName();
+
+    // General tests
+
+    test('for untyped inputs', () => {
+      const issues: [StringIssue] = [
+        {
+          kind: 'schema',
+          type: 'string',
+          input: null,
+          expected: 'string',
+          received: 'null',
+          message: 'message',
+        },
+      ];
+      expect(
+        action['~validate']({ typed: false, value: null, issues }, {})
+      ).toStrictEqual({
+        typed: false,
+        value: null,
+        issues,
+      });
+    });
+
+    test("should match 'example.com' - Standard domain", () => {
+      expectNoActionIssue(action, ['example.com']);
+    });
+
+    test("should match 'example.org' - Another common TLD", () => {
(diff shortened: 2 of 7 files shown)


paksa92:
It still needs docs for the website

fabian-hiller:
Thank you so much! 🙏 I am focusing on our v1 release first before reviewing this PR. In the meantime, I recommend using the regex with our `regex` action as a workaround.

JacKyDev:
Too bad, I wanted to do it right away :( But isn't the following regex better suited for this purpose?

`/^(?=.{1,253}$)(?![-_])(?!.+[-_]\.)([\w-]{1,63}\.)+[a-z]{2,63}$/iu`

It better aligns with [RFC 1035](https://www.ietf.org/rfc/rfc1035.txt), doesn't it? This way, it also includes the restrictions that apply to a domain and its subdomains.

fabian-hiller:
I haven't checked the regex yet. Almost every regex of Valibot is handmade. I will check the regex before merging.

JacKyDev:
This regex is used to validate **domain names**. Exactly that was my intention. I looked more closely into RFC 1035 and created this regex with the goal of adhering to the specifications as much as possible so that tools like Netscaler or DNS processes will consistently accept these domains. There is a lot of guidance in RFC 1035, especially concerning octets and the conditions for when and how certain characters are allowed.

Here’s a detailed breakdown of how the regex works:

`/^(?=.{1,253}$)(?![-])(?!.+[-].)([\w-]{1,63}.)+[a-z]{2,63}$/iu`

### Explanation of each part:
1. **`^(?=.{1,253}$)`**: This part checks that the total length of the input is between 1 and 253 characters, which is the maximum allowable length for domains.

2. **`(?![-_])`**: This ensures that the domain name does not begin with a hyphen (`-`) or an underscore (`_`).

3. **`(?!.+[-_]\.)`**: This negative lookahead ensures there is no combination of a hyphen or underscore immediately before a period (`.`). This prevents invalid patterns like `domain-.com`.

4. **`([\w-]{1,63}\.)+`**: This section validates each label (segment) of the domain name:
   - `[\w-]{1,63}` allows between 1 and 63 alphanumeric characters or hyphens (but no `_`, which is not allowed in domain names).
   - `\.` at the end ensures each label ends with a period.
   - The `+` at the end allows one or more repetitions of this pattern, supporting subdomains.

5. **`[a-z]{2,63}$`**: This section at the end checks if the TLD (top-level domain) contains only lowercase letters and is between 2 and 63 characters long, as is typical for TLDs (e.g., `.com`, `.online`).

6. **`iu`**: The flags `i` and `u` represent:
   - `i` (ignore case): Case-insensitive, so `.COM` and `.com` are both accepted as valid.
   - `u` (unicode): Enables Unicode support for character classes like `\w`.

### Examples of valid domains:
- `example.com`
- `sub.domain.co.uk`
- `my-site123.org`

### Examples of invalid domains:
- `-example.com` (begins with a hyphen)
- `example-.com` (hyphen before a period)
- `toolongtoolongtoolongtoolongtoolongtoolongtoolong.com` (label is longer than 63 characters)

This regex is designed to follow DNS standards closely, enabling reliable domain validation across different DNS systems and tools.

This way, I can still help a bit :)

paksa92:
Sorry @JacKyDev I didn't mean to steal your thunder. I just had some spare time left and since it been a few days since you responded I figured I'd step in instead.

I will look into the regex you proposed when I get the chance today!

JacKyDev:
@paksa92 All good, I'll survive :D  
But since I already put together the regex yesterday based on what I’d read, I thought it might be a way to help out. In the end, it saves me time too :)

JacKyDev:
I'm not sure if this is common practice at Valibot, so I’ll just mention it. I would assume that the documentation could also be updated so that `domainName` not only exists but can also be displayed in the documentation in the future. I’ve also asked Fabian about this in the issue.

Additionally, I would suggest that something similar to `email` should also be included in the necessary documentation files.

If I haven't overlooked anything, those would be:
- `index.mdx` (website/src/routes/api/(actions)/domainName & website/src/routes/api/(async)/customAsync & website/src/routes/api/(async)/fallbackAsync & website/src/routes/api/(async)/intersectAsync & website/src/routes/api/(async)/lazyAsync & website/src/routes/api/(async)/unionAsync)
- `properties.ts` (website/src/routes/api/(actions)/domainName)

Also, the `menu.md` in `routes/api` to find `domainName` in the docs.

But I'm not sure if this is common practice at Valibot or if I might be overthinking it. Perhaps you should wait for Fabian to respond, as I also asked him this question here: https://github.com/fabian-hiller/valibot/issues/894.

paksa92:
@JacKyDev yeah, I already mentioned that it is missing in the PR, but I will add it.

`CONTRIBUTING.md` does not mention a "definition of done" for new features, maybe it should? @fabian-hiller

fabian-hiller:
I suppose devs rarely look at the `CONTRIBUTING.md`. So it is not too important. But feel free to create a PR.

paksa92:
> /^(?=.{1,253}$)(?![-])(?!.+[-].)([\w-]{1,63}.)+[a-z]{2,63}$/iu

@JacKyDev I'm currently testing this regex and I think it's wrong. First off, TLD's may not exceed a length of 6. Changing the regex to:
```
/^(?=.{1,253}$)(?![-])(?!.+[-].)([\w-]{1,63}.)+[a-z]{2,6}$/iu
```
But, this regex still allows TLD's exceeding a length of 6. I suspect one of the negative lookaheads `(?![-])(?!.+[-].)` causes the problem, but I'm not sure. I can't seem to fix it.

I tweaked the original regex in my PR to account for the max length of a domain name, which now should cover all the cases you mentioned:
```
/^(?=.{1,253}$)(?!-)([a-z0-9-]{1,63}(?<!-)\.)+[a-z]{2,6}$/iu
```

Please test it and let me know if you find any issues.

paksa92:
You're right @JacKyDev, there is no limit of 6 characters on the TLD. My assumption was based on some stackoverflow posts I've seen.

I think the regex is all set now! :)

tats-u:
`^(?=.{1,253}$)` references the number of UTF-32 code points if `u` flag is added. `length` property of the input string references UTF-16.
I don't know which is better for domain validation.

JacKyDev:
@tats-u : I understand what you mean, but when I test this regex with emojis or other Unicode characters, it consistently flags them as invalid. This aligns with what I expected, since domains technically only allow ASCII characters. However, I can't replicate what you're describing in a unit test—here, all tests fail as soon as characters outside of `a-z, 0-9, or -` are included. So, I'm not entirely sure what you're aiming for.

From a technical standpoint, this is exactly what I'd expect, as any characters outside the ASCII range should be represented in Punycode to be valid in a domain. Converting to Punycode means that an emoji, for example, doesn't count as a single character, but as multiple ASCII characters, which would also increase the string length—especially since Punycode includes a prefix.

Do you have a test example we can try? Currently, emojis and special characters like "ü" are rejected, just as intended.

tats-u:
> since domains technically only allow ASCII characters. However, I can't replicate what you're describing in a unit test—here, all tests fail as soon as characters outside of a-z, 0-9, or - are included.


If so we don't have to care about surrogate pairs. You designed this validator on the premise that we need to make Unicode domains punycoded in advance. 
I am reminded that users care about bundle size. The current regex is small compared to code using `length` property and sufficient.

JacKyDev:
I think we’re in a good place by strictly requiring specific values... This way, the domain itself is validated very precisely. And we wouldn’t be offering any special exceptions. Separately, one could consider adding a `toPunyCode` action in the future to initially standardize a value, making it usable for everyone. This approach, I think, creates more potential.

I wouldn’t really be against the additional check, because you’re right—it’s not much more code.

I still lack experience with the process here, as a regex action might feel messy if it includes code checks. But a length Check ist really small... :)
 @fabian-hiller could probably answer that better, I think.

tats-u:
> `toPunyCode`

I agree with it. There are 2 strategies:

1. Test the return value of `toPunyCode` by the regex, then pass the return value of `toPunyCode` to the return value of `parse`
2. Test the return value of `toPunyCode` by the regex, then pass the input to the return value of `parse` as is

Users pass the punycoded domains to other libraries in 1. while they utilize unpunycoded Unicode domains as are.

fabian-hiller:
I am very sorry for not getting back to you on this PR. I will try to give you feedback in the next few weeks. I plan to do the final review and merge this PR after Valibot v1 is available.

BerkliumBirb:
@JacKyDev 
I would argue that we should really have separate `toDomainPunyCode` and `toPunyCode` because:
1. domain is not a puny coded string, _domain parts_ are
2. `xn--` can't be default behavior for `toPunyCode` because this prefix is an [implementation detail of IDN](https://en.wikipedia.org/wiki/Punycode#ACE_prefix_for_internationalized_domain_names), not puny code algorithm.

muningis:
By any chance this is based on https://datatracker.ietf.org/doc/html/rfc1123#section-2.1 ?

PR #910, opened 2024-11-03
add `slug` action and its tests

EltonLobo07:


fabian-hiller:
v1.0.0-rc.1 is available

PR #911, opened 2024-11-04
feat: add btcAddress action

shuaixr:
Perform checksum validation for a btc address

## Summary by CodeRabbit

* **New Features**
  * Added Bitcoin address validation actions for legacy Base58Check and native SegWit Bech32/Bech32m, plus Base58Check-only and Bech32/Bech32m-only variants.
  * Added support for custom validation messages with strong TypeScript inference.
  * Exposed the new actions, issue/action contracts, and related validation utilities via the public API.
* **Documentation**
  * Added API reference pages for the new actions and their generic types, including usage examples and validation scope/limitations.
* **Tests**
  * Added runtime and compile-time tests covering valid/invalid addresses, issue generation, and message/type inference.


shuaixr:
> Is there any way we can simplify the bitcoin address check with a simple regular expression and checksum function?
sure, I'll add a regular expression, first using regex then the checksum.

fabian-hiller:
> sure, I'll add a regular expression, first using regex then the checksum.

Is it possible to simplify the implementation in this way and reduce the bundle size without losing functionality? The current implementation contains a lot of functions and seems quite complicated.

shuaixr:
> Is it possible to simplify the implementation in this way and reduce the bundle size without losing relevant functionality? The current implementation contains a lot of functions and seems quite complicated.

I can use crypto.subtle.digest instead of a library/src/actions/btcAddress/sha256-uint8array.ts file. It's supported in Node 15+, Bun, Deno, and browsers. Other functions, in my opinion, are only necessary for checksum calculations

fabian-hiller:
Thank you for your research. I want to let you know that I am focusing on our v1 release first before reviewing this PR.

yslpn:
Hi @shuaixr Thanks for your input. I have a few thoughts.

We have a use case for splitting into three actions: ip/ipv4/ipv6. Do you think we could do the same here? For example, I see that we could split into btcAddressBase58, btcAddressBech32, and then combine btcAddress. This is especially important, as btcAddressBech32 users won't need the heavy sha256 dependency.

What about other cryptocurrencies in the future? Similar to credit card validation, move _base58Check and _bech32 out and then create thin wrappers that reuse the logic. Let's move the logic directly to library/src/utils/.

Another question is whether we need the first dependency in the project or not. sha256-uint8array is a library that was last updated six years ago. Moreover, it hasn't undergone a security audit. I believe the most modern and secure alternative is https://github.com/paulmillr/noble-hashes 

It would also be great if you could write some website documentation for your check, so users know exactly what it checks and what the limitations are.

shuaixr:
Hi @yslpn , sure. It’s been quite a while, so I’ll take another look and address this.

shuaixr:
I introduced [@noble/hashes](https://github.com/paulmillr/noble-hashes) as Valibot first runtime dependency for the Base58Check checksum

Browser Bundle size check:

| Action | Raw | Gzip | 
| --- | ---: | ---: | 
| `btcAddressBech32` | 3.88 KB | 1.57 KB |
| `btcAddressBase58` | 9.16 KB | 3.82 KB |
| `btcAddress` | 11.53 KB | 4.59 KB |

yslpn:
I don't think we can honestly verify a BTC address without using the sha256 function. This is quite dangerous, since users' funds and money depend on it.

We have a choice: add a dependency on noble with a synchronous function for working with sha256, or make the entire verification asynchronous using the web platform's browser API https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest

What do you think @fabian-hiller?

PR #912, opened 2024-11-06
feat: add rfc5322 validator

### LINKED ISSUE #204: email validation: provide spec validation
The currently used email regex does not match emails according to the spec, which means emails that browsers accept, will be rejected by valibot ([by design](https://github.com/fabian-hiller/valibot/pull/180#issuecomment-1751597772))

So we want to use this issue to find out if others would be interested in using a regexp that validates more emails: https://github.com/fabian-hiller/valibot/pull/180#issuecomment-1751630504

hendrikheil:
This PR implements a new action for RFC 5322 based email address validation. A few things are of particular importance.
1. The regex used in this PR was taken from the HTML living standard and has a few willful violations of RFC 5322
2. It is impossible and unrealistic to parse a fully RFC compliant email address with a regular expression, as the spec allows for comments, which are made of balanced potentially nested parenthesis pairs. However, email clients such as Gmail don't accept these addresses either
3. If the willful violations are not acceptable, it is likely that parsing an address without a regex is a better choice.

I opted to reuse the types, reference and `type` of the default `email` action so that there is just a single case for i18n messages. I'm not sure if that approach is ideal, but I'd be happy to change it.
I'll add documentation after this question is answered as not to duplicate the work.

Fix #204


fabian-hiller:
Thank you for creating this PR! I would not reuse `email` and create a brand new action instead, as I think this makes the library simpler. We have never reused an action in this way.

hendrikheil:
@fabian-hiller Sorry for the long delay. I just updated things to not reuse anything

fabian-hiller:
Thank you! I will review this PR after our v1 RC or stable v1 is out.

fabian-hiller:
Thank you for your contribution!

fabian-hiller:
v1.0.0-beta.10 is available

PR #930, opened 2024-11-16
feat: Auto-scroll playground's logs container to the latest log message

### LINKED ISSUE #920: Valibot playground, scroll down on new log messages
**Feature Request**

In the [Valibot playground](https://valibot.dev/playground/), scroll down automatically, when a new log message arrives.
You always have to scroll down manually to see latest log messages, which is annoying.
image

DrWarpMan:
Closes #920


fabian-hiller:
I think there is a native `scrollIntoView` method we could use: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

```ts
logsElement.scrollIntoView({
  block: 'nearest',
  inline: 'center',
});
```

fabian-hiller:
Thanks for this PR! I changed the implementation a bit. Instead of scrolling to the end of the console box, I changed the implementation so that it scrolls to the beginning of the newest logs. I did this because for very long log messages the current implementation scrolled to the end, which led to a bad DX because it forced the user to scroll up to see the entire log message. I also added a scroll animation and used the browser's native `.scrollIntoView` method. I will merge this PR now. Feel free to review my changes and give me feedback.

Issue #933, opened 2024-11-18
JSON type


jhnns:
Hello, thanks for this great library 👋

I was wondering if you'd be interested in a documentation for a JSON serializable type. That's the type that you can safely serialize and de-serialize via `JSON.parse(JSON.stringify(thing))`. [Zod mentions it in their documentation](https://zod.dev/?id=json-type).

My suggestion (based on Zod's implementation) would be:

```ts
const JsonLiteral = v.union([v.string(), v.number(), v.boolean(), v.null()]);
type JsonLiteral = v.InferOutput<typeof JsonLiteral>;

type Json = JsonLiteral | { [key: string]: Json } | Array<Json>;
const Json = v.lazy(
  (): v.UnionSchema<any, undefined> =>
    v.union([JsonLiteral, v.array(Json), v.record(v.string(), Json)]),
);
```

Maybe this could even be included in valibot so that people don't need to copy it.

fabian-hiller:
Where would you place to code in the docs?

Here is a JSON schema I wrote a few months ago:

```ts
type Json = string | number | boolean | null | { [key: string]: Json } | Json[];
type LiteralIssue = StringIssue | NumberIssue | BooleanIssue | NullIssue;
type JsonIssue = LiteralIssue | ArrayIssue | RecordIssue;
type JsonSchema = BaseSchema<Json, Json, UnionIssue<JsonIssue> | JsonIssue>;
export const jsonSchema = v.lazy(() =>
	v.union([v.string(), v.number(), v.boolean(), v.null_(), v.array(jsonSchema), v.record(v.string(), jsonSchema)])
) as JsonSchema;
```

jhnns:
I think I'd expect it under Guides > Advanced.
What do you think about implementing it as valibot action?

fabian-hiller:
We could at it to https://valibot.dev/guides/other/. Feel free to create a PR.

I see and DX advantage of adding a `json` schema. But I have to think about it because it is a very special kind of schema compared to our other schemas that just mimic TS type functionality. Also, we have had ideas in the past to add something like a `json` action to validate JSON strings. We need to make sure the naming makes sense. I think this is something I would prefer to investigate after our v1 release.

jhnns:
[I've created a PR](https://github.com/fabian-hiller/valibot/pull/936). [valibot.dev/guides/other](https://valibot.dev/guides/other/) is the perfect spot, right below `lazy` :)

[...]

PR #936, opened 2024-11-21
Add JSON type schema

jhnns:
Adds documentation for a JSON type schema as discussed in #933

I changed the code snippet to use the second naming convention. Not sure if there's a better alternative for `any` in `v.UnionSchema<any, undefined>`, but I also don't think it's a big issue as the type is not used for inference.


fabian-hiller:
I don't have access to add commits to this PR. Therefore, I will close it and add this change myself to the docs. I will add you as a co-author to my commit. Thank you for your contribution!

jhnns:
Ahh sorry, I forgot to allow it 😞

PR #940, opened 2024-11-22
improve `Flatten errors` section's (issues guide) example

EltonLobo07:
The example under the `Flatten errors` section uses a `nested` key to define the object schema. This might confuse the readers, as the result returned by the `flatten` function might also use `nested` as a key to indicate nested issues.

Before:
image

After:
image


fabian-hiller:
Another option would be to rename the keys to `nested1` and `nested2`. What do you think?

EltonLobo07:
I am not sure if I understand. We can't rename the key returned by the `flatten` method.
image
What we can do is rename the key of the example input object used (`nested` -> `nestedObject` or something else) to make it easier to understand the relation between the input and output. So we have the flexibility to change one object key. That means `nested1` and `nested2` won't make sense.

To provide more context: While reading the section, initially I was confused.
Input:
image
Output:
image
Why is `key` inside `nested`? In the input the outer `key` is not inside `nested` but in the output it is. But after a while I figured it was one of the standard keys returned by the `flatten` method.

I think we can improve this using a different name that is not one of the standard keys returned by the `flatten` method. What do you think?

EltonLobo07:
Renaming all of the keys is even better! Will update this PR shortly.

PR #941, opened 2024-11-25
Fix issues with enum options filtering logic in enum_ function

ikeyan:
The current implementation of the enum_ function in Valibot filters enum options using the following code:

https://github.com/fabian-hiller/valibot/blob/8ab047f44650ab0fbece0ffac4c2972a0a10b836/library/src/schemas/enum/enum.ts#L92-L98

The purpose of this filtering logic is to remove the reverse mapping entries that TypeScript adds when transpiling enums. Specifically, when an enum member is defined as Name = value and value is a number, TypeScript generates a reverse mapping in the form `Enum[value] = 'Name'`. The key of this reverse mapping is stringified by [Number.prototype.toString](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-numeric-types-number-tostring) with radix=10 via [ToPropertyKey](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-topropertykey) as defined by the ECMAScript specification.

There are several issues with using `+key` and `isNaN` in the filtering logic:

* **String Conversion**: The +key operation follows [the StringToNumber abstract operation as defined in the ECMAScript specification](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-stringtonumber). Since `enum_` filters out any number, this operation has several implications:
  * It trims leading and/or trailing whitespaces or line terminators.
  * If the string is empty after trimming, it returns 0.
  * Strings such as `Infinity` or `-Infinity` are interpreted as their respective number representations.
  * It interprets prefixes like `0b`, `0o`, or `0x` as binary, octal, or hexadecimal literals.
  * Decimal variations such as `0.1`, `.1`, and `0`. are all valid and yield the respective numeric values.
  * It also allows leading zeros.
  * It accepts the `+` symbol as a valid sign.
* **NaN Values in Enum**: If the enum contains `NaN`, as in the following example:
  ```typescript
  enum N { X = NaN }
  // equivalent to js: const N = { X: NaN, NaN: 'X' }
  ```

  The `isNaN` check will not work as expected since `isNaN(key)` will return true when key is 'NaN', causing the 'NaN' key to be retained rather than filtered out.

To correctly filter out these reverse mappings, I made the filtering logic more precise. In this PR, rather than `isNaN`, the parsed value is stringified and is matched against the original key. If it matches, it is filtered out.

Additionally, the filtering logic is not reflected in the `EnumSchema` type. For example, when using `enum_({ 1: "A" })`, the `options` end up being empty, which was not obvious until runtime. Therefore I implemented type-level option filtering.


fabian-hiller:
Thanks for creating this PR! Unfortunately I have almost no time to review and merge it in the next 3 weeks, but I plan to do it before we release our stable v1.

library/src/schemas/enum/enum.ts
```diff
@@ -18,6 +18,31 @@ export interface Enum {
...
+ * This type-level function filters out numeric keys including Infinity and NaN.
+ *
+ * @example FilterEnumKeys<'foo' | '1' | 'Infinity' | 'NaN'> will be 'foo'.
+ */
+type FilterEnumKeys<K extends string> = K extends K
+  ? IsNumericString<K> extends true
+    ? never
+    : K
+  : never;
+type EnumValues<TEnum extends Enum> = TEnum[FilterEnumKeys<
```
fabian-hiller, line 42: I did not fully understand why `FilterEnumKeys` is needed. Can you provide a minimal example via our [playground](https://valibot.dev/playground/) that demonstrates that the current implementation results in a wrong type?

library/src/schemas/enum/enum.ts
```diff
@@ -94,7 +119,7 @@ export function enum_(
   message?: ErrorMessage<EnumIssue>
 ): EnumSchema<Enum, ErrorMessage<EnumIssue> | undefined> {
   const options = Object.entries(enum__)
-    .filter(([key]) => isNaN(+key))
+    .filter(([key]) => (+key).toString() !== key)
```
fabian-hiller, line 122: In my tests, your change did not solve the `v.enum({ 1: "A" })` problem. My recommendation would be as follows:

```ts
return Object.entries(enum__)
    .filter(([key, value]) => isNaN(+key) || enum__[value] === undefined)
    .map(([, value]) => value);
```

The additional `enum__[value] === undefined` check ensures that only reverse mapping entries are removed.

Regarding the `+key` issue you mentioned. Can you provide an example that results in a bug with the current implementation?

fabian-hiller (commented):
Thank you for creating this PR! I look forward to your feedback to merge this PR soon.

library/src/schemas/enum/enum.ts
ikeyan: These are my motivating examples:

https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh6BjCAO1XgGUmALAUxDI4AXkwA6PqwCuIABQBvOAElW2YK2AwAngC44ABjgBfAJQNmbDnCh9UU0vFEYxqZNj4AFZFFR9Z3fkEAGgMzRhZ2CFI+MVIIAHNZGzsHMzgAenTlVhJyABM4bTA+PQBRAA9iphg+AtY+DD4oOAopeBsmPmBGgv1GWQB9PX0TPWcVdygAeTawNoAeIr4IbDgAgWQAPhFtgdoMrJmYOfhUHgh7AvUmUik8vgN6IA
```ts
import * as v from 'valibot';

const Schema = v.enum({ Infinity: 0 });

const result = v.safeParse(Schema, 0);

console.log(result); // Invalid type: Expected never but received 0

(_: 0): v.InferOutput<typeof Schema> => _; // Output should include 0
```

https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh6BTAOwFcCBBOAb3rjgBpOAF44AOWTj6AX0YBjCC1TwAyvIAWTEMlGYAdKw4AKTgEoG9RcvhQmqNqXhiM+1MmxMACsiiomxupaOgA0RIKEFgpKqBCkTPqkEADmxnYOThZwAPTZcA7y8vaojMYA+gBc4ZFVrgCSLJ5QAPJsMGBtADwwAJ5gTBDYcEHayAB8ohNltDl5re1t+RoQjgAmcCzUcMAs8qRsq0zV9EA
```ts
import * as v from 'valibot';

enum A {
  K = NaN
}

const Schema = v.enum(A);

const result = v.safeParse(Schema, 'K');

console.log(result); // success

(_: 'K'): v.InferOutput<typeof Schema> => _; // Output should not include 'K'
```

https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh6BTAOwFcCBBOAb3riICAjADpBAXiIB5Fk0L0AvowDGEFqngBlZQAsmIZHEkYRrDgApOASgb1V6+FCao2peMZGpk2JgAVkUKhM5tp6BgA00rKENipqqBCkTCKkEADm5k4ubjZwAPR5cACSLCTkACZwMACeYEwAXHAAogAedcowTJWyGExQcBRsjkzKTMC9lQA6AEQyTDOM5gD6jeY1dRDYcNYA2oTCYoQAulaNJiU+UFJDYEMAPOtMm3Ch+sgAfEafS7T5hdcwW7wVA6CCuSrAFjKUhscpMKJyIA
```ts
import * as v from 'valibot';

enum A {
  '  1. ' = 'One'
}

const Schema = v.enum(A);

const result = v.safeParse(Schema, 'One');

console.log(result); // Invalid type: Expected never but received \"One\"

(_: (typeof A)['  1. ']): v.InferOutput<typeof Schema> => _; // Output should include 'One'
```

fabian-hiller: Thanks for the code examples. I'm very sorry, but I still don't understand why `FilterEnumKeys` is necessary.  It is true that there is currently a bug in the way we filter the enum options, which leads to wrong runtime behavior, but the type information should be correct.

library/src/schemas/enum/enum.ts
fabian-hiller: I further investigated this PR and the following code seems to fix all cases mentioned above. Should we add this code to this PR and revert the `FilterEnumKeys` changes?

```ts
Object.entries(enum__)
  .filter(([key, value]) => `${+key}` !== key || enum__[value] === undefined)
  .map(([, value]) => value);
```

fabian-hiller:
I will improve the filtering of reverse mappings, undo the changes to types and merge this PR if you have no objections.

library/src/schemas/enum/enum.ts
ikeyan: @fabian-hiller A wrong type occurs when enum_ accepts an object literal and the literal contains reverse mapping entries.

https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh6BjCAO1XgGUmALAUxDI4AXkwA6PqwCuIABQBvenGVwAWnzwAuOAAYANEpU7thdXkIGVcAIwmA8qz4X6AXwCUDGAE8wfON35BOykYMBCRcQBJVmwNYNCQgB5vXwhsf14BZAA+WjgAenzdOAAfOAAiMwhy0oqHPnL6QvFJGRxgUhgNdAhwoQBrPi9dPThUCAzA5Hiw+FQeXtIAEzgKPx1a8vrGoA
```
import * as v from 'valibot';

const Schema = v.enum({
    Zero: 0,
    0: 'Zero',
    1: 'One',
});
type SchemaOutput = v.InferOutput<typeof Schema>; // 0 | "Zero" | "One"
// v.enum filters out a key 0, so SchemaOutput should be 0 | "One"
```

library/src/schemas/enum/enum.ts
ikeyan: @fabian-hiller I refined the `enum__[value] === undefined` part. Is this better?

```
  const options = Object.entries(enum__)
    .filter(
      ([key, value]) =>
        !(
          `${+key}` === key &&
          typeof value === 'string' &&
          Object.is(enum__[value], +key)
        )
    )
    .map(([, value]) => value);
```

library/src/schemas/enum/enum.ts
fabian-hiller: In a perfect world, I would only filter reverse mapping entries of "real" TS enums and not of normal objects, as this could lead to unexpected behaviour if someone accidentally provided an object with a reverse mapping entry that was not intended as a reverse mapping entry. But the problem is that we can't tell at runtime what is a TS enum and what is a normal object. Therefore, I will probably follow your recommendations and document this behavior in our API reference.

library/src/schemas/enum/enum.ts
fabian-hiller: Yes, this looks good to me. Should I take over and finalize the PR or do you want to add this change?

library/src/schemas/enum/enum.ts
```diff
@@ -18,6 +18,48 @@ export interface Enum {
...
+      : V
+    : never;
+
+/**
+ * enum_ only accepts enum values of string & non-numeric keys that has its reverse mapping.
+ * This type-level function filters out numeric keys including Infinity and NaN.
+ *
+ * @example EnumValues<{ 1: NaN; NaN: 1; 2: 'foo' }> = NaN | 'foo'
+ */
+type EnumValues<TEnum extends Enum> = {
```
fabian-hiller, line 55: I would change the implementation of `EnumValues`. I think the current code produces wrong types in some cases. I tried to "copy" the JS code we use for filtering to the TS type.

```ts
type EnumValues<TEnum extends Enum> = {
  [TKey in keyof TEnum]: TKey extends number
    ? TEnum[TKey] extends string
      ? TEnum[TEnum[TKey]] extends TKey
        ? never
        : TEnum[TKey]
      : TEnum[TKey]
    : TKey extends 'NaN' | 'Infinity' | '-Infinity'
      ? TEnum[TKey] extends string
        ? TEnum[TEnum[TKey]] extends number
          ? never
          : TEnum[TKey]
        : TEnum[TKey]
      : TKey extends `${infer TNumber extends number}`
        ? TEnum[TKey] extends string
          ? TEnum[TEnum[TKey]] extends TNumber
            ? never
            : TEnum[TKey]
          : TEnum[TKey]
        : TEnum[TKey];
}[keyof TEnum];
```

fabian-hiller: Once we have decided how to proceed with `EnumValues`, I will do a final review and merge this PR.

fabian-hiller: I will continue with my `EnumValues` implementation. Hope that is ok. Feel free to give me feedback.

fabian-hiller:
Thank you! 🧩

fabian-hiller:
v1.0.0-beta.10 is available

PR #948, opened 2024-11-30
improve `optionals` guide

EltonLobo07:
Improvements:
1. Replace `Input` and `Output` with `InferInput` and `InferOutput` in one of the examples.
2. Change the behavior of the `Dependent default values` example to enable TypeScript to infer the correct output type.

More information related to the 2nd improvement:
The example without the improvement does not work as [intended](https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh6BjCAO1XgGEymBXU5DGBsAykwAWAUxDI4AXkwA6MMDCSAFPThKIFAFaSmMdQG8t2lAC4lrXiAqSo6gJQAac9orWMi2-ccu7hZwqHbeihBgQmxk6j5+Dk7ObuYAvinaPjBQyOzY0CDq6sCsYLwwzvIAfHBmwcDYcMWl5YqhBHKdcLysACaS2CWSvZV1wXBQkjC8UKy1cIqLJWUwriFhcMutsgDUmy0wihRwqQzBqR4TUzNzWzBnJ870zgz02QCeteYs7PC9gshrD0ANasCAAdzmChMgLgAEY1l44AAmNbtawAZlO3zYHCuoVI8AUPjAyCgqA03FIfAE0VYYikMjW-xgyBe5gA9By4AAVd5qOCOPBQOBSSY437rEDWBKOeT4-iHdoMVJwJiCCTqIXQUYS1AQUiSRTapwm9kXIA). This might be because TS does not narrow the type of a generic function parameter ([related](https://github.com/microsoft/TypeScript/issues/13995)) but using the ternary operator narrows the type and everything works as [intended](https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh6BjCAO1XgGEymBXU5DGBsAykwAWAUxDI4AXkwA6MMDCSAFPThKIFAFaSmMdQG8t2lAC4lrXiAqSo6gJQAac9orWMi2-ccu7hZwqHbeihBgQmxk6j5+Dk7ObuYAvinaPjBQyOzY0CDq6sCsYLwwzvIAfHBmwVCSMLxQrLUeFoqdJWUwQcEhYXDd5YqhBHITcLysACaS2CWSM3AA-EOlI7IA1Os9ihRw1sMwo3Z92qkMF870zgz02QCebdos7PAzgsjW0wDWrBAAO6tBQmb5wACMrjgXjgACZoWNrABmS7mN4cOANUKkeAKHxgZBQVAabikPgCaKsMRSGTQz4wZB3cwAehZcAAchA4DBHmo4I48FB0WxMUi4AlHPIsZIcScxgxUnAmIIJOpBdBKnVlaKIKRJIoNU4jczUvQgA). There is one major difference between the current and improved examples. The current example returned the same object reference but the improved example doesn't when `sum` is not `undefined` and I think it is fine because:
1. The `transform` action is meant to transform values, so relying on referential equality should be discouraged.
2. I wasn't able to find a better way to let TS infer the type of the output (without a type assertion or something similar).

### DIFF
--- a/website/src/routes/guides/(schemas)/optionals/index.mdx
+++ b/website/src/routes/guides/(schemas)/optionals/index.mdx
@@ -79,11 +79,9 @@ const CalculationSchema = v.pipe(
     b: v.number(),
     sum: v.optional(v.number()),
   }),
-  v.transform((input) => {
-    return {
-      ...input,
-      sum: input.sum === undefined ? input.a + input.b : input.sum,
-    };
-  })
+  v.transform((input) => ({
+    ...input,
+    sum: input.sum === undefined ? input.a + input.b : input.sum,
+  }))
 );
 ```
--- a/website/src/routes/guides/(schemas)/optionals/index.mdx
+++ b/website/src/routes/guides/(schemas)/optionals/index.mdx
@@ -7,6 +7,7 @@ description: >-
 contributors:
   - fabian-hiller
   - fartinmartin
+  - EltonLobo07
 ---
 
 import { Link } from '@builder.io/qwik-city';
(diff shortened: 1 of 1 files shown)


fabian-hiller:
Thank you! I chose the current implementation to not clone the object unnecessarily for performance reasons when not needed. But your code is ok too. But let us simplify the code and remove the return statement by writing `() => ({ ... })`.

PR #950, opened 2024-12-02
feat: flavored types

vktrl:
Less strict version of `brand`. The code is basically the same, just a couple of more test cases.


fabian-hiller:
Thank you for creating this PR. Can you explain the pros and cons of `flavor` vs. `brand`? What are the use cases for `flavor`? Do other schema libraries have the same or a similar feature in addition to `brand`?

vktrl:
> Thank you for creating this PR. Can you explain the pros and cons of `flavor` vs. `brand`? What are the use cases for `flavor`? Do other schema libraries have the same or a similar feature in addition to `brand`?

I'm not aware of other validation libraries using flavored types in addition to branded ones, but there's a database schema library called [kanel](https://kristiandupont.github.io/kanel/kanel-kysely.html) that makes use of them. The purpose there is to avoid mixing up ids of different entities, but not require the use of a validation library or custom functions for casting new ids to be used as arguments.

I prefer using flavored types for ids and ISO timestamps because all the existing values are more or less guaranteed to be flavored from either validated API inputs or database queries. There's a loss in ergonomics for no or very little extra safety when branded types force me to implement wrapped functions a la `createUserUuidv4()` or `createUuidv4('UserId')` or `toIsoTimestamp(new Date())`. There are some more examples in this [blog post](https://spin.atomicobject.com/typescript-flexible-nominal-typing/).

tl;dr - flavored types don't get in my way, but will stop me from mixing up validated data (and all my data is validated.) In critical places I use runtime assertions and/or branded types.

fabian-hiller:
Is the only difference that the `FlavorSymbol` key is optional while the `BrandSymbol` key is required?

vktrl:
> Is the only difference that the `FlavorSymbol` key is optional while the `BrandSymbol` key is required?

That's right.

fabian-hiller:
Am I understanding correctly that a flavored type can be passed on to an unflavored type, but not to another flavored type, and that's the difference with a branded type and the idea behind `flavor`? Here is an example:

```ts
type A = string & { foo?: 'A' };
type B = string & { foo?: 'B' };

function createA(): A {
  return 'hello';
}

const test1: A = createA();       // This works
const test2: string = createA();  // This works
const test3: B = createA();       // But this does not work
```

vktrl:
Yes, I added similar examples to the docs [here](https://github.com/fabian-hiller/valibot/blob/a0b70e2d198afe926f4c18417be821685819f89a/website/src/routes/api/(actions)/flavor/index.mdx).

In essence:
- `Flavored<string, 'a'>` is a valid `string` and vice versa
- `Flavored<string, 'b'>` is not a valid `Flavored<string, 'a'>`

fabian-hiller:
In general, I like the idea of a `flavor` action. I am just careful not to make the library too complicated. It will be very important to document the difference to `brand` well so that users do not get confused. Do you have an alternative name for this action?

vktrl:
Maybe `weakBrand`? Could even use the same symbol for convenient interop.

fabian-hiller:
Here is a list of names with `Brand` as a suffix:
- `weakBrand`
- `softBrand`
- `partialBrand`
- `optionalBrand`

And here is a list of completely new and independent names:
- `flavor`
- `mark`
- `tag`
- `flag`
- `tint`

What's your favorite so far?

vktrl:
I like `weakBrand` the most + using the same symbol. I'll update the PR if you're ok with that.

fabian-hiller:
I am not sure we should use the same symbol. This could cause problems when people extend existing schemas that end up using both actions. Let me think about the name a bit more. I'm not sure what's best yet.

fabian-hiller:
Thanks for your feedback. I will have to investigate this before moving forward. I may prioritize our v1 RC first before doing this.

vktrl:
Thanks for merging the PR!

Would you consider interop with brand by using the same symbol as I preposed earlier?

fabian-hiller:
Would the main benefit be to be able to use both together somehow?  Feel free to provide a code example.

fabian-hiller:

vktrl:
The fundamental purpose of Brand and Flavor are basically the same - to distinguish between primitive types. Using them both together only makes sense if you're extending a schema to make it stricter, i.e. from flavored to branded.

The main benefit is developer ergonomics. For example:

```ts
// Using Valibot's Flavor type for the database ids for consistency.
// Otherwise the db client would accept any Valibot branded or flavored string
// and only check for its own flavor type mismatch.

type DB = {
  users: {
    id: FlavorAction<string, "UserId">,
    email: FlavorAction<string, "Email">
  }
}

const db = new Kysely<DB>();

// Using branded ids when validating user input
// and defining base arg types shared across the app in one place

const UserIdSchema = v.pipe(v.string(), v.brand('UserId'));
type UserId = v.InferOutput<typeof UserIdSchema>;

const UserArgsSchema = v.object({ id: UserIdSchema });
type UserArgs = v.InferOutput<typeof UserArgsSchema>;

// Let's say we want to use a branded id here to be extra sure
async function deleteFromExternalService(id: UserId) {
  // ...
}

// A. Shared symbol
const router = t.router({
  delete: t.procedure.input(UserArgsSchema).mutation(async ({ input }) => {
    input.id satisfies FlavorAction<string, 'UserId'> // 🟢 no need to cast
    await db
      .deleteFrom('users')
      .where('id', '=', input.id)
      .executeTakeFirstOrThrow();
    await deleteFromExternalService(input.id);
  })
})

// B. Distinct symbols
const router = t.router({
  delete: t.procedure.input(UserArgsSchema).mutation(async ({ input }) => {
    // contrived example, args are usually passed through multiple functions across multiple files
    const userId = input.id as BrandAction<string, 'UserId'> // 🔴 needs to be cast or an assertion function
    await db
      .deleteFrom('users')
      .where('id', '=', userId)
      .executeTakeFirstOrThrow();
    // if we changed the input schema to use FlavorAction then we'd have to cast it to BrandAction here intead
    await deleteFromExternalService(input.id);
  })
})
```

I think it's safe to assume that `BrandAction<string, "UserId">` being a valid `FlavorAction<string, "UserId">` wouldn't be unexpected behavior since the distinguishing part from the developer's point of view is not which symbol used in the background, but the name itself - `UserId`. It might make sense to rename it to weakBrand for extra clarity though, or keep flavor as a distinct action and add a weakBrand.

Flavor having a distinct symbol would work fine, but it forces you to commit to using one or the other exclusively to avoid type assertions. Since my database ids are flavored then branding brings more pain than it's worth.

fabian-hiller:
I am generally open to improving the implementation in this regard, but we should probably find a generic name for the symbol that makes sense for both actions. Do you have an idea?

vktrl:
> I am generally open to improving the implementation in this regard, but we should probably find a generic name for the symbol that makes sense for both actions. Do you have an idea?

Since the goal is nominal typing, then perhaps `NominalSymbol`?. I also wouldn't mind renaming `Flavor` to `WeakBrand`.

fabian-hiller:
We could also call it `ValibotSymbol`. Do you want to create a new issue to discuss this? That way I can add it to our repo milestones.

vktrl:
https://github.com/fabian-hiller/valibot/issues/1194

PR #956, opened 2024-12-03
Feat: Abort async implementation (promises) early if `abortEarly` is set

### LINKED ISSUE #954: Looks like `abortEarly` doesn't abort async parsing
In the following code I expect that parsing will be stopped right away, because `email` field is not valid, and `abortEarly` is set to `true`. But parsing awaits for the whole 10 seconds before returning result.

Ideally I expect that promise will not be even started in that case.

And the related question:
Is there a way to abort running promise? Using `AbortController` or something?

```js
import * as v from 'valibot';

const Schema = v.objectAsync({
  email: v.pipe(v.string(), v.trim(), v.email()),
  username: v.pipeAsync(
    v.string(),
    v.nonEmpty(),
    v.maxLength(30),
    v.checkAsync(() => new Promise((res) => {
      console.log('request');
      setTimeout(() => res(true), 10000)
    }))
  ),
});

const result = await v.safeParseAsync(Schema, {
  email: '',
  username: 'qwerty',
}, {
  abortEarly: true,
  abortPipeEarly: true,
});

console.log(result);
```

https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh6BjCAO1XgGUmALAUxDI4AXkwA6CBQBWfJjACCqAJ6smACgDe9OHAHJgpAFziwwMHzUYxHKMFYBzNQEoANOJi2Qzt1b0HnrtpwAK6ofFCsyCB8xlam5ooq6kE6VjZ2joE6qWKsbACi4DBK3iniggAeADJ8DjA8agDMAAxZ2Va8sgDWiapqziIAfHCsfADucAAKeCDAYf1QfKhOQ3Ba2dks7BCkfGKkEI6EiwCOwUs0TgwbOmEwACqgfBDBMP0rwsOLqGoe565wACMzRBrTKAF8nE4goFIQxmGwOHBvsFSPBRMgxvp4GlkNg+JNkFAwr11Nx+II3OtdIIDMZCIQXEFQuFItF6ScxuFioz6OCqUFkFRYPkiaQlMY-nwmTohdAYJMzHxRVBxZKoOcmXDGFtUDs9gdHCi0Vd6EA

yumauri:
PoC to fix #954

### DIFF
--- a/website/src/routes/guides/(main-concepts)/issues/index.mdx
+++ b/website/src/routes/guides/(main-concepts)/issues/index.mdx
@@ -6,6 +6,7 @@ description: >-
 contributors:
   - fabian-hiller
   - BastiDood
+  - EltonLobo07
 ---
 
 import { Link } from '@builder.io/qwik-city';
@@ -119,13 +120,13 @@ If you are only interested in the error messages of each issue to show them to y
 import * as v from 'valibot';
 
 const ObjectSchema = v.object({
-  key: v.string('Value of "key" is missing.'),
-  nested: v.object({
-    key: v.string('Value of "nested.key" is missing.'),
+  foo: v.string('Value of "foo" is missing.'),
+  bar: v.object({
+    baz: v.string('Value of "bar.baz" is missing.'),
   }),
 });
 
-const result = v.safeParse(ObjectSchema, { nested: {} });
+const result = v.safeParse(ObjectSchema, { bar: {} });
 
 if (result.issues) {
   console.log(v.flatten<typeof ObjectSchema>(result.issues));
@@ -142,15 +143,16 @@ The `result` returned in the code sample above this text contains the following 
     input: undefined,
     expected: 'string',
     received: 'undefined',
-    message: 'Value of "key" is missing.',
+    message: 'Value of "foo" is missing.',
     path: [
       {
         type: 'object',
         origin: 'value',
         input: {
-          nested: {},
+          bar: {},
         },
-        key: 'key',
+        key: 'foo',
+        value: undefined,
       },
     ],
   },
@@ -160,22 +162,23 @@ The `result` returned in the code sample above this text contains the following 
     input: undefined,
     expected: 'string',
     received: 'undefined',
-    message: 'Value of "nested.key" is missing.',
+    message: 'Value of "bar.baz" is missing.',
     path: [
       {
         type: 'object',
         origin: 'value',
         input: {
-          nested: {},
+          bar: {},
         },
-        key: 'nested',
+        key: 'bar',
         value: {},
       },
       {
         type: 'object',
         origin: 'value',
         input: {},
-        key: 'key',
+        key: 'baz',
+        value: undefined,
       },
     ],
   },
@@ -187,8 +190,8 @@ However, with the help of <Link href="/api/flatten/">`flatten`</Link> the issues
 ```ts
 {
   nested: {
-    key: ['Value of "key" is missing.'],
-    'nested.key': ['Value of "nested.key" is missing.']
-  }
-}
+    foo: ['Value of "foo" is missing.'],
+    'bar.baz': ['Value of "bar.baz" is missing.'],
+  },
+};
 ```
--- a/website/src/routes/guides/(schemas)/optionals/index.mdx
+++ b/website/src/routes/guides/(schemas)/optionals/index.mdx
@@ -7,6 +7,7 @@ description: >-
 contributors:
   - fabian-hiller
   - fartinmartin
+  - EltonLobo07
 ---
 
 import { Link } from '@builder.io/qwik-city';
@@ -47,8 +48,8 @@ import * as v from 'valibot';
 
 const OptionalStringSchema = v.optional(v.string(), "I'm the default!");
 
-type OptionalStringInput = v.Input<typeof OptionalStringSchema>; // string | undefined
-type OptionalStringOutput = v.Output<typeof OptionalStringSchema>; // string
+type OptionalStringInput = v.InferInput<typeof OptionalStringSchema>; // string | undefined
+type OptionalStringOutput = v.InferOutput<typeof OptionalStringSchema>; // string
 ```
 
 By providing a default value, the input type of the schema now differs from the output type. The schema in the example now accepts `string` and `undefined` as input, but returns a string as output in both cases.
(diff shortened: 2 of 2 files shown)


fabian-hiller:
Thank you for creating this PR. There is a problem. The current implementations ensure consistency. The same dataset will always return the same result, regardless of whether the promises take different amounts of time. But if we abort after the first input validation with an issue, the order may be random. Of course, we could choose performance over consistency in this case, but it would be a breaking change.

Your implementation has added a lot of extra code. Let us try to keep it to a minimum. Here is a quick idea I had in mind. Feel free to experiment with it and share your thoughts.

```ts
const valueDatasets = await Promise.all(
let shouldAbort = false;
const valueDatasets = await Promise.all(
  Object.entries(this.entries).map(async ([key, schema]) => {
    if (!shouldAbort) {
      const value = input[key as keyof typeof input];
      const valueDataset = await schema['~run']({ value }, config);
      if (config.abortEarly && valueDataset.issues) {
        shouldAbort = true;
      }
      return [key, value, valueDataset] as const;
    }
  })
);
```

fabian-hiller:
> This will not work because, once launched, a promise can’t be aborted from the outside.

That may be true, but we can still somehow change what the promise actually does, and e.g. not execute the actual expensive logic, or am I wrong?

If I understand your current code correctly it executes the promises in sequential order and I am not sure if this is an option.

> But even if you think it is, the library is still in its zero version, so it’s not a big deal, in my opinion.

I agree, and am willing to investigate this case and favor performance if we find a good and uniform solution that works for any async schema.

yumauri:
> but we can still somehow change what the promise actually does

No, a promise is like a black box. Once launched, you can only wait for it to resolve (or reject). Or _ignore it_.
The pattern of wrapping a promise in `Promise.race` to "abort" it doesn’t actually _abort_ the promise itself; it just _ignores it_.

Or maybe I just don’t understand what you’re pointing out 🤔

> If I understand your current code correctly it executes the promises in sequential order and I am not sure if this is an option.

No, the idea was that I collect promises in an array but do not await them. Later, using `Promise.race` on that array, I get the winning promise/validation, check if it has issues or not, and if it doesn’t, I remove it from the array and then await the next winner, and so on. However, all asynchronous validations are launched simultaneously.

But after I added a few tests, I found a critical issue in my approach: after removing a promise from the array, the indexes of the remaining promises became misaligned.

So, I rewrote the PoC without using `Promise.race` at all :)
I also added a test that clearly demonstrates that all async validations start in parallel.

Also, I added logic so that if I encounter synchronous issues while executing validations, it stops right there and doesn’t launch the async validation at all (if it is located below the synchronous one).

As for the rest, what I want to try is to add the possibility to abort running promises by passing an AbortController.signal to each one of them. This way, the user can hook into this signal to abort an HTTP request, for example, if validation fails on another field.

fabian-hiller:
> Or maybe I just don’t understand what you’re pointing out 🤔

Look at [this simple example](https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh7SBTeZK2AIWSjgF4cZVMwb0AxhAB2qeAGUxAC2Yhk-TADoIFAFbMxMAIKoAnpLEAKAN704OCBABcGsMDDNzN2xplRgkgObmAJQANJ62GOqKegDW5sH8AHxw1l5eEtIQLOqkEIGE0WIxcABE6tj2JYRBDGm27NAw3LwCMFAAriLhXlCs7VCSgqTCtV4AvqGek7YUPE6RLm5GphbdkT5+gdNekYUxy2bmaCtwCXzJqXXA2KcAhA1cPEEp3WkZqFnMOXnmBUpFpXUsygVRqry8yAA7shgPBJMxIXAAAp4EDAYTxXqoZ7nODCGAAFVAzAg7Rg5ixITgAEYAAz02lBMF1OBjcG9GD9QZtTqjWwTMK2SYTUTveBY9qkeACKEw+DrZDYZhInjCA6rWzyJQqQUpOyOIiEKnApyEQis3WWFAcGAAUR4pGMTh5zAt9DB4ikH2yuUCEqlYKAA). `.foo` is executed before `.bar` and can therefore abort the expensive await timeout execution. This probably only solves the problem of this PR if sync validation is mixed with async validation.

Thank you for your research! I really appreciate your efforts and hope we can improve the current implementation. I don't have much time at the moment, but I will try to have a closer look at the end of December or in January. Feel free to share any research you do here, as well as alternative implementations if you find any. I can also share this PR on social media to get additional feedback and ideas from others.

yumauri:
Yeah, this is almost the same as what I’m doing [here](https://github.com/fabian-hiller/valibot/pull/956/files#diff-ea7ace46d94403decc367006fc9219e2cb038e728ee4fe477f7f96e7b08fe482R147-R157) — if the validation result is data (not a promise-like object) and has issues, it stops the iteration right there and doesn’t even launch the following promises (already launched ones will be ignored). And yes, this will fail a test with two asynchronous operations of different durations.

> I can also share this PR on social media to get additional feedback and ideas from others.

Sure, why not 👍

yumauri:
I was thinking about how I could add an `AbortController` here to signal other running promises about an aborted validation, so they can handle it correctly (for example, aborting a pending HTTP request). In doing so, I stumbled upon a question that hadn't occurred to me before: why are there no exception handling mechanisms in schemas or actions? If I throw an exception from a `check` action or `custom` schema, it will break parsing, even with `safeParse` (which doesn't seem very "safe" to me).

I'm raising this in the following context: if you add an `AbortController.signal` to a `fetch` call and then call abort on the controller, `fetch` will be rejected with an `AbortError` and will throw an exception inside the `await`. So, users must manually handle this case by catching possible errors and returning `false`. Why not wrap any validation in a try-catch block and treat exceptions as `false`?

Maybe I'm shooting from the hip here, as the library's API has been extensively discussed, and it seems no one is concerned about this issue — I haven't found any related discussions or issues. I also haven't found any reasoning behind this design decision.

yumauri:
On the other hand, maybe this is not a problem in my proposal, because after validation abortion all pending promises results are ignored anyways 🤔 I will investigate this.

But I'm still curious about question above in general.

fabian-hiller:
So far, we expect users to catch errors and return true or false. This makes our code consistent throughout the library. We could add a try/catch block to `check` and `checkAsync`, but this would unnecessarily increase the bundle size when not needed, and raises the question of whether other actions with custom code should also be wrapped in try/catch. I am open to discuss this as there are pros for both sides.

fabian-hiller:
I have a little more time from now on and will try to review this PR more closely in the next few weeks.

fabian-hiller:
I will have another look at this PR in the next few weeks, but I may decide to leave this improvement for another major release after v1.

yumauri:
Sure, whatever you think is best.

I just wanted to ask if you could provide a different approach to the problem in #954, using the current API (or the planned v1 API), because this might be a blocker for us to use valibot.

Hope you had a wonderful Christmas and wishing you a happy New Year 2025! 🎄✨

PR #957, opened 2024-12-04
feat: add json string validation action

hendrikheil:
This PR defines a JSON action to validate strings to be valid json strings (not objects).

Recently an Issue was created to document a valibot schema to validate objects to be json compliant. In that PR @fabian-hiller mentioned a validation action to ensure strings are valid JSON. This PR aims to add that action to valibot.

I'm a bit torn if this action should be called `json` or `jsonString`. I lean towards just calling it `json`, which is why I went with that.

Relevant comment: https://github.com/fabian-hiller/valibot/issues/933#issuecomment-2489934225


fabian-hiller:
Thank you Hendrik! I will review this PR after our v1 RC or stable v1 is out.

PR #969, opened 2024-12-09
add `undefinedableAsync` API reference

EltonLobo07:


website/src/routes/api/(async)/undefinedableAsync/index.mdx
```diff
@@ -1,10 +1,237 @@
...
+);
+```
+
+### New user schema
+
+Schema to validate new user details.
+
+```ts
+const NewUserSchema = v.objectAsync({
+  // Assume the email is checked
```
fabian-hiller, line 74: Same here than with the other two PRs.

PR #973, opened 2024-12-11
Only include necessary files for JSR

### LINKED ISSUE #972: JSR-published package contains unnecessary files
Hi, thanks for making this package. :) I'm comparing the file contents of the package from how its published on NPM versus JSR, and on JSR it seems to contain a lot more files than needs be. 

- https://www.npmjs.com/package/valibot?activeTab=code, contains config files like vite.config.ts, eslint.config.js, etc
- https://jsr.io/@valibot/valibot/0.42.1/package.json, contains only what's needed

This fix though is easy as adding a `publish.include` field to `jsr.json`, see https://jsr.io/docs/package-configuration#include-and-exclude and https://jsr.io/docs/publishing-packages#filtering-files. Would be happy to write a PR

----

image
image

neoncitylights:
This will publish `jsr.json`, `LICENSE.md`, `README.md`, and everything under `src/` directory. To locally verify this:
```sh
jsr publish --dry-run
```

Fixes #972


fabian-hiller:
Thank you for creating this PR. Why do you think it is better to publish only selected files? Is `package.json`, `LICENSE.md` and other files not necessary?

neoncitylights:
No problem :) And thanks for the question, I should have been more explicit in the PR.

There's no official recommendation from the JSR team on that; the reason for this is more just to reduce disk size when users install the library via JSR's compat layer with `npm` or the sorts (any that uses `node_modules`). 

Regarding files:
- `LICENSE.md` isn't explicitly in the include array, but is [specially whitelisted by JSR](https://jsr.io/docs/troubleshooting#missinglicense) to be included no matter what. 
- JSR packages only need either a `jsr.json`, `deno.json`, or `package.json` when published. It's still fine to have a mix of `jsr.json`/`deno.json` (one of these two) and `package.json`, but the JSR package would still be usable with just `jsr.json`.
- The [root `mod.ts` file](https://github.com/fabian-hiller/valibot/blob/main/library/mod.ts) is currently ignored since you tell `jsr.json` to only export from `src/index.ts`. Like in a lot of the official `@std/` packages, it *does* make sense to include `mod.ts` if when you're writing a package where you don't have a `src/` directory, the code is in the root, but you still want to export everything under a single entrypoint; however, that doesn't apply here. (I didn't initially delete `mod.ts` file since I had accidentally not noticed that, sorry for the confusion)

Hope that clarifies things!  Likewise I understand this is a bit of an opinionated PR (and not an official JSR team recc), so feel free to close it.

fabian-hiller:
Sorry for my late reply. I was focused on my studies. I will review this PR in the next weeks.

fabian-hiller:
Thank you for this PR. I added `library/mod.ts` back because we need it for `deno.land/x/valibot`.

Issue #968, opened 2024-12-09
Best practice for valibot to work with typescript's isolated declarations feature


hyf0:
With enabling https://www.typescriptlang.org/tsconfig/#isolatedDeclarations, `tsc` require to write code with explicit type annotations:

For exmaple:

The case in README.md

```ts
const LoginSchema: /* tsc force you write type here, so what should be write here for valibot? */ = v.object({
  email: v.pipe(v.string(), v.email()),
  password: v.pipe(v.string(), v.minLength(8)),
});
```

For small schemas, I could just type them there. But I have very big schemas which isn't proper to manually write them.

`zod` has a helper type `ZodType` to infer schemas from types. So for zod, it could be written as

```ts
const LoginSchema: ZodType<{ email: string, passworld: string }> = v.object({
  email: v.pipe(v.string(), v.email()),
  password: v.pipe(v.string(), v.minLength(8)),
});
```

shulaoda:
A hacker's way:

```js
const LoginSchema = v.object({
  email: v.pipe(v.string(), v.email()),
  password: v.pipe(v.string(), v.minLength(8)),
});

export const loginSchema: typeof LoginSchema = LoginSchema;
```

fabian-hiller:
Valibot offers `GenericSchema` (similar to `ZodType`). But in general this is not a good idea because a lot of information will be lost. I recommend to disable this rule for your schema files or find another workaround. Please keep me posted and feel free to contact me if you have any further questions.

shulaoda:
> But in general this is not a good idea because a lot of information will be lost.

For simple schemas, I can directly use `GenericSchema,` but for more complex schemas like `UnionSchema`, it will be insufficient, leading to potential type conflicts.

shulaoda:
> but for more complex schemas like `UnionSchema`

I have conducted a thorough study and found that `GenericSchema` can be used in all cases, although there are some unexpected outcomes.

shulaoda:
This causes an type error because `v.function()` generates an additional type of `(...args: unknown[]) => unknown`. Could you tell me how to handle this? @fabian-hiller 

```ts
const argsFnSchema: v.GenericSchema<(...args: number[]) => boolean> = v.pipe(
  v.function(),
  v.args(v.tuple([v.array(v.number())])),
  v.returns(v.boolean()),
)
```

fabian-hiller:
[...]

PR #976, opened 2024-12-12
fix(library): correct type inference in `ArgsAction`

shulaoda:
Actually, I found that `type Output = v.InferOutput<typeof ComplexSchema>` will be deduced as:
```ts
type Output = {
    fn: (args_0: string) => boolean;
    argsWithFn: (args_0: (...args: unknown[]) => unknown) => unknown;
}
```
Is this the expected behavior?  Is there a problem with my writing for the `args_0` of the `argsWithFn`?

_Originally posted by @shulaoda in https://github.com/fabian-hiller/valibot/issues/968#issuecomment-2537846099_

The second generic of `BaseTransformation` should be `InferOutput<...>`.

Playground: https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh6BjCAO1XgGF8xSBTAB4BlJgAs+IZHAC8mAHQcowJjADyFAFZ8VACgDe9ODlYAueWGBg+Ow0fnYArqxXA2OgJQAaW0YxzkUADmqDp+MA681gDaforArIEeALruXj7yUHzhUOyhclQQ-MisHmlGZShBqADqwDCiAGKm5pbW6X6OzjCuJRW+-lU2dnZhEfw6UenDfhZWQ8MLHU4ubn0L-QHBeeGRE34FRb0pa+t+mdm5sTBKCaXe6+X3CylPj+1y5w45IVc3ialPNIAX3cDHoMAAnlY4KoHDAwHCZPIAJKsbB8KCw+FwgA8kKsEGwcG44H4wjEEmQAD4wSx2PAIHCETAzFjmUiDEZsM0dJszHEEu4ZFS4Ho4J8cnB8XxCZVAjJpLJCALAoQ4ECnpsanVGjzuWZeUEzAc+MUhdIRSrzSKxRLWMYdNgyKg+EKgfR3cw2KhCnw5KQIIlGdiYO56EA

### DIFF
--- a/library/src/actions/args/args.ts
+++ b/library/src/actions/args/args.ts
@@ -13,7 +13,7 @@ import type {
   BaseSchema,
   BaseTransformation,
   ErrorMessage,
-  InferInput,
+  InferOutput,
   TupleItems,
 } from '../../types/index.ts';
 import { ValiError } from '../../utils/index.ts';
@@ -42,7 +42,7 @@ export interface ArgsAction<
   TSchema extends Schema,
 > extends BaseTransformation<
     TInput,
-    (...args: InferInput<TSchema>) => ReturnType<TInput>,
+    (...args: InferOutput<TSchema>) => ReturnType<TInput>,
     never
   > {
   /**


fabian-hiller:
I think there was a reason why I choose `InferInput`. I will investigate it to check if it is a bug and report back to you. Thank you!

fabian-hiller:
Hey, I think the current implementation with `InferInput` is correct and your change with `InferOutput` would be a bug. The reason for this is that when you call the function, the argument you pass must match the input of the schema, not the output. The argument schema will then parse the input and pass its value to the function. I know this is hard to follow, but [here](https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh6BjCAO1XgGUYphWBzACoQAcgFcQFAKZROTABZSQyOAF5MAOjDAwUgBT04mjrwF6AlABpDmgCZSmoMhetGMGnsnbZoIPeMkZK3pzBmY2DjgAMTFWJhhgNjlFZTVNbV0DNw1sWPjE1hcbd2QoflQ9dxgxMFJ9AG1uUyFRCWlZBSVkAF1zYOyoKWqodkqNVjaZC2DQxhZ2eFsIMQo6gPa09zBS1H0YuISkzuVLOEqyMSkALjgJwKhzNQA+TAupRDgAJlnwhbhB1BiUjwdRLFZrSZQPSET4aACshFCcAA9Mi4AAeAC0mLggkUKDKcBAYki0hQcAABiY+PwKbdqOSKXd2hTfqgIHUNKQIPw9ACgTAkai4HD6EA) is an example that demonstrates this case.

```ts
import * as v from 'valibot';

const StringToNumberSchema = v.pipe(
  v.string(),
  v.decimal(),
  v.transform(Number)
);

const FunctionSchema = v.pipe(
  v.function(),
  v.args(v.tuple([StringToNumberSchema])),
  v.returns(v.number())
);

const doubleNumber = v.parse(FunctionSchema, (value: number) => value * 2);

const result = doubleNumber('2.5'); // <-- The arg must be a `string` not a `number`
console.log(result); // 5
```

shulaoda:
Got it. Thanks again!

PR #984, opened 2024-12-15
Reduce TypeScript target to ES2020

andersk:
Support for extremely modern ES syntax still isn’t universal, and is totally unnecessary for us, since the only diff this results in is

```diff
--- library/dist/index.js
+++ library/dist/index.js
@@ -237,8 +237,4 @@
 var ValiError = class extends Error {
   /**
-   * The error issues.
-   */
-  issues;
-  /**
    * Creates a Valibot error with useful information.
    *
```

(The .d.ts file is unaffected, so no documentation is lost.)

### DIFF
--- a/library/tsconfig.json
+++ b/library/tsconfig.json
@@ -10,7 +10,7 @@
     "noEmit": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "ESNext"
+    "target": "ES2020"
   },
   "include": ["src"]
 }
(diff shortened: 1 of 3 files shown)


fabian-hiller:
Thank you for creating this PR. Is this even necessary if there is no relevant change in the output files?

andersk:
The change that I quoted is necessary and relevant, since it removes an unnecessary line (`issues;`), but fixes an incompatibility with ES2020 parsers.

fabian-hiller:
Thanks for your quick reply. Do you think there are any drawbacks if this line is removed?

andersk:
There are no drawbacks to removing this line—that’s what I’ve been saying.

fabian-hiller:
Thank you for your feedback! I will investigate it in the next few days and probably merge it.

fabian-hiller:
v1.0.0-beta.10 is available

Issue #945, opened 2024-11-29
feat: return a template literal for `uuid` action


sandros94:
This is just a small idea, feel free to close.

While working with [node/web crypto](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) I've noticed that `randomUUID` returns a template literal:

```
(alias) randomUUID(): `${string}-${string}-${string}-${string}-${string}`
import randomUUID
Available only in secure contexts.
```

Would it be possible to do the same for `uuid` action? Is it sufficient to change the `TOutput` of its `BaseValidation`?

https://github.com/fabian-hiller/valibot/blob/eac543c31d720aeb310045c7b37118a59a33cb50/library/src/actions/uuid/uuid.ts#L41

fabian-hiller:
Thanks for creating this issue. At the moment `uuid` is just a validation action and they don't transform output types. Nevertheless, your idea is great and I will take a close look at it in the next few weeks.

fabian-hiller:
Feel free to join our discussion in #986, which is strongly related to this issue.

fabian-hiller:
As I wrote in PR #986, I plan to look at this again before releasing v2. If you would like such a feature (for other similar actions as well), please give this comment a thumbs up and share your use case with us.

lotap:
I was looking for similar functionality in the output of `v.ip()` (and ipv4 and ipv6) and found this issue

From a user perspective, it makes sense that after passing validation the type should be narrowed

After reading https://github.com/fabian-hiller/valibot/pull/986 I understand it may not be as straightforward as it seems on the surface. But for my use-case, `v.custom` provided the desired functionality
```ts
const NarrowingIpv4 = v.custom<`${number}.${number}.${number}.${number}`>(
  (input) =>
    typeof input === "string"
      ? /^(?:(?:[1-9]|1\d|2[0-4])?\d|25[0-5])(?:\.(?:(?:[1-9]|1\d|2[0-4])?\d|25[0-5])){3}$/u.test(
          input
        )
      : false
);
```
It would be nice if that behavior was built-in

fabian-hiller:
Thank you for your feedback! I will consider it in the long run.

sandros94:
> Hi, [@sandros94](https://github.com/sandros94). I'm [Dosu](https://dosu.dev), and I'm helping the Valibot team manage their backlog. I'm marking this issue as stale.

should we let this auto-close? This is just a small QoL improvement, not really required

[...]

PR #986, opened 2024-12-17
Feat start with types

aadito123:
Changes the output type for `startsWith` and `endsWith` to actually reflect.

Previously:
`startsWith("abc")` -- output --> `string`
Now:
`startsWith("abc")` -- output --> `abc${string}`

Similarly with endsWith.

### DIFF
--- a/library/src/actions/endsWith/endsWith.test-d.ts
+++ b/library/src/actions/endsWith/endsWith.test-d.ts
@@ -37,7 +37,7 @@ describe('endsWith', () => {
     });
 
     test('of output', () => {
-      expectTypeOf<InferOutput<Action>>().toEqualTypeOf<string>();
+      expectTypeOf<InferOutput<Action>>().toEqualTypeOf<`${string}foo`>();
     });
 
     test('of issue', () => {
--- a/library/src/actions/endsWith/endsWith.ts
+++ b/library/src/actions/endsWith/endsWith.ts
@@ -43,7 +43,7 @@ export interface EndsWithAction<
   TMessage extends
     | ErrorMessage<EndsWithIssue<TInput, TRequirement>>
     | undefined,
-> extends BaseValidation<TInput, TInput, EndsWithIssue<TInput, TRequirement>> {
+> extends BaseValidation<TInput, `${TInput}${TRequirement}`, EndsWithIssue<TInput, TRequirement>> {
   /**
    * The action type.
    */
--- a/library/src/actions/startsWith/startsWith.test-d.ts
+++ b/library/src/actions/startsWith/startsWith.test-d.ts
@@ -37,7 +37,7 @@ describe('startsWith', () => {
     });
 
     test('of output', () => {
-      expectTypeOf<InferOutput<Action>>().toEqualTypeOf<string>();
+      expectTypeOf<InferOutput<Action>>().toEqualTypeOf<`foo${string}`>();
     });
 
     test('of issue', () => {
--- a/library/src/actions/startsWith/startsWith.ts
+++ b/library/src/actions/startsWith/startsWith.ts
@@ -45,7 +45,7 @@ export interface StartsWithAction<
     | undefined,
 > extends BaseValidation<
     TInput,
-    TInput,
+    `${TRequirement}${TInput}`,
     StartsWithIssue<TInput, TRequirement>
   > {
   /**


fabian-hiller:
Thanks for creating this PR. The idea is great. Right now `startsWith` and `endsWith` are just validation actions and don't transform their output type. This would force us to change `.kind` from `'validation'` to `'transformation'`, but that feels wrong somehow because the main use case is the validation. We could also allow validation actions to transform their output type in such cases. For me, a strong mental model is very important. So it's important for me to make the right decision in this case.

Issue #945 is related.

aadito123:
Fixed an issue to stop piping the TInput.
For example, `pipe(string(), startsWith('hi'), startsWith('bye'))`, would result in an invalid type of `byehi${string}`.
This would fail validation so there won't be any runtime consequences, and also, breaks the following case:
`pipe(string(), startsWith('hi'), endsWith('bye'))` since it would give a type `${string}bye` instead of `hi${string}bye`.

I think I align with validation transforming the output types, even if the runtime value is unaffected. I believe the validation should serve as a way of effectively narrowing the type. The input for the parse methods already transform the input type, since they can take `unknown` typed inputs, so further narrowing the type via these methods seems consistent to me.

fabian-hiller:
Thank you for your feedback. I will further investigate it in the next few weeks.

fabian-hiller:
I though about this change and noticed a big problem. Have a look at the following example:

```ts
const Schema = v.pipe(v.string(), v.startsWith('foo@'), v.email());
```

If the `startsWith` validation fails with the current implementation, the output is still a string and therefore valid. Therefore, the `email` action can still be executed. If we were to change the implementation so that `startsWith` transforms its output type, we would be forced to mark the output of `startsWith` as untyped, since the output type cannot be guaranteed. This forces the `pipe` to stop its execution as subsequent actions may require this type.

aadito123:
I am not sure I understand what you mean. My PR does not make any runtime changes to the validation. Why would narrowing the output type force us to mark the output as untyped? Furthermore, even if subsequent actions require the type, the input type on the validation actions in pipe should always be the type defined by the schema, and not the previous validation. Therefore, it should not rely on the output type of previous validation. I do understand why that behaviour may not be intuitive.

Is there any way we can have some ability to opt into this more accurate narrowing?

fabian-hiller:
The problem is this. The data type of a pipeline item is based on the previous item. So in the following example, the input type of `email` is based on the output type of `startsWith`. Currently, `startsWith` does not change its output type. Even if `startsWith` fails, we still know that the output will be a string and `email` can be executed.

```ts
const Schema = v.pipe(v.string(), v.startsWith('foo@'), v.email());
```

But if we change the output type of `startsWith` from `string` to ``` `${TRequirement}${string}` ``` the input type of `email` will be based on this type. If `startsWith` fails, we know that the output value does not match the output type. Therefore, we can perform any subsequent action that is based on that type.

In simple words: We can't perform any subsequent action that expects the input to match the ``` `${TRequirement}${string}` ``` type because this may result in a runtime error.

fabian-hiller:
At the moment I see one possible workaround and one solution.

The workaround would be to manually transform to the expected type. This works because a transformation is only executed if there was no problem before. So whenever `transform` is executed, you know that `startsWith` was successful.

```ts
const Schema = v.pipe(
  v.string(),
  v.startsWith('foo@'),
  v.transform((input) => input as `foo@${string}`)
);
```

A possible solution would be to provide a `startsWith2` transformation action (the name is just an example) that not only checks for the start of the string, but also transforms the output type to ``` `${TRequirement}${string}` ``` if successful.

sandros94:
Ah, I do start to understand the issue now.

Although it is not fully clear why we are forced to mark the output as untyped instead of the last successful validation step (in example caching it until the current validation is successful)

fabian-hiller:
I agree. This is probably easy to solve using pure JavaScript, but I am not sure how to make it type safe using TypeScript. 😐

ishiguro-junya:
I was having trouble with type issues, so I'm in favor of adding this action.

hyunbinseo:
Now you can use the `guard` API which removes the need for additional `transform` and type casting:

```ts
const Schema = pipe(
  string(),
  guard((v): v is `foo@${string}` => v.startsWith('foo@')),
);

// type Output = `foo@${string}`;
type Output = InferOutput<typeof Schema>;
```

PR #988, opened 2024-12-18
fix `keyof` method's recursion error when many keys used

EltonLobo07:


fabian-hiller:
Amazing work! Thank you Elton!

fabian-hiller:
v1.0.0-beta.10 is available

PR #989, opened 2024-12-19
Fix VariantIssue mdx: issues property not exist

ikeyan:
This causes 500 error


fabian-hiller:
Great catch! Thank you!

Issue #256, opened 2023-11-19
Add example usage of our Reflection API to the docs


fabian-hiller:
Related to https://github.com/fabian-hiller/valibot/pull/211#issuecomment-1817864021 of @ciscoheat we should add sample code to our documentation that can be copied to easily read the metadata of a schema.

fabian-hiller:
Don't mark it as stale. I will work on it soon.

PR #991, opened 2024-12-22
Doc improve `objects` guide with extraction and composition patterns

Bilboramix:
On the discord server we've seen few questions about how to extract a field's schema from an `object`.
Fabian suggested to explain it in the objects guide so here it is.

### DIFF
--- a/website/src/routes/guides/(schemas)/objects/index.mdx
+++ b/website/src/routes/guides/(schemas)/objects/index.mdx
@@ -117,3 +117,42 @@ const CalculationSchema = v.pipe(
   )
 );
 ```
+
+## Extract an object field schema
+
+Valibot tries to reflect all schema information and make it available for further processing. For example you can extract a field schema from an <Link href="/api/object/">`object`</Link> schema to reuse it in other parts of your code.
+
+```ts
+import * as v from 'valibot';
+
+const PersonSchema = v.object({
+  name: v.string(),
+  age: v.number(),
+  address: v.object({
+    street: v.string(),
+    city: v.string(),
+  }),
+});
+
+const NameSchema = PersonSchema.entries.name;
+const StreetSchema = PersonSchema.entries.address.entries.street;
+```
+
+### Extraction vs composition
+
+Extracted field are reusable but depending on your use case you might want to use a composition pattern instead to improve the modularity of your code.
+
+```ts
+import * as v from 'valibot';
+
+const AddressSchema = v.object({
+  street: v.string(),
+  city: v.string(),
+});
+
+const PersonSchema = v.object({
+  name: v.string(),
+  age: v.number(),
+  address: AddressSchema,
+});
+```


fabian-hiller:
Thank you very much! I think I would move it before the `## Record schema` heading and use a level 3 heading with `###` for the sections you added. What do you think?

Bilboramix:
Yes this sounds good to me,
Also, at first i added an example with `array` but then i noticed that `array` is not mentioned at all in the objects guide.

That make sense as there's a dedicated guide about arrays on the next page. This being said i could add similar sections to the array guide as the way to access nested schema change between the two.

fabian-hiller:
Another option would be to add an extra guide explaining the accessibility of schemas in general instead of adding it everywhere.

Bilboramix:
Yeah that can make sense. I can change to this idea.
How would you name the page ?
I think about putting it in the "main concept" section of the guides, after Pipeline. But i hesitate on the name... Guessed about"Schema insights" or "Composition" (I like this one as we could also explore reusability of the `pipe` function)

fabian-hiller:
We could also call this guide "Internals" or something else in that direction. At the time of implementation, we called it the "Schema Reflection API". On the other hand, I think library authors and hardcore users will be most interested in this guide. I don't have a perfect title yet, but I recommend that I write the first version of this guide as I have the most knowledge on this regards. But I am happy to create a PR and receive feedback before publishing it.

fabian-hiller:
Let's see how it goes. Maybe we merge this PR too to add some info to the objects guide in addition.

Bilboramix:
Yeah ok. For more advanced use cases we could put a link to the next guide here maybe, to avoid too much repetitive content.
I also though that maybe a place where we have some playgrounds links could be nice and could avoid content duplication if the links are well named.
Something like this :

## How to :

[Use a composition pattern](https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh6BjCAO1XgFERlhSBlJgAsApjzgBeTADowwMCIAUGaRyjBWAc0UBKADQyxvUrp0NmbDnEGjxUlRAoArEUxiKA3vThwjfAFxw3MY2RnrecGBoqADu0AAmgSpyCsqqMOpaugYqIBoAMiJaMEKKABw6+vQAvmaMLOzwUCKoAK6k8PaqyNgiAArIUKhKoTwGXj5+pIGETsisIgACIgAeyOCkItIsIIThPlGosQkzAIwATADMACwArABsAOxlezV1FuwQm9KkENrNbQ6dSAA)

[Use an extractionpattern](https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh6BjCAO1XgGUmALAUxDI4AXkwA6CBQBWfJjAAUAb3pw4A5MFIAucWGBg+8jGI5RgrAObyAlABpx6zTbsq4YNKgDu0ACY7jegZGJjBmljb2xiDmADJ8ljA88gAc1i4AvtYMzGwccACigprc-IIicCXqYvGhwHyo1UWk2Szs8FD1AK6k8KLGqMjYfAAKyFCohoUapJWC9oRSyKx8AAJ8AB7I4KR8YiwghFmMragQO2KkEFYdqN0wR0A)

And so on...

I realize the responsibility you got as an author to pick the right descisions and to compose with what the community want, all the ideas submitted and ensure that the lib and the doc keep a good quality. Massive respect !

fabian-hiller:
> I also though that maybe a place where we have some playgrounds links could be nice to avoid content duplication.

Great idea! I agree there is room for improvement. In the long run we should add a "copy" button to the code examples and maybe also an "open in playground" button.

> I realize the responsibility you got as an author to pick the right descisions and to compose with what the community want, all the ideas submitted and ensure that the lib and the doc keep a good quality. Massive respect !

Thank you very much! Yes, it is not always an easy task. With more and more users, the responsibility also increases. That's why I have to slow down a lot of decisions. Prioritizing the right things and keeping an eye on quality will be very important for the success of the project in the long run.

fabian-hiller:
The "Reflection API" guide is on my list, but I can't tell you when I will find the time to write it. Right now I am trying to focus on important fixes and getting Valibot v1 RC ready.

fabian-hiller:
Related issue: #256

fabian-hiller:
Yes, we could. However, `entriesFromObjects` is more or less a tree-shaking enhancement utility.

PR #997, opened 2024-12-27
feat: disallow additional props in intersections

43081j:
Sets `unevaluatedProperties: false` in intersections which have `additionalProperties` consistently set to `false`.

This will enforce that to validate against the JSON schema, your input must have only properties from each of the intersecting schemas.

If any of the sub-schemas have `additionalProperties` set to a non-false value, `unevaluatedProperties` will not be set.


43081j:
Closing since it turns out the object schema is already broken in other ways

In future we may still want this for intersection between strict objects

PR #998, opened 2024-12-27
feat: support more actions in JSON schema conversion

43081j:
Adds support for the following actions in JSON schema conversion:

- `bic`
- `cuid2`
- `decimal`
- `digits`
- `empty`


packages/to-json-schema/src/convertAction.ts
```diff
@@ -191,6 +207,28 @@ export function convertAction(
       break;
     }
 
+    case 'bic':
+    case 'cuid2':
+    case 'decimal':
+    case 'digits': {
+      jsonSchema.pattern = typeToPatternMap[valibotAction.type].source;
```
fabian-hiller, line 214: Is `typeToPatternMap` necessary? These actions should include the regex at the `.requirement` key.

fabian-hiller:
I first need to check that the `.source` of a regex assigned to the `.pattern` does not produce any unexpected behaviour or bugs. At the moment I suspect that this implementation is buggy because `.source` does not contain any regex flags and JSON schema does not support regex flags as far as I know. Flags like `i` for case insensitivity would therefore lead to bugs if missing.

43081j:
no worries

you are right that the flags will be lost during conversion. not a lot to be done about that

we could throw if _any_ flags are set, or we could accept that the conversion is lossy

fabian-hiller:
I am not sure how to proceed. So far the only important flag we are using is `i` for case insensitivity. We could remove this flag by rewriting these regular expressions so that they can simply be passed to `.pattern`. The drawback would be that they might become a bit more complex to maintain, and I expect a small increase in bundle size. Supporting only the regular expressions without the `i` flag could be another option, but this is a buggy option as we may introduce bugs when changing a regex.

43081j:
I think we do both:

- throw when flags are set, or don't set the pattern
- try rewrite some of the simpler patterns to no longer need flags

if it turns out this isn't possible (especially since we use `u` a lot), maybe we just cant support these actions in conversion?

we don't support any of them yet, so throwing seems to make more sense than "best effort"

fabian-hiller:
I think we can ignore the `u` flag for JSON schema. This flag usually has no effect and we only use it because it seems to be a good practice for newer regular expressions.

I will look into rewriting some or all of the regular expressions. However, our v1 RC release has a higher priority. So it may be a while before we can move forward with this PR.

andersk:
Perhaps we can translate the `i` flag using a [modifier](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Modifier): `/foo/i` → `/(?i:foo)/`? This isn’t [recommended](https://json-schema.org/understanding-json-schema/reference/regular_expressions) but is probably better than ignoring it.

fabian-hiller:
Is the idea to wrap the entire `regex.source` in ``` `(?i:${regex.source})` ``` when `regex.flags` includes `'i'`?

fabian-hiller:
v1.0.0-beta.4 of `@valibot/to-json-schema` is available

43081j:
Thanks for getting this over the line 🙏

PR #1000, opened 2024-12-27
feat: treat objects as loose in JSON schema intersections

43081j:
An alternative to #997 

Instead of doing smarter detection of if to use `unevaluatedProperties` or not, we simply loosen the type here as if it was a loose object. This results in `additionalProperties: true` for `object` schemas which are used in an intersection, but outside of intersections, leaves the schema as-is now.

Once we can support the newer draft of JSON schema, we should do #997 instead

### DIFF
--- a/packages/to-json-schema/src/convertSchema.test.ts
+++ b/packages/to-json-schema/src/convertSchema.test.ts
@@ -703,7 +703,6 @@ describe('convertSchema', () => {
               foo: { type: 'string' },
             },
             required: ['foo'],
-            additionalProperties: false,
           },
           {
             type: 'object',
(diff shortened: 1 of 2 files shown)


andersk:
I thought what we agreed on is

- #1001

43081j:
Seems a misunderstanding 

That pr will alter how regular object schemas are converted unnecessarily. They should still have `additionalProperties: false` when they're not part of an intersection

This pr will only loosen the type if used in an intersection, which is the problem the op was having

andersk:
> This behaviour then matches more closely with the typescript types too,

That’s not true; both Valibot’s runtime and TypeScript types match more closely with #1001.

```ts
const o = { foo: 1, bar: 2 };
v.parse(v.object({ foo: v.number() }), o);  // ok
const x: { foo: number } = o;  // ok
```

43081j:
Why is that the case when we have loose objects? 

Loose object schemas and object schemas both loosely type as if they're a partial? 

I think some of the confusion here is that there is a loose type and an object type. Also, I am tackling this problem so it would have been helpful to have the feedback instead of an alternative pr. I'd always be happy to update to what is agreed

andersk:
The only difference between `v.object` and `v.looseObject` is that [`v.looseObject`](https://valibot.dev/api/looseObject/) includes unknown properties in the output value (without including them in the TypeScript type) rather than filtering them out. There is no difference in which objects are accepted.

43081j:
I understand

- `object` is actually a loose object
- `strictObject` is a strict object

So we currently output the wrong schema for `object` since we treat it as if it was strict 

It was meant to be loose all along, i.e allow additional properties

43081j:
Closing since the schema is already being converted incorrectly unrelated to intersections

Will leave you to it

fabian-hiller:
The difference is the following:
- `object` ignores and "removes" unknown properties
- `looseObject` includes unknown properties and reflects this in the type signature
- `strictObject` returns an issue if unknown properties are detected

PR #1001, opened 2024-12-27
toJsonSchema: Only disallow additionalProperties for strictObject

### LINKED ISSUE #990: toJsonSchema generates unsatisfiable schema for intersect of objects
## Problem

When using `toJsonSchema` with `intersect` of objects, it generates a JSON Schema that cannot be satisfied due to conflicting `additionalProperties: false` constraints.
I understand that Valibot is not primarily designed for JSON Schema compatibility, but this makes the conversion feature unusable for any JSON Schema interoperability scenarios.

## Reproduction
```typescript
import { object, string, number, intersect } from 'valibot';
import { toJsonSchema } from '@valibot/json-schema';

const schema = intersect([
  object({ foo: string() }),
  object({ bar: number() }),
]);
console.log(JSON.stringify(toJsonSchema(schema), null, 2));
/*
{
  "allOf": [
    {
      "type": "object",
      "properties": {
        "foo": {
          "type": "string"
        }
      },
      "required": [
        "foo"
      ],
      "additionalProperties": false
    },
    {
      "type": "object",
      "properties": {
        "bar": {
          "type": "number"
        }
      },
      "required": [
        "bar"
      ],
      "additionalProperties": false
    }
  ]
}
*/
```

## Current Behavior
The generated JSON Schema includes `additionalProperties: false` for each object in `allOf`, making it impossible to satisfy both constraints simultaneously.

## Expected Behavior
The generated JSON Schema should allow objects that have both properties, matching Valibot's runtime validation behavior.
Possible solutions:
- Remove `additionalProperties: false` from individual objects in allOf
- Add `additionalProperties: false` only at the parent level

### versions
```
$ npm list | grep valibot
...
├── @valibot/to-json-schema@1.0.0-beta.3
└── valibot@1.0.0-beta.9
```

andersk:
`additionalProperties: true` was preventing validation of valid intersected objects.

- Fixes #990.

### DIFF
--- a/packages/to-json-schema/CHANGELOG.md
+++ b/packages/to-json-schema/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to the library will be documented in this file.
 - Change Valibot peer dependency to v1.0.0
 - Change extraction of default value from `nullable`, `nullish` and `optional` schema
 - Change `force` to `errorMode` in config for better control (issue #889)
+- Change `additionalProperties` for `object` and `looseObject` schema (pull request #1001)
 
 ## v0.2.1 (September 30, 2024)
 
--- a/packages/to-json-schema/src/convertSchema.test.ts
+++ b/packages/to-json-schema/src/convertSchema.test.ts
@@ -275,7 +275,6 @@ describe('convertSchema', () => {
           key4: { anyOf: [{ type: 'number' }, { type: 'null' }] },
         },
         required: ['key1', 'key3'],
-        additionalProperties: true,
       });
     });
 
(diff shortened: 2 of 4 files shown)


fabian-hiller:
Thank you! I plan to review and merge this PR in the next days.

fabian-hiller:
According to my research, `additionalProperties` is `true` by default. So I will remove `additionalProperties` for the `looseObject` schema to reduce the bundle size.

andersk:
Sure, sounds good.

fabian-hiller:
v1.0.0-beta.4 of `@valibot/to-json-schema` is available

PR #1008, opened 2025-01-08
docs: improve nullable/nullish type definitions

frenzzy:
Updates the JSDoc documentation for `nullable()` and `nullish()` functions to better illustrate their type behavior. 

Changes:
- Added TypeScript union types (`T | null` vs `T | null | undefined`) to clearly show the difference

This change makes it easier for developers to quickly understand (especially non english speaking) when to use each function based on their TypeScript type requirements.

P.S.: Is it only me who keeps googling "nullable vs nullish difference" every other week? 😅

### DIFF
--- a/library/src/schemas/nullable/nullable.ts
+++ b/library/src/schemas/nullable/nullable.ts
@@ -44,8 +44,7 @@ export interface NullableSchema<
 }
 
 /**
- * Creates a nullable schema that allows null values but not undefined.
- * Type: T | null
+ * Creates a nullable schema that allows null values.
  *
  * @param wrapped The wrapped schema.
  *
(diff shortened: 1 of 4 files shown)


fabian-hiller:
I don't think I want to merge this PR. I agree that your changes make it clearer, but the current approach is consistent across all of our actions. In simple words: We use almost exactly the same description template for every JSDoc, and I think I prefer consistency in this case. If we decide to change the JSDocs it should be consistent across all actions and probably all other functions too.

frenzzy:
Ok, I agree it's good to have consistent docs, i.e., it would be better to add examples to all the functions.

But what about the latest code suggestion with just minor wording improvements?

> * Creates a nullish schema that allows null and undefined values.
> * Creates a nullable schema that allows null values.

Alternatively, for consistency, we could just add parentheses (they don't have to be everywhere):
> * Creates a nullish schema (T | null | undefined).
> * Creates a nullable schema (T | null).

fabian-hiller:
Your ideas are good. I will come back to this and other similar issues and PRs after V1 is available.

Issue #983, opened 2024-12-15
Runtime behavior of `optional`/`undefinedable` doesn’t match TS with exactOptionalPropertyTypes


andersk:
#385 was resolved by adding `v.undefinedable` with different typing from `v.optional`:

```ts
import * as v from "valibot";

const optional_schema = v.object({ field: v.optional(v.number()) });
type OptionalOutput = v.InferOutput<typeof optional_schema>;
// { field?: number }

const undefinedable_schema = v.object({ field: v.undefinedable(v.number()) });
type UndefinedableOutput = v.InferOutput<typeof undefinedable_schema>;
// { field: number | undefined }
```

However, this typing is inconsistent with the runtime behavior: at runtime, both schemas accept and return both of `{}` and `{ field: undefined }`. This means that both schemas break type safety with `tsc --strict --exactOptionalPropertyTypes`.

```ts
const a = v.parse(optional_schema, { field: undefined });
if ("field" in a) { // narrows a.field to string
  const n: number = a.field; // oops, this is undefined
}

const b = v.parse(
  v.union([undefinedable_schema, v.object({ bogus: v.string() })]),
  {}
);
if (!("field" in b)) { // narrows b to { bogus: string }
  const s: string = b.bogus; // oops, this is undefined
}
```

Either the types should be adjusted to match the runtime behavior, or the runtime behavior should be adjusted to match the types.

Zod does the former:

```ts
import { z } from "zod";

const optional_schema = z.object({ field: z.optional(z.number()) });
type OptionalOutput = z.output<typeof optional_schema>;
// { field?: number | undefined }
```

andersk:
`v.unknown` has the same problem.

```ts
import * as v from "valibot";

const unknown_schema = v.object({ field: v.unknown() });
type UnknownOutput = v.InferOutput<typeof unknown_schema>;
// { field: unknown }

const b = v.parse(
  v.union([unknown_schema, v.object({ bogus: v.string() })]),
  {}
);
if (!("field" in b)) { // narrows b to { bogus: string }
  const s: string = b.bogus; // oops, this is undefined
}
```

In Zod, `unknown` implies optionality:

```ts
import { z } from "zod";

const unknown_schema = z.object({ field: z.unknown() });
type UnknownOutput = z.output<typeof unknown_schema>;
// { field?: unknown }

const b = z.union([unknown_schema, z.object({ bogus: z.string() })]).parse({});
if (!("field" in b)) { // safely avoids narrowing b
[...]

PR #1011, opened 2025-01-11
Feat: Add `exactOptionalProperties` config match `exactOptionalPropertyTypes` in TypeScript

fabian-hiller:
So far, this is just a draft to further investigate our final decision.

Fix: #983

### DIFF
--- a/library/src/schemas/object/object.ts
+++ b/library/src/schemas/object/object.ts
@@ -110,53 +110,82 @@ export function object(
         // The reason for this decision is that it reduces the bundle size, and
         // we also expect that most users will expect this behavior.
         for (const key in this.entries) {
-          // Get and parse value of key
+          // Get value of key
           const value: unknown = input[key as keyof typeof input];
-          const valueDataset = this.entries[key]['~run']({ value }, config);
 
-          // If there are issues, capture them
-          if (valueDataset.issues) {
-            // Create object path item
-            const pathItem: ObjectPathItem = {
-              type: 'object',
-              origin: 'value',
-              input: input as Record<string, unknown>,
-              key,
-              value,
-            };
-
-            // Add modified entry dataset issues to issues
-            for (const issue of valueDataset.issues) {
-              if (issue.path) {
-                issue.path.unshift(pathItem);
-              } else {
-                // @ts-expect-error
-                issue.path = [pathItem];
-              }
-              // @ts-expect-error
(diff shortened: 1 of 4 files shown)


andersk:
- This is _too_ strict at runtime. This fails and should succeed:

  ```ts
  v.setGlobalConfig({ exactOptionalProperties: true });
  v.parse(v.object({ p: v.optional(v.undefined()) }), { p: undefined });
  ```

  Similarly for `v.optional(v.undefinedable(…))` or `v.optional(v.nullish(…))` or `v.optional(v.void())` or `v.optional(v.unknown())` or `v.optional(v.union([…, /* any of the above */]))` or `v.optional(v.intersect([/* any of the above */]))` or `v.optional(v.lazy(() => /* any of the above */))`.

- The dependence on global configuration is concerning, because it will lead to unexpected behavior in a library using Valibot within a project that also uses Valibot; the configuration from one will leak into the other.

  If we want to support both (a) property could be missing and (b) property could be missing or `undefined`, then I think we should provide different functions for (a) and (b) rather than switching on ambient configuration. Maybe `v.optional` and `v.optionalish`?

andersk:
- This doesn’t address the other half of #983, where `v.object({ field: v.undefinedable(v.number()) })` incorrectly accepts the value `{}` for the type `{ field: number | undefined }`.

fabian-hiller:
> This is too strict at runtime. This fails and should succeed

I am aware of that. That's what I mentioned at the end of [this comment](https://github.com/fabian-hiller/valibot/issues/983#issuecomment-2585361256). It is fixable, but requires additional code that results in an even larger bundle size.

> The dependence on global configuration is concerning

Yes, that is definitely something we should think about. But you don't have to do it globally. This config can be set in several other places. For example, when calling [`parse`](https://valibot.dev/api/parse/) or [`safeParse`](https://valibot.dev/api/safeParser/) via the third argument or when using [`config`](https://valibot.dev/api/config/).

> This doesn’t address the other half of #983, where `v.object({ field: v.undefinedable(v.number()) })` incorrectly accepts the value `{}` for the type `{ field: number | undefined }`.

True, great catch. Thank you!

What do you think about my last comment in #983? Would you rather add a lot of code to cover all these edge cases or would you prefer a simpler and more limited approach with the goal of a small bundle size? What is the main reason you prefer Valibot over other solutions?

PR #1013, opened 2025-01-13
Change `object` and `optional` to support exact optional props

### LINKED ISSUE #983: Runtime behavior of `optional`/`undefinedable` doesn’t match TS with exactOptionalPropertyTypes

fabian-hiller:
FIx #983


andersk:
Seems mostly good, but optional defaults are broken:

```js
> v.parse(v.object({ a: v.optional(v.string(), "foo") }), {})
{}
```

(expected `{ a: 'foo' }`)

fabian-hiller:
I think this is fixed now. I also added support for the `nullish` schema. This PR increases the bundle size of `object` by 100 bytes. If those 100 bytes fix all our problems, it should be fine.

The new behavior (these are breaking changes) of this PR is as follows:
- `optional` no longer allows `undefined` as a value, it only allows missing entries
- Users must explicitly allow `undefined` by using `undefinedable` or `union` inside `optional`
- `object` now distinguishes between missing and `undefined` values and only allows missing entries if the outer schema of an entry is `optional` or `nullish`
- Any other schema will return an issue for missing object keys

If we think this is the right solution for Valibot, I would also update these changes in `looseObject`, `strictObject` and `objectWithRest`. I would also update the tests to match the new behavior.

fabian-hiller:
I have implemented your feedback and removed support for missing keys when using the `nullish` schema. The idea behind this change was to make our API more consistent. The only way to make an object entry optional now is to explicitly use `optional` as the outermost schema of the entry.

Furthermore, I refactored the implementation of the `object` schema and optimized its bundle size. These changes lowered the addition required bytes from about 100 to 77.

I think this PR is going in the right direction. My next steps are to apply these changes to the entire codebase and update our tests. Once this PR is done, I will probably create a GitHub discussion to explain these changes and reach out to the community and ask for feedback.

fabian-hiller:
Please discuss with use in #1022.

fabian-hiller:
I followed @xcfox's idea: https://github.com/fabian-hiller/valibot/discussions/1022#discussioncomment-11873529

fabian-hiller:
Valibot v1.0.0-beta.13 is available

PR #1023, opened 2025-01-18
feat: create `entriesFromObjects` utils

### LINKED ISSUE #1017: Side effects when using the spread operator to merge objects
Currently, if we use the spread operator (`...`) to [merge objects](https://valibot.dev/api/object/#merge-several-objects) as shown below, it causes side effects and cannot be tree-shaken:

```js
const ParentSchema = v.object({
  foo: v.string(),
});

const LoginSchema = v.object({
  ...ParentSchema.entries, // <- side effect
  email: v.pipe(v.string(), v.email()),
  password: v.pipe(v.string(), v.minLength(8)),
});
```

Playground
https://stackblitz.com/edit/node-gq8gcqea?file=index.js

The reason is that bundlers like Rollup set [`propertyReadSideEffects`](https://github.com/rollup/rollup/blob/1f1e6c7929a3b1d47ca304882b503bbda12fa617/src/utils/options/options.ts#L115) to `true` by default, which means the spread operator is treated as a side effect.

We can set the [`treeshake.preset`](https://rollupjs.org/configuration-options/#treeshake-preset) to `"smallest"` to disable `propertyReadSideEffects` and make it tree-shakeable but this may cause other issues.

## Solution

I wonder if we could have a `v.merge` function with a `@__NO_SIDE_EFFECTS__` annotation for merging objects, making it possible for bundlers to tree-shake it:

```js
// @__NO_SIDE_EFFECTS__
function merge(a, b) {
  return v.object({
    ...a.entries,
    ...b.entries,
  });
}
```

Usage

```js
const ParentSchema = v.object({
  foo: v.string(),
});

const LoginSchema = v.merge(
  ParentSchema,
  v.object({
    email: v.pipe(v.string(), v.email()),
    password: v.pipe(v.string(), v.minLength(8)),
  })
);
```

It would be even better if the `v.merge` function supported an unlimited number of arguments to merge multiple objects at once.

### Alternative: Supports input array in `v.object`

Also, I am wondering if we can support input an array of `entries` in `v.object`, like:

```js
// @__NO_SIDE_EFFECTS__
function object(entries, message) {
  if (Array.isArray(entries)) {
    return v.object(
      entries.reduce((acc, current) => ({ ...acc, ...current }), {}),
      message
    );
  }
  // ... original implementation
}
```

Usage

```js
const ParentSchema = v.object({
  foo: v.string(),
});

const LoginSchema = v.object([
  ParentSchema.entries,
  {
    email: v.pipe(v.string(), v.email()),
    password: v.pipe(v.string(), v.minLength(8)),
  }
]);
```

It can also be tree-shaken as the spread operator is not used.

---
Thanks @antfu for helping clarify this issue.

serkodev:
Resolve #1017

This PR create an new util `entriesFromObjects`. You can use it extract entries from multiple objects for prevent using spread operator (`...`) and `.entries` (property access behavior) for better tree-shaking result.

Original extends `v.object`

```js
const ParentSchema = v.object({
  foo: v.string(),
});

const Schema = v.object({
  ...ParentSchema.entries,
  bar: v.string(),
});
```

With new `entriesFromObjects`

```js
const ParentSchema = v.object({
  foo: v.string(),
});

const Schema = v.object(
  v.entriesFromObjects(
    ParentSchema,
    v.object({
      bar: v.string(),
    })
  )
);
```


fabian-hiller:
Thanks for creating this PR!

I think your second code example in was meant to be:

```ts
const ParentSchema = v.object({
  foo: v.string(),
});

const Schema = v.object(
  v.entriesFromObjects(
    ParentSchema,
    v.object({
      bar: v.string(),
    })
  )
);
```

fabian-hiller:
Should we add a TypeScript overload signature that forces the user to pass at least two object schemas to `entriesFromObjects`?

```ts
export function entriesFromObjects<
  const TSchema1 extends Schema,
  const TSchema2 extends Schema,
  const TSchemas extends Schema[],
>(
  schema1: TSchema1,
  schema2: TSchema2,
  ...schemas: TSchemas
): MergedEntries<TSchemas>;
```

serkodev:
Thanks for pointing out the mistake of the example 😆 just fixed.

> Should we add a TypeScript overload signature that forces the user to pass at least two object schemas to entriesFromObjects?

A great suggestion about forcing the user to pass a minimum of object schemas.

How about only enforcing at least 1 object schema?
I’m not sure if there might be some edge cases where someone uses this utility on a single object schema to avoid accessing the `.entries` property directly.

fabian-hiller:
I plan to finalize and merge this PR in the next one or two weeks.

fabian-hiller:
Sorry for my delay. I am busy with personal stuff. Hope to merge this after v1 RC is out.

serkodev:
> Sorry for my delay. I am busy with personal stuff. Hope to merge this after v1 RC is out.

No problem, take your time. 😁 Thanks!

fabian-hiller:
I changed the implementation a bit to better match our usual API design. Hope thats ok. Reach out if my changes cause any problems.

fabian-hiller:
v1.0.0-rc.1 is available

PR #1030, opened 2025-01-23
[WIP]: add `toSnakeCase` & `toSnakeCaseKeys` actions

EltonLobo07:


EltonLobo07:
@fabian-hiller I'll have to change the algorithm to transform a string to its snake case version, as I have missed some important edge cases. I'll need your feedback on the [new algorithm](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABFApgZygHgCoD4AUKAHgA4rQoAmAXItgDSIBO5KMAblbdgJSIDeAKESIYwRIVLlUlRAEIAvAuasOVPkJEiICNHAA2KAHQomTOE3wADYmQo1EAEn63pVAL6MWENpwfPvXw8rHgBuYS0WKBAmJGAAQ300FHCRdwiomKQoJhAUwXTBUEhYBEQ0MHiAaxQAYXjk-BgwEhAoWgwmZoBzHg6cnoEInTAMFTR+rrBugG0AXURledTEQyhEAHcGgAUWdlqACwBVEjImeuTFxASk-JFgCwkRsYgDxDhxZtaoIwGAW3wPA0EW0unW+jgG1MFzqb2Ur1+cAAMpDoQ0UICVqDRusYGgAOoHGCoADKJHiPiuCP+gMWSkQACIGVjEM9cWgSShyUx4lBHso8YTiSgyRSUIgAD4S1lw+kMgD6DMl0tedOUDIAtEqpTK1YyjMyQaywaI0CczjCqbL4QdEea0Y0+AAyJ26xTKCFQ87ow4ssQSPGc7m8izArRaf34FhoIyGaZQN64RAABkQLvGM2jsZQ8beGsQAEYFu7GYqw+HIugjK00Ad8AqGWEjWkje5EChbqJxE0zacHShy+HI1m490E4gk6n09HM1XR+P80X5HLFWnXXItmhdih9sc+97koOK1ma3WG02K4h0peTyBa-hPf3DhetG2O5dNMeq6f8K8Xy3w03bdd3tA9xQFXsLXRFZry7CQRxzMdExTNcMwQ3NEEXBYlHVMshnDE84BITEIlgzJYnGIwACs4GaetG3CQo2VZdEJkQGZXHsSYekYL42m46Y5nmK4ZgiAB6MTyigeIwEoeImEoCIZgZA4OwheUNgsfRKAZRgVLUuBNi0nS5kYS9wwk1ZUSYFjkiU-T9HUzSmG03TGQACQMxB8WMhlTPMizJIfazEAAakQEB9z4CB0XsmK-g7eUYuSNyGXijsYT8syAsCliEv0WyUHs8k0Bi-QkvRVLtgaMrMv8nKREskqysK+yagAI3idqKpSvSOq6jVkpQLKGq0Sz+va1qRGUyKznlCpqhQHrhr0o5tm2ABRAAleUSQAOQAQQAaQ2+VagOkkNpGyzZtMcpKhqKbEEsqhunFIa0Hs1Ksq+vStB+6aGVQDBUq0YH1hEAH2IcpzfL01THLgeUfJckz6HshHYdR1LMcMkRnNc0yMYM1L5VxqHlPJ+GDMVInAeaYlSflBmoHlWn0fp+VABNyQB4P-lABPOAQFSgBJRBecQQXhbp6Hvpl5S-hgIh5Q+eaoH5ww0FSgBZJWAHlxBJdXNYphl4nlABT1L4gAUdNuB2qo6R5RqfnUodp3oCMV3Ta6iB5UoFBgFSjVQ4OgAhWpQ4AEQ2gAxUOtTmQQ5nCQQ1kQRJ9BhLcGmSWRlByPJwgeGzfxNDipHsXiWjaBYPkKtByyznPqrQfOrhb1i2479NwckOwZEYBaahhJpa6gIFGMESMu+SXP2-UfDjVGAxjAhbp6wOxzG8QEr87kBiCkEIA).

fabian-hiller:
I will try to review it next week.

EltonLobo07:
There are some tricky problems that might have broken the type safety. To make sure the type safety is maintained and to make sure the TS type checker doesn't get overwhelmed, I have loosened the types a bit while reimplementing the action, and I think it is the correct path to take. I'll reply with a detailed message tomorrow which explains the decisions taken.

fabian-hiller:
Sounds good. I can try to review it on the weekend and give feedback on your code.

EltonLobo07:
Ideally, when dealing with duplicate keys, this is what I expected:
```ts
const Schema = v.pipe(
    v.object({FooBar: v.string(), fooBar: v.boolean()}),
    v.toSnakeCase()
);

const input = {FooBar: "", fooBar: false}; // type: {FooBar: string, fooBar: boolean}
const output = v.parse(Schema, input); // type: {foo_bar: string}
```
While working on the action, I realized that this expectation was incorrect because:
1. The key order of an object could be anything (I know the key order is based on the key creation order but not always). So how can we know which key will be transformed first? We can't. Now, I have a workaround that only solves this problem. Even if we implemented that, we still can't solve problem number 2.
2. Let's assume we can determine the runtime key order. The next challenge is to sync the runtime key order with the type key order which is not possible because `{FooBar: string, fooBar: boolean}` and `{fooBar: boolean, FooBar: string}` both types can accept the same object but they will produce a different output type once the keys are transformed. The output types would be `{foo_bar: string}` and `{foo_bar: boolean}` respectively. This will break type safety.
    ```ts
    const Schema = v.pipe(
      v.object({ FooBar: v.string(), fooBar: v.boolean() }),
      v.transform<
        { FooBar: string; fooBar: boolean },
        { fooBar: boolean; FooBar: string }
      >((v) => v),
      v.toSnakeCase(),
    );
    
    const output = v.parse(Schema, {
      FooBar: '',
      fooBar: false,
    }); // value: {foo_bar: ""}, type: {foo_bar: boolean}
    ```
Additionally, the `keyof` type operator is not ordered in any way. We can't change the random key iteration order at the type level.

Since it is difficult to sync the type and runtime key order, we avoid it. This is where we add a bit of inconvenience to maintain type safety. At runtime, the last used duplicate key's value will be in the output. At the type level, the duplicate key values will form a union.
```ts
const Schema = v.pipe(
  v.object({ FooBar: v.string(), fooBar: v.boolean() }),
  v.toSnakeCase(),
);

const output = v.parse(Schema, {
  FooBar: '',
  fooBar: false,
}); // value: {foo_bar: false}, type: {foo_bar: string | boolean}
```

EltonLobo07:
The other issue is related to the key select list. When we have two keys, the number of possible select list tuples is four.
```ts
const Schema = v.pipe(
  v.object({ FooBar: v.string(), fooBar: v.boolean() }),
  // union of 4 tuples:
  // ["fooBar"] | ["FooBar"] | ["fooBar", "FooBar"] | ["FooBar", "fooBar"]
  v.toSnakeCase(["fooBar", "FooBar"]),
);
```
The number of possible tuples increases very rapidly as the number of keys in the object increases. This slows down the type checker. In some cases, the type checker never finishes calculating the possible types.

I recommend making the select list type loose.
```ts
type SelectedStringKeys<T extends ObjectInput> = (keyof T & string)[];
```

EltonLobo07:
Do you think it is a good idea to expose an additional action related to this PR? 
`toSnakeCase` - an action that converts a string to its snake case form
`objectToSnakeCase` - this PR's action

fabian-hiller:
Sorry that I don't have a lot of time right now. I will try to review more PRs in about two weeks. Yes, we could provide a `toSnakeCase` action for strings and something like a `toSnakeCaseKeys` actions for objects.

EltonLobo07:
# Summary
Since there is a lot to discuss, I have summarized everything I know until now.

## API usage

### `toSnakeCase`
```ts
const Schema = v.pipe(v.string(), v.toSnakeCase());

const output = v.parse(Schema, "hello world");

/*
  output's 
    type : string
    value: hello_world
*/
```

### `toSnakeCaseKeys`

#### No selected keys
```ts
const Schema = v.pipe(
  v.object({
    fooBar: v.literal('fooBar'),
    'hello world': v.literal('hello world'),
    'HELLO WORLD': v.literal('HELLO WORLD'),
  }),
  v.toSnakeCaseKeys(),
);

const output = v.parse(Schema, {
  fooBar: 'fooBar',
  'hello world': 'hello world',
  'HELLO WORLD': 'HELLO WORLD',
});

/*
  output's 
    type : { foo_bar: "fooBar", hello_world: "hello world" | "HELLO WORLD" }
    value: { foo_bar: "fooBar", hello_world: "HELLO WORLD" }
*/
```
> The latest key's value is used when dealing with duplicate keys. A key is said to be the latest if it appears at the end 
      while iterating over the object's keys using `Object.keys`

#### Selected keys
```ts
const Schema = v.pipe(
  v.object({
    fooBar: v.literal('fooBar'),
    'hello world': v.literal('hello world'),
    'HELLO WORLD': v.literal('HELLO WORLD'),
  }),
  v.toSnakeCaseKeys(["hello world", "HELLO WORLD"]),
);

const output = v.parse(Schema, {
  fooBar: 'fooBar',
  'hello world': 'hello world',
  'HELLO WORLD': 'HELLO WORLD',
});

/*
  output's 
    type : { fooBar: "fooBar", hello_world: "hello world" | "HELLO WORLD" }
    value: { fooBar: "fooBar", hello_world: "HELLO WORLD" }
*/
```

> The select list's type that is passed to `toSnakeCaseKeys` is loose on purpose. For more information, read [this comment](https://github.com/fabian-hiller/valibot/pull/1030#issuecomment-2672582845)

## `toSnakeCaseKeys` edge cases

### All duplicates are optional
Since all duplicates are optional, the transformed key is also optional
```ts
const Schema = v.pipe(
  v.object({
    'hello world': v.optional(v.literal('hello world')),
    'HELLO WORLD': v.optional(v.literal('HELLO WORLD')),
  }),
  v.toSnakeCaseKeys(),
);

const output = v.parse(Schema, { });

/*
  output's 
    type : { hello_world?: "hello world" | "HELLO WORLD" }
    value: { }
*/
```
### Not all duplicates are optional
There will always be a key-value pair, so the transformed key is not optional
```ts
const Schema = v.pipe(
  v.object({
    'hello world': v.optional(v.literal('hello world')),
    'HELLO WORLD': v.literal('HELLO WORLD'),
  }),
  v.toSnakeCaseKeys(),
);

const output = v.parse(Schema, { "HELLO WORLD": "HELLO WORLD" });

/*
  output's 
    type : { hello_world: "hello world" | "HELLO WORLD" }
    value: { hello_world: "HELLO WORLD" }
*/
```

### At least one duplicate is readonly
Since it is better not to rely on the order of the keys in the type system, we don't really know if the readonly key was the actual key that is present in the output. Hence, the action marks the transformed key readonly
```ts
const Schema = v.pipe(
  v.object({
    'hello world': v.pipe(v.literal('hello world'), v.readonly()),
    'HELLO WORLD': v.literal('HELLO WORLD'),
  }),
  v.toSnakeCaseKeys(),
);

const output = v.parse(Schema, {
  'hello world': 'hello world',
  'HELLO WORLD': 'HELLO WORLD',
});

/*
  output's 
    type : { readonly hello_world: "hello world" | "HELLO WORLD" }
    value: { hello_world: "HELLO WORLD" }
*/
```

### Duplicate number-string keys
Even though the example below returns a value, this is an error in the type world. The returned type will be `never`
```ts
const Schema = v.pipe(
  v.object({
    1: v.literal(1),
    ' 1 ': v.literal(' 1 '),
  }),
  v.toSnakeCaseKeys(),
);

const output = v.parse(Schema, { 1: 1, " 1 ": " 1 " });

/*
  output's 
    type : never
    value: { 1: " 1 " }
*/
```
### Unexplored
@fabian-hiller I haven't explored working with the schemas mentioned below, but we will have to decide how to work with them. If you can help me decide what should be the type of the output for the following cases, that would be great.

#### `looseObject`
```ts
const Schema = v.pipe(
  v.looseObject({
    "hello world": v.literal("hello world")
  }),
  v.toSnakeCaseKeys(),
);

const output = v.parse(Schema, { "hello world": "hello world", "hello World": 123 });

// output's value: { hello_world: 123 }
```

#### `objectWithRest`
```ts
const Schema = v.pipe(
  v.objectWithRest(
    {
      'hello world': v.literal('hello world'),
    },
    v.boolean(),
  ),
  v.toSnakeCaseKeys(),
);

const output = v.parse(Schema, {
  'hello world': 'hello world',
  'hello World': true
});

// output's value: { hello_world: true }
```

EltonLobo07:
I think processing the keys in the entries object of `looseObject` and `objectWithRest` schemas maintain type safety since the keys that are processed later have the chance to override the values assigned to the same key.
```ts
const Schema = v.pipe(
  v.looseObject({
    "hello world": v.literal("hello world")
  }),
  v.toSnakeCaseKeys(),
);

const output = v.parse(Schema, { "hello world": "hello world", "hello World": 123 });

/*
  If the key "hello world" is processed last because 
  it is a part of the entries object, the output's type and value match!
*/ 

// output's type  : { hello_world: "hello world" } 
// output's value : { hello_world: "hello world" }
```

yslpn:
Hi @EltonLobo07 . Sorry for the delay. Are you interested in working on this PR further? If so, we need to resolve the conflicts, since we already have toSnakeCase. And fix all comments from AI codereview or close them

Also, I don't quite understand the use case: when and why toSnakeCaseKeys might be needed? Could you provide a real-world, useful example? We'll also need to add documentation for the website. What are the limitations and how are edge cases handled?

And I don't want to release an asymmetric API. If we're convinced we need toSnakeCaseKeys, we'll need to create analogs for other similar functions.

toCamelCaseKeys()
toSnakeCaseKeys()
toKebabCaseKeys()
toPascalCaseKeys()

Issue #1018, opened 2025-01-17
Proposal: Provide tree-shakable `v` shorthand for valibot functions instead of namespace imports


k257588376:
The current approach is where you have to manually import everything using `import * as v from 'valibot'`.
While this method works, it introduces a few inconveniences. First, you lose the convenience of having `v` as a shorthand globally available with autocompete. using `import * as v` lacks autocomplete support and requires you to manually type `import * as v from 'valibot'`.

Modern JavaScript bundlers like Webpack, Rollup, and ESBuild fully support tree shaking with the `export * as ... from '...'` syntax in ECMAScript Modules. When using `export * as ... from '...'`, a bundler can analyze the module and prune out unused exports even if they are imported under a namespace. For instance, lets rename [index.ts](https://github.com/fabian-hiller/valibot/blob/main/library/src/index.ts) to valibot.ts:

```ts
// File: src/valibot.ts
export * from './actions/index.ts';
export * from './methods/index.ts';
export * from './regex.ts';
export * from './schemas/index.ts';
export * from './storages/index.ts';
export * from './types/index.ts';
export * from './utils/index.ts';

// File: src/index.ts
export * from "./valibot.ts"
export * as v from "./valibot.ts"
```

```ts
// now the user of the library can use `v` with autocompletion and fully tree-shakable!
import { v } from "valibot"
const EmailSchema = v.pipe(v.string(), v.email());
```

[...]

PR #1041, opened 2025-01-30
[WIP] Support browserslist + initial size limit and es-check scripts

belgattitude:
Following that discussion: https://github.com/fabian-hiller/valibot/issues/1018#issuecomment-2623207906, I'd like to take a look a bundle performance.

Before doing so I'd like to get a baseline ensured. Here's the changes I propose

## Features

- [x] Add `pnpm check.size` with a temporary size-limit.ts
- [x] Add `pnpm check-dist` to ensure bundled code is es2022 compliant
- [x] Add initial browserslist configuration to ensure bundled code is browser compliant

### 1. Browserslist

I feel we should decide what browserslist compat we want. Locally I've tested impact on the bundle size:

By default `es2022,chrome109,edge131,firefox128,ios15.6,opera113,safari18.1` will produce a bundled `10.34 kB`

```
✔ Adding to empty webpack project
  
  Size limit: 10.4 kB
  Size:       10.34 kB with all dependencies, minified and brotlied
```

For example if we want to support more browsers `es2022,chrome96,edge90,firefox90,ios14,opera84,safari14`The bundle would be

```
✔ Adding to empty webpack project
  
  Package size limit has exceeded by 181 B
  Size limit: 10.4 kB
  Size:       10.58 kB with all dependencies, minified and brotlied
```

### 2. Evaluate `import { v } from valibot`

From this P/R once we have a baseline, I'm open to tune the tsup config to emit separate files and try out mulitple scenarios.

Open to discuss...

### DIFF
--- a/library/.browserslistrc
+++ b/library/.browserslistrc
@@ -1,0 +1,8 @@
+# Supported browsers
+defaults
+#chrome >= 96
+#firefox >= 90
+#edge >= 90
+#safari >= 15
+#ios >= 15
+#opera >= 84
--- a/library/.size-limit.ts
+++ b/library/.size-limit.ts
@@ -1,0 +1,10 @@
+import type { SizeLimitConfig } from 'size-limit';
+
+module.exports = [
+  {
+    name: `import * as v from 'valibot' (ESM)`,
+    path: ['dist/index.js'],
+    import: '*',
+    limit: '10.4KB',
+  },
+] satisfies SizeLimitConfig;
--- a/library/package.json
+++ b/library/package.json
@@ -49,11 +49,19 @@
     "lint.fix": "eslint \"src/**/*.ts*\" --fix",
     "format": "prettier --write ./src",
     "format.check": "prettier --check ./src",
+    "check.size": "size-limit",
+    "check.dist": "es-check --not './dist/*.map.js' -v es2022 --module './dist/**/*.js'",
     "build": "tsup"
   },
   "devDependencies": {
     "@eslint/js": "^9.17.0",
+    "@size-limit/file": "11.1.6",
+    "@size-limit/webpack": "11.1.6",
     "@vitest/coverage-v8": "^2.1.8",
+    "@types/node": "^22.12.0",
+    "browserslist": "4.24.4",
+    "browserslist-to-esbuild": "2.1.1",
+    "es-check": "8.0.1",
     "eslint": "^9.17.0",
     "eslint-plugin-import": "^2.31.0",
     "eslint-plugin-jsdoc": "^50.6.1",
@@ -66,7 +74,8 @@
     "typescript": "^5.7.2",
     "typescript-eslint": "^8.18.2",
     "vite": "^6.0.6",
-    "vitest": "2.1.8"
+    "vitest": "2.1.8",
+    "webpack": "5.97.1"
   },
   "peerDependencies": {
     "typescript": ">=5"
--- a/library/tsup.config.ts
+++ b/library/tsup.config.ts
@@ -1,3 +1,4 @@
+import browserslistToEsbuild from 'browserslist-to-esbuild';
 import { defineConfig } from 'tsup';
 
 export default defineConfig([
@@ -8,6 +9,7 @@ export default defineConfig([
     minify: false,
     dts: true,
     outDir: './dist',
+    target: ['es2022', ...browserslistToEsbuild()],
   },
   {
     entry: ['./src/index.ts'],
@@ -19,5 +21,6 @@ export default defineConfig([
     outExtension: ({ format }) => ({
       js: format === 'cjs' ? '.min.cjs' : '.min.js',
     }),
+    target: ['es2022', ...browserslistToEsbuild()],
   },
 ]);
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '9.0'
+lockfileVersion: '6.0'
 
 settings:
   autoInstallPeers: true
@@ -10,13 +10,13 @@ importers:
     devDependencies:
       '@trivago/prettier-plugin-sort-imports':
         specifier: ^5.2.0
-        version: 5.2.0(prettier@3.4.2)
+        version: 5.2.2(prettier@3.4.2)
       prettier:
         specifier: ^3.4.2
         version: 3.4.2
       prettier-plugin-tailwindcss:
         specifier: ^0.6.9
-        version: 0.6.9(@trivago/prettier-plugin-sort-imports@5.2.0(prettier@3.4.2))(prettier@3.4.2)
+        version: 0.6.11(@trivago/prettier-plugin-sort-imports@5.2.2)(prettier@3.4.2)
       typescript:
         specifier: 5.7.2
         version: 5.7.2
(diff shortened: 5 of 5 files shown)


belgattitude:
For example of browserslist you can try out different scenarios here: https://browserslist.dev/?q=ZGVmYXVsdHMsIGNocm9tZSA%2BPSA5NixmaXJlZm94ID49IDkwLGVkZ2UgPj0gOTEsc2FmYXJpID49IDEyLGlvcyA%2BPSAxMixvcGVyYSA%2BPSA5MA%3D%3D

fabian-hiller:
Hey, thanks for creating this PR and for your further research for possible improvements in this regard.

How compatible is Valibot's current bundle? I wonder if we even need Browserslist. I try to keep things as simple as possible, but if there is a value, I will consider it.

What does size-limit do? I ask because for Valibot the total bundle size isn't really important due to our modular architecture. Adding new features doesn't grow the bundle for users who don't use them. But maybe I have misunderstood size-limit.

belgattitude:
@fabian-hiller 

# Regarding browserslist

> I wonder if we even need Browserslist. I try to keep things as simple as possible, but if there is a value, I will consider it.

If simplicity is a goal browserslist will do. It's mature and supported by all transpiler/minifiers/bundlers (swc, esbuild/tsup, webpack, babel ...). The browserslist config will inform them about what features shouldn't be transpiled.

The idea behind is simple. Code with latest ecmascript features, let the transpiler decide how to replace syntaxes that aren't supported by target. For the maintainers and the reviewers it's much easier. 

Contrived example of a PR including a js syntax that would break ios < 14:  [nullish colascing operator](https://caniuse.com/mdn-javascript_operators_nullish_coalescing_assignment). Using this feature would potentially create a ticket leading to a fix.

As of today most browsers, including node, deno and bun supports most of  es2024 (except decorators). Browserlist helps to inform the transpiler/bundler about specific features that should be transpiled.

What's at stake ? A javascript syntax error can really break an entire website. So it's good to get it under control with proper tooling rather than conventions or reviews. We don't know what es202x will bring in the future.

Another aspect and it's quite important: communicate which browsers valibot officially support. That eventually helps devs  to choose to transpile the library (ie: https://nextjs.org/docs/app/api-reference/config/next-config-js/transpilePackages) based on their compatibility requirements (some projects should be usable across enterprise systems that tend to work with really outdated browsers: firefox esr, edge 19..).

## What browserslist doesn't do

- The actual transpilation. It just inform the transpiler,minifier...
- Polyfilling. ie: Promise.try (some bundlers might use it to polyfill)
- AFAIK it  doesn't transpile down regexps
 
## Propsal

In my [last commit](https://github.com/fabian-hiller/valibot/pull/1041/commits/89e119d0f221d272be3bdb2abc4e835afcbc9e75), I've added this suggested baseline:

```
defaults
chrome >= 96
firefox >= 105
edge >= 113
safari >= 15
ios >= 15
opera >= 103
not dead
```

It should cover [94% of browsers at minimum](https://browserslist.dev/?q=ZGVmYXVsdHMsIGNocm9tZSA%2BPSA5NiwgZmlyZWZveCA%2BPSAxMDUsIGVkZ2UgPj0gMTEzLCBzYWZhcmkgPj0gMTUsIGlvcyA%2BPSAxNSwgb3BlcmEgPj0gMTAzLCBub3QgZGVhZA%3D%3D)

**Size-limit**: the same
**Dist code**: only one difference:

image

And I think it's a valid choice (made by esbuild)

## Reasons to set it

I feel it's always a good idea to have one. For example if valibot suddently receives bug reports about a particular browser version (safari ios ...), it's easier to tackle it straight from the tooling than in the code. We can even imagine having 2 dist files. One modern, one compat. But nowadays it's less an issue.

belgattitude:
@fabian-hiller 

# Why size-limit ?

You can check my [last commit](https://github.com/fabian-hiller/valibot/pull/1041/commits/dda7839957b8df80abf8cdc74ceb250af64a956b). You'll see a more interesting example

```
import type { SizeLimitConfig } from 'size-limit';

module.exports = [
  {
    name: `import * as v from 'valibot' (ESM)`,
    path: ['dist/index.js'],
    import: '*',
    limit: '10.35KB',
  },
  // Core
  {
    name: `import { parser } from 'valibot' (ESM)`,
    path: ['dist/index.js'],
    import: '{ parser }',
    limit: '335KB',
  },
  // Functions
  {
    name: `import { pipe } from 'valibot' (ESM)`,
    path: ['dist/index.js'],
    import: '{ pipe }',
    limit: '370B',
  },
  // Schemas
  {
    name: `import { array } from 'valibot' (ESM)`,
    path: ['dist/index.js'],
    import: '{ array }',
    limit: '815B',
  },
] satisfies SizeLimitConfig;
```

size-limit can use webpack or esbuild to really estimate the cost of a specific import. as ebpack is probably the most used bundler, I kept the configuration with webpack.


## Why it's useful

Estimating the impact of a refactoring or a build change is not always easy. A simple change can make the bundle grow unoticed.

## Is it useful now ?

Probably yes if we want to experiment with different bundling techniques. On the long run of course the size limit could be part of ci or a reporting action...

belgattitude:
@fabian-hiller  if you feel convinced and wants to proceed... my steps are

- Add arethetypeswrong check
- Add packageManager field

Then 

- Try to bundle in separate files (that would make the `// @PURE` style annotations obsolete)
- Try to allow `import {} from 'valibot'`

fabian-hiller:
My understanding is that most Valibot users don't embed the library code as is into their application. Instead, they run a build step that bundles their application to apply things like stree shaking. So I am not sure if Valibot should implement browserslist. It feels like developers who are using Valibot and shipping it as part of a website to a browser should instead add it to their build step.

`size-limit` is generally a good idea, especially for a library like Valibot that prioritizes small bundles. I don't think we should add it right now, but we should evaluate it after our stable v1 release to ensure a small bundle size for major changes in the long run. It might make sense to check common scenarios, such as validating a login schema, for bundle size changes.

belgattitude:
> most Valibot users don't embed the library code as is into their application. Instead, they run a build step that bundles their application to apply things like stree shaking. So I am not sure if Valibot should implement browserslist. It feels like developers who are using Valibot and shipping it as part of a website to a browser should instead add it to their build step.

Not sure I understand the relation between browserlists and tree shaking.

> they run a build step that bundles their application to apply things like stree shaking... website to a browser should instead add it to their build step.

Totally the build step will use a bundler (either vite, webpack, ...), but, correct me if I'm wrong... they don't transpile dependencies imported from node modules (deno is different in that matter). 

So imagine, you install reack-hook-form, react-query, valibot... they are installed in node_modules. Whatever you build your application with (nextjs, remix, svelte,.... - vite/webpack/...), the dependencies will just be inlined into the build outputs (chunks...). Only the application source code is transpiled, not the node_modules/deps. (at least not by default). If you're unsure maybe take a look to some projects that are generally bundled for browser usage:

ie: react-query : https://github.com/TanStack/query/blob/main/scripts/getTsupConfig.js ... and probably all frontend libs that gained some popularity

> size-limit is generally a good idea, especially for a library like Valibot that prioritizes small bundles. I don't think we should add it right now, but we should evaluate it after our stable v1 release to ensure a small bundle size for major changes in the long run. It might make sense to check common scenarios, such as validating a login schema, for bundle size changes.

Yes good idea... you'll find many examples in the wild. The fact is to get measurements at some point, less asumptions.

Good luck for the v1.

belgattitude:
By the way I started the experiment to find ideas about this question : https://github.com/fabian-hiller/valibot/issues/1018#issue-2794740406 and especially the import { v } from valibot... 

Does it make sense to go on ?

fabian-hiller:
I thought that bundlers also apply the target setting to its dependencies, but I think you are right and bundlers only bundle, tree-shake and minify the code.

What advantage do you see in using a browser list instead of just setting "ES2020" as the target?

> By the way I started the experiment to find ideas about this question ... and especially the import { v } from valibot...

I'm not sure if we should support two ways to import the library via the "v" letter, as I'm pretty sure it will confuse a lot of people and cause questions about which is better. I see the DX advantage, but maybe we should taggle that with a VS code extension instead.

belgattitude:
> What advantage do you see in using a browser list instead of just setting "ES2020" as the target?

Runtimes generally implements the ecma standards progressively. By runtime I mean, node, deno, edge, cloudflare, v8, webkit, hermes in react native… so targeting an ecma target is generally not enough to cover compatibility in the wild. I’d say nowadays it’s less an issue as ecma haven’t introduced new syntax changes since long. Ie null coalescing operator is baseline nowadays.

That said having browserslist ready is a good thing imho. The thing in my experience is that there’s always a moment where you’ll get bug reports or questions. 

It’s also a no op on the build system. It virtually does nothing except informing the transpiler about features available. In a more nuanced form than saying ie es2024.

belgattitude:
> I'm not sure if we should support two ways to import the library via the "v" letter, as I'm pretty sure it will confuse a lot of people and cause questions about which is better.

You’re totally right. I don’t consider it important myself. But I’m sure the ecosystem will converge to named exports at some point. I can clarify in another pr if you like. Might fix issues in others environments too, ie expo, rn… where the support for import * as isn’t really a priority.

On my phone so it’s a rough explanation :)

fabian-hiller:
Thank you for your words and thank you for helping to improve Valibot. I have almost no time at the moment and have to focus on the most important parts to ship our v1 release in the next weeks. Let's discuss the PR again after that.

tyteen4a03:
Hey @fabian-hiller I am looking to use Valibot in the browser and need to support older browsers. Would this be on the roadmap? 

It's not entirely clear from the docs whether valibot can be run in a browser context.

fabian-hiller:
Hey, unfortunately, I have very limited time. That's why, at the moment, I usually prioritize shipping new features or fixing important bugs. Valibot is modular. It depends on what you are using. However, most basic schemas and actions should work in older browsers. There should only be a few actions that use newer web APIs.

belgattitude:
Hello @fabian-hiller i can totally understand. If situation changes ping me, I’ll update the PR and the proposition with latest discoveries I made.

in the meantime for those impacted, most frameworks / allows to transpire the library. Ie next js transpireModules option. It should cover most cases.

PR #1052, opened 2025-02-10
add `base64url` action

### LINKED ISSUE #947: Add support for base64url regex validator
It would be great if support for a base64url regex validator could be added in addition to base64. We heavily rely on base64url over base64.

A very simple/naive regex would be:
```
[a-zA-Z0-9_-]+
```

But looking at the regex for base64 it may needs some extra parts.

EltonLobo07:
closes #947

## Summary by CodeRabbit

- **New Features**
  - Added a `base64Url` validation action for checking Base64URL-formatted strings.
  - Supports custom validation messages and strong type inference.
  - Exposes structured validation details for invalid values.

- **Documentation**
  - Added API reference pages, examples, and navigation entries for the new action and related types.

- **Tests**
  - Added runtime and TypeScript type coverage for valid inputs, invalid formats, error messages, and inferred types.


EltonLobo07:
@fabian-hiller Can you answer a couple of questions?
1. Based on my research, the word 'base64url' is used as a single word (base64url or Base64url) instead of base64Url or Base64Url. To remain consistent, I have treated base64url as a single word. So the names of the types are `Base64urlIssue`, `Base64urlAction` and so on. Do you think we should use camel case instead (`Base64UrlIssue`, `Base64UrlAction` and so on)?
2. Usually the doc comment has a Wikipedia link associated with the term. But base64url is not directly mentioned on Wikipedia. Out of the three links mentioned later, which one would you have used for the doc comment? [#1](https://en.wikipedia.org/wiki/Base64#URL_applications), [#2](https://base64.guru/standards/base64url), or [#3](https://datatracker.ietf.org/doc/html/rfc4648#section-5)?

fabian-hiller:
Thank you for the implementation!
1. I would go with `Base64Url`
2. I would use the first wiki link

yslpn:
@EltonLobo07 Sorry for the long response. Do you want to continue working on this PR? If so, we'll need to clearly clarify in the website documentation what exactly our validation accepts. And please provide the source of your regular expression.

PR #1060, opened 2025-02-17
Support readonly action in the JSON schema serialization package

dylnslck:
This adds support for the `v.readonly` action in the `to-json-schema` package. Per the [JSON Schema spec](https://json-schema.org/draft/2020-12/json-schema-validation#name-readonly-and-writeonly), this'll simply add the `readOnly: true` flag to the JSON output This is helpful for rendering UI forms where some properties should not be editable by the user.


fabian-hiller:
Thanks for the PR. There is a "problem". v.readonly is a transformation action. This means that readonly only affects the output of a Valibot schema. Since JSON Schema does not support transformations (different input and output type) it might be confusing that the behavior of Valibot and our JSON Schema conversion is different in this respect. I am happy to get feedback on this.

PR #1064, opened 2025-02-24
feat(actions): timestamps with optional "T"

emilesabatier:
Hi, 

According to ISO 8601 (cf. [https://en.wikipedia.org/wiki/ISO_8601#Times](https://en.wikipedia.org/wiki/ISO_8601#Times)), the "T" character may be optional in unambiguous contexts.

> ISO 8601-1:2019 allows the T to be omitted in the extended format, as in "13:47:30", but only allows the T to be omitted in the basic format when there is no risk of confusion with date expressions. 

I encountered the issue using pipe validation on data from postgresql + [drizzle](https://github.com/drizzle-team/drizzle-orm), where the iso timestamp string is returned without the "T" letter.

Maybe a discussion is needed on the "unambiguous" part ?


fabian-hiller:
Thank you for creating this PR. Can you share the schema that led to this issue?

emilesabatier:
Sure, here is the code sample

`
import { timestamp } from "drizzle-orm/pg-core"
import * as v from "valibot"

// Drizzle column definition, returns "2025-02-26 12:42:03+00" string from database
const timestampColumn = timestamp("created_on", { mode: "string", withTimezone: true, precision: 0 })

// Valibot schema for that column
const timestampSchema = v.pipe(
    v.string(),
    v.isoTimestamp()
)
`

fabian-hiller:
I reached out to the Drizzle team. I will get back to you in the next few days.

fabian-hiller:
I have received an answer. This seems to be a database related issue. Drizzle may be planning to address this issue in the long term. What database are you using? I will have to do some research and then decide if we want to change our current regex.

emilesabatier:
> I have received an answer. This seems to be a database related issue. Drizzle may be planning to address this issue in the long term. What database are you using? I will have to do some research and then decide if we want to change our current regex.

I am using postgresql. Reading the [docs](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-DATETIME-OUTPUT), I found this note about the "T" character under the _Table 8.14._:

> Note
ISO 8601 specifies the use of uppercase letter T to separate the date and time. PostgreSQL accepts that format on input, but on output it uses a space rather than T, as shown above. This is for readability and for consistency with [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) as well as some other database systems.

fabian-hiller:
Thank you for your research! This will be fixed in the next version.

fabian-hiller:
Valibot v1.0.0-rc.4 is available

PR #1065, opened 2025-02-25
feat: allow custom error message to be empty

andreww2012:
Hello!

I'm using Valibot together with [`@vee-validate/valibot`](http://npmjs.com/@vee-validate/valibot) for form validation on frontend.

In some cases, I'd like to give some fields an empty error message to indicate that the form is invalid, but no error message should be shown to an end user.

Currently valibot skips custom empty error messages as if it's missing and uses the default error message instead.

This PR allows custom empty error messages, which I also think was the initial intention, given the final error message is constructed in such a way that it respects empty strings:

https://github.com/fabian-hiller/valibot/blob/f25b7c7e1621e918f2670f0ba5b8abcfbe7aa550/library/src/utils/_addIssue/_addIssue.ts#L105-L112

### DIFF
--- a/library/CHANGELOG.md
+++ b/library/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to the library will be documented in this file.
 - Change `MarkOptional` type to fix order of entries and TS error when using generic schemas (issue #1021)
 - Change `VariantOption` and `VariantOptionAsync` type to fix TS error when using generic schemas (issue #842)
 - Change implementation of `variant` and `variantAsync` to support optional discriminators using `exactOptional`, `exactOptionalAsync`, `optional`, `optionalAsync`, `nullish` or `nullishAsync`
+- Change `_addIssue` to not ignore empty strings as error message (pull request #1065)
 - Refactor `bytes`, `maxBytes`, `minBytes` and `notBytes` action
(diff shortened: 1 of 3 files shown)


fabian-hiller:
Thanks for creating this PR! Can you share an example case where we want the form to error but don't want to display an error message? How does the user know the submission failed?

andreww2012:
We have a requirement that some required string fields should be filled with at least 1 character for the form submit button to work; otherwise it should be disabled. When such a field is not filled, the error message should not be shown as it's considered obvious that it must be.

Simplified code:

```ts
import {toTypedSchema} from '@vee-validate/valibot';
import * as v from 'valibot';
import {useForm} from 'vee-validate-vue3';

const {errors: formErrors} = useForm({
  validationSchema: toTypedSchema(
    v.object({
      name: v.pipe(v.string(), v.minLength(1, '')),
    }),
  ),
});

const formHasValidationErrors = computed(() => Object.keys(formErrors.value).length > 0);

const shouldDisableFormSubmit = computed(
  () => formHasValidationErrors.value || (... other conditions ...),
);
```

fabian-hiller:
v1.0.0-rc.3 is now available

PR #1068, opened 2025-03-03
Strip u flag in email regexp and improve description on v.email

### LINKED ISSUE #1067: Email regex is not strictly equivalent to that exemplified in WHATWG HTML Living Standard
https://github.com/fabian-hiller/valibot/blob/c1d8e27a20ec1fae35a1af36c70c77badd9e0ef9/library/src/regex.ts#L144-L150

The regex for `email()` seems to have been taken from the whatwg HTML Living Standard, but the upstream one seems to have been changed.

https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address

> The following JavaScript- and Perl-compatible regular expression is an implementation of the above definition.
>
> ```js
> /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/
> ```

- no `i` or `u` flag
- `\w` is no longer used

tats-u:
Fixes #1067

All tests passed by chance so I submit this PR.

Also I improved the description on `v.email()`. Readers will be able to know that it is compatible with `<input type="email">`.

### DIFF
--- a/library/src/actions/email/email.ts
+++ b/library/src/actions/email/email.ts
@@ -66,8 +66,9 @@ export interface EmailAction<
  * action.
  *
  * Hint: This validation action intentionally only validates common email
- * addresses. If you are interested in an action that covers the entire
- * specification, please use the `rfcEmail` action instead.
+ * addresses accepted by `<input type="email">`. If you are interested in an
+ * action that covers the entire specification, please use the `rfcEmail`
+ * action instead.
  *
  * @returns An email action.
  */
@@ -78,8 +79,9 @@ export function email<TInput extends string>(): EmailAction<TInput, undefined>;
  * action.
  *
  * Hint: This validation action intentionally only validates common email
- * addresses. If you are interested in an action that covers the entire
- * specification, please use the `rfcEmail` action instead.
+ * addresses accepted by `<input type="email">`. If you are interested in an
+ * action that covers the entire specification, please use the `rfcEmail`
+ * action instead.
  *
  * @param message The error message.
  *
--- a/library/src/actions/email/email.ts
+++ b/library/src/actions/email/email.ts
@@ -66,9 +66,8 @@ export interface EmailAction<
  * action.
  *
  * Hint: This validation action intentionally only validates common email
- * addresses accepted by `<input type="email">`. If you are interested in an
- * action that covers the entire specification, please use the `rfcEmail`
- * action instead.
+ * addresses. If you are interested in an action that covers the entire
+ * specification, please use the `rfcEmail` action instead.
  *
  * @returns An email action.
  */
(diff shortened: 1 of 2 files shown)


tats-u:
Probably the current regex is the form that was modified by ESLint. I may as well add a note that the regex was modified by ESLint.

fabian-hiller:
Thanks for creating this PR. Isn't this exactly the same regex? We just minified it for Valibot, but the meaning should be the same as the original regex. Can you double-check it?

tats-u:
Example: `"\u212a\u017f@example.com"` (Kſ@example.com)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_class_escape

> [Unicode character class escapes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape) start with \p and \P, but they are only supported in [Unicode-aware mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode#unicode-aware_mode). In Unicode-unaware mode, they are [identity escapes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_escape) for the p or P character.

https://unicode.org/Public/UCD/latest/ucd/CaseFolding.txt

This is due to `iu`. `u` should be omitted for the current regex.

fabian-hiller:
Thanks for your research! We should probably use the exact same regex to avoid such confusion in the long run. I will merge this PR in the next few minutes.

fabian-hiller:
I reversed the changes to email comments, because not everyone uses Valibot with HTML in the frontend.

tats-u:
We should have changed the issue title in advance.

tats-u:
Should we edit the Hint in https://valibot.dev/api/email/?

fabian-hiller:
`email` is not using the regex we modified.

tats-u:
Wow, I have overlooked the prefix `RFC_`. It's used in `rfcEmail`.

c.f. https://zenn.dev/igz0/articles/email-validation-regex-best-practices (Japanese) —Are you using regular expressions for email addresses by googling them at random?

`RFC_EMAIL_REGEX` does not cover all local parts in email addresses accepted by SMTP unlike the note "This validation action intentionally validates the entire RFC 5322 specification.". It doesn't accept quoted local-parts or comments in local-parts.

tats-u:
It would be ideal that `email` were the current behavior of `rfcEmail` and `rfcEmail` used something like `rhttps://www.npmjs.com/package/email-addresses.

tats-u:
We have only to add something like `looseRfcEmail`.

fabian-hiller:
Maybe something for later. I think `email` and `rfcEmail` cover more then 99% usage and for the rest `regex` or `custom` can be use.

fabian-hiller:
Valibot v1.0.0-rc.4 is available

PR #1070, opened 2025-03-05
Update Romanian translations

navaru:
I've simplified the expressions used, to be closer to the `en` variant.


fabian-hiller:
Thank you for this PR! @victornswd and @ancaspatariu can you confirm this improvement?

fabian-hiller:
Does the rest look all right to you @victornswd?

victornswd:
Sorry it took so long to respond, i had a very hectic week. It looks good, my recommendation would be to keep "bytes" as "byți" because that is the correct form from the dictionary, also used on the romanian wikipedia and several tech articles i checked to be sure. 
Also, "hexadecimal" should be "hexazecimal" because that is the main form of the word and is consistent with "zecimal" already used

victornswd:
sure, i can get the "bytes" argument, but not the "hexadecimal" one. there already is "zecimal" in the translation and i have never heard "hexadecimal" used in romanian, even though wiki has it as a fallback for "hexazecimal", but i have heard "hexazecimal"

victornswd:
> Is this PR ready to get merged?

yes, in my opinion

PR #1079, opened 2025-03-12
Support second and timezone in `isoDateTime`

typed-sigterm:


fabian-hiller:
Isn't this exactly what `isoTimestamp` is doing? https://valibot.dev/api/isoTimestamp/

PR #1096, opened 2025-03-22
feat: postal code validation

ysknsid25:
Japanese often implement postal code checks. But all are self-implemented.

 Even Zod does not provide it as standard.

Therefore, we thought that the addition of postal code validation could be a factor that would give valibot an advantage in Japan.

In addition, although this time we have prepared for Japan and the U.S., it is possible to extend it by adding the country code to CountryCode and the regular expression pattern corresponding to the country code to getPostalCodePattern.

### DIFF
--- a/library/src/actions/postalCode/postalCode.ts
+++ b/library/src/actions/postalCode/postalCode.ts
@@ -116,7 +116,7 @@ export function postalCode<TInput extends string, CountryCode>(
 export function postalCode<
   TInput extends string,
   CountryCode,
-  const TMessage extends
+  TMessage extends
     | ErrorMessage<PostalCodeIssue<TInput, CountryCode>>
     | undefined,
 >(
--- a/library/src/actions/postalCode/postalCode.ts
+++ b/library/src/actions/postalCode/postalCode.ts
@@ -1,0 +1,154 @@
+import type {
+  BaseIssue,
+  BaseValidation,
+  ErrorMessage,
+} from '../../types/index.ts';
+import { _addIssue } from '../../utils/index.ts';
+
+/**
+ * Type to determine postal code format by country.
+ * Country code is based on [ISO 3166-1 alpha-2](https://ja.wikipedia.org/wiki/ISO_3166-1).
+ *
+ */
+const COUNTRY_CODE = {
+  JP: 'JP',
+  US: 'US',
+} as const;
+
+export type CountryCode = keyof typeof COUNTRY_CODE;
+
+/**
+ * get postal code pattern by country code
+ *
+ * @param countryCode Country code which based on [ISO 3166-1 alpha-2](https://ja.wikipedia.org/wiki/ISO_3166-1)
+ *
+ * @returns postal code pattern
+ */
+const getPostalCodePattern = (countryCode: CountryCode): RegExp => {
+  switch (countryCode) {
+    case COUNTRY_CODE.JP:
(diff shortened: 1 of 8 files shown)


fabian-hiller:
Thanks for creating this PR! I am not sure if we should add a `postalCode` action because it is country specific and hard to get right, and supporting every country will likely lead to a larger bundle size. I will wait to see if more developers show interest in this PR before reviewing your code. In the meantime, a good workaround is to use our `regex` action.

ysknsid25:
@fabian-hiller 

Nice to meet you!

I agree with you, the reason why I implemented only JP and US was because I thought it would be difficult to cover all countries and if there is a demand for each country, I would be happy to have it expanded 👍

fabian-hiller:
This is also the reason why we don't provide a `phone` action yet. While `postalCode` is probably a bit easier to implement, it might still be hard to support every country without making a mistake. Therefore, it might make sense to implement such a validation function as an independent library that can be used together with our `check` action.

ysknsid25:
I agree. It might be interesting if the method is extended with plug-ins, etc.

library/src/actions/postalCode/postalCode.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Type to determine postal code format by country.
+ * Country code is based on [ISO 3166-1 alpha-2](https://ja.wikipedia.org/wiki/ISO_3166-1).
+ *
+ */
+const COUNTRY_CODE = {
+  JP: 'JP',
+  US: 'US',
+} as const;
```
sebcomeau, line 16:
```suggestion
const COUNTRY_CODE = {
  CA: 'CA',
  JP: 'JP',
  US: 'US',
} as const;
```

library/src/actions/postalCode/postalCode.ts
```diff
@@ -0,0 +1,154 @@
...
+ */
+const getPostalCodePattern = (countryCode: CountryCode): RegExp => {
+  switch (countryCode) {
+    case COUNTRY_CODE.JP:
+      return /^\d{3}-?\d{4}$/u;
+    case COUNTRY_CODE.US:
+      return /^\d{5}(?:-\d{4})?$/u;
+    default:
+      throw new Error(`Unsupported country code: ${countryCode}`);
+  }
```
sebcomeau, line 35:
```suggestion
  switch (countryCode) {
    case COUNTRY_CODE.CA:
      return /^(?!.*[DFIOQU])[A-VXY][0-9][A-Z] ?[0-9][A-Z][0-9]$/i;
    case COUNTRY_CODE.JP:
      return /^\d{3}-?\d{4}$/u;
    case COUNTRY_CODE.US:
      return /^\d{5}(?:-\d{4})?$/u;
    default:
      throw new Error(`Unsupported country code: ${countryCode}`);
  }
```

sebcomeau (commented):
### Canadian postal code

**The Validation**

The regex validates the pattern of a Canadian postal code. It checks for a 6-character alphanumeric string with an optional middle space. The code ensures the input matches the case-insensitive format of Letter-Number-Letter Number-Letter-Number (e.g., K1A 0B1).

**Core Requirements**

* Structure: Exactly A1A 1A1 (or A1A1A1 without the space).
* Illegal Letters: The letters D, F, I, O, Q, and U are banned anywhere in the code.
* Illegal Starts: The first letter can never be W or Z.

PR #1099, opened 2025-03-22
RFC: Add `to-valibot` package

muningis:
This packages allows transforming OpenAPI Declarations (.yaml and .json) and JSON Schemas (tested and unit test covered with draft-07).

Some of JSON Schema validations (minProperties, maxProperties, isoDuration) are covered by https://github.com/fabian-hiller/valibot/pull/1100#issuecomment-2746253192


fabian-hiller:
Thanks for creating this PR! In simple words, is the general idea of this package to read a JSON or YAML file containing OpenAPI or JSON schema and convert it to a Valibot schema? If so, does it depend on Node.js to read the files? Can users also just pass an OpenAPI or JSON Schema object (without reading a file) to convert it to a Valibot schema? Does it make sense to split it into two packages `from-josn-schema` and `from-openapi`? On of the packages could use the other as a dependency.

muningis:
> is the general idea of this package to read a JSON or YAML file containing OpenAPI or JSON schema and convert it to a Valibot schema

Yes, but not limited to that. With what we have, we could allow parsing other formats too.


> If so, does it depend on Node.js to read the files? Can users also just pass an OpenAPI or JSON Schema object (without reading a file) to convert it to a Valibot schema?

It's up to users how they will retrieve said files. Something such as OpenAPI Declaration file would most likely retrieved over an API call. However, at the moment current implementation depends on node to write to the file system (we could allow instead just to return generated code so users can save it however they wish)


> Does it make sense to split it into two packages from-josn-schema and from-openapi? On of the packages could use the other as a dependency.

I'm open to discussion, but don't really see benefit of that. Bundle size difference is barely noticeable, and it would only be ran on dev machine (or CI where code is built, considering types output gets .gitignored).

muningis:
Questionable functionalities, which can't be replicated by Typescript's types (but can be validated with custom validators/actions)

- https://json-schema.org/understanding-json-schema/reference/array#contains (and minContains/maxContains)
- https://json-schema.org/understanding-json-schema/reference/object#patternProperties - while it's possible to have template literal types, without magic similar to arktype (we would have to parse regexp) this can't be replicated in typescript types
- https://json-schema.org/understanding-json-schema/reference/array#tupleValidation (prefixItems) - unsure if can have tuple+array `[string, number, User, ...string[]]` inferable.

Additional concerns on https://json-schema.org/understanding-json-schema/reference/combining
- anyOf, allOf, oneOf -> we can check for those, but probably in TS it could only be `(T | U)[]`
- not -> we can check it (validate type and expect it to be not valid), but unsure if we can replicate in types, except maybe for `type Not<T, U> = T extends U ? never : T;` but I'm not too confident.

fabian-hiller:
Hey, I'm very sorry for my late reply, but I don't have much time right now. I recommend to continue this library as a community project for now because of time constraints on my side. Sorry for that! I am happy to help you share it on social media and connect you with other developers in our ecosystem if you are interested. Feel free to contact me again in about two months. I may have a little more time then. I also want to thank you for your recent contributions to the project! I hope to be able to give back some day! Please reach out if you are motivated to contribute even more. For example, we are looking for someone interested in investigating a possible VS Code extension implementation.

PR #1102, opened 2025-03-23
add iso 8601 duration action

muningis:
ISO8601 Duration (https://en.wikipedia.org/wiki/ISO_8601#Durations) - used  (primarly) by `string()` schema


fabian-hiller:
Thank you very much! It may take me a while to review this PR as it is very important to me to check the validation closely. I hope to get back to you next week.

muningis:
Yeah, this one is _quite biggish_.

There's still also a question of whether 8601 (current), 8601-1 (allows to use Week designator with Year/Month/Day) or 8601-2 (allows +- before P, to determine if it's positive or negative value) should be supported.

fabian-hiller:
I will focus on reviewing and merging other actions first. If users want to see this PR merged, please give a 👍.

muningis:
This schema is very niche, and I wouldn’t expect many people to need it. This is more of to have better compatibility with JSON Schemas. 

with that in mind, this PR misses adjustments for to-json-schema package, so that this gets transformed to string’s format. Same for `minEntries` and `maxEntries`.

fabian-hiller:
I will try to release it with Valibot v1.2

yslpn:
@muningis Hi. Are you interested in refining this PR?

Hi. Are you interested in refining this PR?

I see you've created a complex function with 7 regexps ​​that thoroughly validates duration. This is quite verbose; our library is aimed at reducing bundle size and speed.

Could we use a more flexible and common validation, a single regular expression? Yes, it won't be as precise, but I think it's good enough.

https://github.com/ajv-validator/ajv-formats/blob/master/src/formats.ts#L54

Link https://github.com/open-circle/valibot/issues/1497

And please fix conflicts

PR #1111, opened 2025-03-25
Feat: Add minDigitChars

will-stone:
Hi Fabian 👋 🙂 

Long time fan of this project. Thought I'd make a contribution...

Added a check for minimum amount of digits in a string. Useful for password validation. Only thing I haven't done is update the i18n package, are we expected to do that?

Also, I noticed that when running `pnpm -F valibot test` it creates a `tsconfig.vitest-temp.json` file so have added a commit to ignore that.

Thanks,

Will.

## Summary by CodeRabbit

* **New Features**
  * Added `minDigitChars` validation action for strings, plus a digit-count utility used by the action.
* **Tests**
  * Added unit and type-level coverage for digit counting, `minDigitChars` construction, and validation/inference behavior.
* **Documentation**
  * Added complete API documentation for `minDigitChars` and its related types, and updated API listings/menus to include it.
* **Chores**
  * Updated project ignore settings for test artifacts and extended public re-exports for the new utilities/actions.


fabian-hiller:
Thanks for the PR! Great idea! Since this is a very specific action, I may wait for more developers to express interest in such a feature before reviewing and merging it.

will-stone:
Thanks @fabian-hiller 🙂 If this was accepted, I was going to propose:

- `maxDigits`
- `minLetters`
- `maxLetters`

I'd also like `minSpecialCharacters` but not sure yet how to do that, that couldn't lead to possible regex injection.

will-stone:
If anyone stumbles across this and is looking for a way to implement this now, you can use this:

```ts
const minDigits = (min: number) =>
  v.pipe(
    v.string(),
    v.check(
      (input) => (input.match(/\d/gu) || []).length >= min,
      `Must contain at least ${min} digit${min === 1 ? "" : "s"}.`,
    ),
  )

v.safeParse(minDigits(2), 'I have no digits')
```

fabian-hiller:
In general, I am very interested in providing actions to validate the number of digits, letters and special characters, but we need to find perfect names for them. Valibot usually provides actions like `length`, `notLength`, `minLength` and `maxLength` to cover every possible case. So I am not sure if we should stick to the name `minDigits` since we already have a `digits` action that does not validate the number of digits. So maybe we should rename it to something like `minDigitChars`. I recommend brainstorming about the perfect names in a first step. In a second step we will make this PR perfect and merge it, and in a third step we can create PRs for all the other actions.

will-stone:
Ah, of course; I saw `digits` and associated it with `bytes` and `length`. As for naming, it does feel like `digits` is the odd one out here. Perhaps it could be enhanced, to avoid a breaking change, like so?

```js
v.digits('The string contains something other than digits.')
v.digits(8, 'Exactly 8 digits are required.')
```

fabian-hiller:
I don't recommend this as it would make our API inconsistent. What about `minDigitChars`, `minSpecialChars` and `minLetterChars`?

will-stone:
Okay let's go with that 👍 🙂 I'll update this PR when I can a moment. Thanks.

will-stone:
Okay, that should be good to go; changed name and rebased onto `upstream/main` 👍 🙂

fabian-hiller:
Thank you so much! It looks good on a first view. I will try to review and merge it soon. Are you also interested in adding `maxDigitChars`, `digitChars` and `notDigitChars`?

will-stone:
I certainly am interested in adding those.

Are there any plans to automate the docs creation? That’s probably the fiddliest part about contributing to this project 😅 [I’ve dabbled with typdoc](https://github.com/will-stone/tings/blob/main/website/src/data/typedoc.json) on my personal utility library, [Tings](https://tings.wstone.uk/), and it works quite nicely. Your set up is a bit more sophisticated as you’re linking pages, but maybe that could go into a JSDoc directive. It would reduce contributor effort, and your review effort massively.

fabian-hiller:
I'm not looking for a perfect solution. There will still be a manual part. I am more looking for a CLI script that speeds up the process by generating an initial template with all the info we already know from the source code.

will-stone:
Thanks @fabian-hiller I'm not sure if I'm going to get the time unfortunately; my time is quite limited these days with family commitments. But if someone else wants to implement this I highly suggest tying it into some custom JSDoc, e.g.

```ts
/**
 * Creates a min digits validation action.
 *
 * @param requirement The minimum digits.
 *
 * @returns A min digits action.
 *
 * @schemas any, array, custom, instance, string, tuple, unknown
 * @methods pipe
 * @utils isOfKind, isOfType
 */
export function minDigitChars<
```

It should get to the point where contributors do not need to touch the docs at all 👍🤓

fabian-hiller:
That is ok. I am sure someone else will look at it in the long run. For now, I prefer to keep our JSDoc comments simple and don't add extra stuff there. We need a separate JSDoc comment for each TypeScript overload function signature, and adding all that info would force us to duplicate it, which is annoying to maintain.

fabian-hiller:
Sorry for the delay. I try to catch up in the next few weeks.

yslpn:
@will-stone  Hi. Are you interested in further developing this pull request to bring it to a merge? If so, I'll review it.

Unfortunately, I don't think we can release just minDigitChars; the API needs to be symmetrical.

will-stone:
Hi @yslpn Ah, sorry! I've not had much time to review this. I'm more than happy if someone else would like to take the reins or close this off for a future time. Apologies I was unable to finish it.

PR #1112, opened 2025-03-25
partial POC implementation for deepVariant

muningis:


library/src/schemas/deepVariant/deepVariant.test.ts
```diff
@@ -0,0 +1,47 @@
...
+        type: "number"
+      },
+      data: {
+        minValue: 5,
+        maxValue: 5
+      }
+    });
+
+    matchVariation(data, "node.type", {
+      number(data) {
```
muningis, line 38: Value here is narrowed down

library/src/schemas/deepVariant/deepVariant.test.ts
```diff
@@ -0,0 +1,47 @@
...
+        console.log("It's a number!", {data});
+      },
+      string(data) {
```
muningis, line 41: Value here is narrowed down

fabian-hiller:
Thanks for this POC! I will wait for more developers to express interest in `deepVaraint`, but I can imagine providing such an API in the long run.

PR #1117, opened 2025-03-27
Use GitHub profiles in documentation

idleberg:
Not sure whether I should go to great lengths explaining why it's the right thing to do. In the past couple of months the web community has really grown on Bluesky, so it's probably worth promoting it even more.


fabian-hiller:
Thank you for creating this PR. Maybe we should link to GitHub instead. What do you think?

fabian-hiller (approved):
Thank you!

Issue #1114, opened 2025-03-27
Website: Logging object with null prototype causes error in playground


bjohn465:
In the Playground section of the Valibot website, attempting to log an object that has a `null` prototype results in an error.

[Here's a playground link](https://valibot.dev/playground/?code=MYewdgziA2CmB00QHMAUB5ARgK1sALvMAE6wCG+sqYArtNAJQMDcQA) with the following code:

```ts
console.log(Object.create(null));
```

Running this code results in the following error in the browser console (this is what is logged in Google Chrome, at least):

```
Uncaught TypeError: Cannot read properties of null (reading 'constructor')
    at Object.<anonymous> (about:srcdoc:55:54)
    at JSON.stringify (<anonymous>)
    at about:srcdoc:42:29
    at Array.map (<anonymous>)
    at stringify (about:srcdoc:35:6)
    at console.<computed> [as log] (about:srcdoc:130:52)
    at <anonymous>:1:9
```

There is an attempt to display the same error message in the console output side panel on the page, but the `<anonymous>` and `<computed>` bits are interpreted as HTML markup, so they are not displayed.

It appears that the problem is with [the following line](https://github.com/fabian-hiller/valibot/blob/d19e9e9/website/src/routes/playground/iframeCode.js#L47) in `website/src/routes/playground/iframeCode.js`:

```js
const name = Object.getPrototypeOf(value).constructor.name;
```

Since the object's prototype is `null`, attempting to access the `constructor` property on `null` fails.

PR #1118, opened 2025-03-27
Fix stringification of objects with null prototype

bjohn465:
`Object.getPrototypeOf` can return `null` if the object passed to it has a `null` prototype. If that is the case, we can not access any properties on the `null` prototype.

Use optional chaining after getting the object's prototype to attempt to get the name of the prototype's constructor. Then check to see if `name` is a string before checking to see if the constructor's name is something we want to use in place of the value in the stringified result.

This fixes fabian-hiller/valibot#1114.


fabian-hiller:
Great catch and thank you for fixing it! 🙏

fabian-hiller (approved):

PR #1133, opened 2025-04-06
fix(to-json-schema): allow any order of schemas in conversion config definitions

### LINKED ISSUE #1132: to-json-schema: Dependee not `$ref`ed if listed after dependent schema in `definitions`
In order to create named definitions in the output JSON Schema consistently, it seems toJsonSchema requires the schemas in `definitions` to be ordered such that if A depends on B, B is listed before A. Otherwise, toJsonSchema inserts the expanded definition of B inside of the definition of A.

```ts
const ageSchema = v.pipe(v.number(), v.minValue(0));
const agesSchema = v.array(ageSchema);

toJsonSchema(agesSchema, {
  definitions: { ageSchema, agesSchema }, // ageSchema *before* agesSchema
});
// result:
// {
//   "$schema": "http://json-schema.org/draft-07/schema#",
//   "$ref": "#/$defs/agesSchema",
//   "$defs": {
//     "ageSchema": {
//       "type": "number",
//       "minimum": 0
//     },
//     "agesSchema": {
//       "type": "array",
//       "items": {
//         "$ref": "#/$defs/ageSchema"    // ageSchema is `$ref`ed, as expected
//       }
//     }
//   }
// }

toJsonSchema(agesSchema, {
  definitions: { agesSchema, ageSchema }, // ageSchema *after* agesSchema
});
// result:
// {
//   "$schema": "http://json-schema.org/draft-07/schema#",
//   "$ref": "#/$defs/agesSchema",
//   "$defs": {
//     "agesSchema": {
//       "type": "array",
//       "items": {
//         "type": "number",              // ageSchema is expanded here despite
//         "minimum": 0                   // $defs.ageSchema existing in the output
//       }
//     },
//     "ageSchema": {
//       "type": "number",
//       "minimum": 0
//     }
//   }
// }
```

I believe it would make more sense for the output to be consistent regardless of the order of schemas in `definitions` (unless I am missing the intention behind the current design).

If this was just an oversight, I would be willing to try implementing a quick fix if that is alright.

Thanks.

sruenwg:
Fixes #1132.

This change should make it so named schema definitions passed to toJsonSchema are always `$ref`ed to in the output JSON Schema (rather than unexpectedly expanded) regardless of the order the user lists them in. (See [example](https://github.com/fabian-hiller/valibot/compare/main...sruenwg:valibot:fix-json-schema-named-schema-defs-order?expand=1#diff-21505a6f9dce6c1b54b223d10f6a10796bdd18a7ec54b616176566ad30ed8cffR81-R93))

Please let me know if something doesn't look right / could be improved. Thanks!


sruenwg:
> [nitpick] Consider using a named constant or refactoring to an options object instead of passing a bare boolean literal to clarify the purpose of the parameter.

I was unsure about having the bare boolean parameter and so I did consider placing `isDefinition` inside of the context object, but thought that the verbosity of that design outweighed its benefits.
Most recursive calls to `convertSchema` would end up looking something like this:
```ts
convertSchema(
  {},
  innerSchema,
  config,
  { ...context, isDefinition: false }, // <--
);
```
which I preferred to avoid.

fabian-hiller:
Great work! Thanks a lot for this PR!

fabian-hiller (approved):

fabian-hiller:
[v1.1.0](https://github.com/fabian-hiller/valibot/releases/tag/v1.1.0-to-json-schema) of `toJsonSchema` is available

PR #1142, opened 2025-04-09
feat: add formData schema, checkbox action, and formDataEntries action

EskiMojo14:
Given a form:
```html
<form enctype="multipart/form-data" method="post">
  <input name="username" type="text" />
  <label>
    <input type="checkbox" name="accepted" />
    Accept T&Cs
  </label>
  <fieldset>
    <legend>Contact me</legend>
    <label>
      <input type="checkbox" name="contact" value="SMS" />
      SMS
    </label>
    <label>
      <input type="checkbox" name="contact" value="Email" />
      Email
    </label>
  </fieldset>
</form>
```

You could validate it:
```ts
const FormSchema = v.pipe(
  v.formData(),
  v.formDataEntries(["contact"]),
  v.object({
    username: v.string(),
    accepted: v.pipe(v.string(), v.checkbox(), v.literal(true))
    contact: v.array(v.string())
  })
)
```

## Summary by CodeRabbit

* **New Features**
  * Added a checkbox transformation that converts `"on"` to `true` and other values to `false`.
  * Added a FormData schema for validation.
  * Added a FormData transformation that groups repeated fields into arrays, with support for selecting fields that should always be arrays.

* **Documentation**
  * Added API reference pages and navigation entries for the new actions, schema, and related types.

* **Tests**
  * Added runtime and TypeScript type coverage for validation, transformations, and inferred types.


fabian-hiller:
Thanks for your feedback. this PR! I think I would recommend using a library like `decode-formdata` instead for such use cases: https://github.com/fabian-hiller/decode-formdata

What do you think?

EskiMojo14:
> Thanks your for this PR! I think I would recommend using a library like `decode-formdata` instead for such use cases: https://github.com/fabian-hiller/decode-formdata
> 
> What do you think?

For more complex forms, I agree:

https://github.com/fabian-hiller/valibot/blob/0e9fd56013c8fe94de392db4d4119f164b1115e6/library/src/actions/formDataEntries/formDataEntries.ts#L40

I still think there could be room for a built-in alternative for simpler forms, and then the user can switch when they need.

Alternatively, it could be interesting to have an action wrapper for `decode-formdata`:

```ts
v.pipe(
  v.formData(),  
  decode({
    arrays: ['tags', 'images'],
    booleans: ['active'],
    dates: ['created', 'images.$.created'],
    files: ['images.$.file'],
    numbers: ['price'],
  }),
  finalSchema
)
```

fabian-hiller:
Yes, you are right. `decode-formdata` is too opinionated to provide an official schema or actions based on it. I will think about following your recommendation and provide a `formData` schema and a basic `decodeFormData` action that just converts the entries to a JS object without any special behaviour.

What is your use case for using `v.pipe(v.formData(), v.decodeFormData())`? Why not do the decoding before and pass the result to Valibot. I ask because I want to evaluate if we should add it, as there is always a downside when extending the API interface of the library.

EskiMojo14:
just preference really - i like having the transformation contained in the schema, rather than having to do it separately before feeding it in

fabian-hiller:
Maybe it is a good idea to mention on what we are woking on in our v1.1 blog post so people can reach out if they are interested in this PR:

PR #1154, opened 2025-04-12
Create getMetadata/getTitle/getDescription methods

EskiMojo14:
Loops through schema (including nested pipes) to shallowly merge metadata objects.

```ts
const schema = v.pipe(
  v.string(),
  v.title('title'),
  v.email(),
  v.description('description'),
  v.metadata({ key: 'value' }),
  v.pipe(v.string(), v.metadata({ extra: 'extra' }))
);
expect(v.getTitle(schema)).toBe('title')
expect(v.getDescription(schema)).toBe('description')
expect(v.getMetadata(schema)).toStrictEqual({
  key: 'value',
  extra: 'extra',
});
```


library/src/methods/getMetadata/getMetadata.test.ts
```diff
@@ -0,0 +1,49 @@
+import { describe, expect, test } from 'vitest';
+import * as v from '../../index.ts';
+
+describe('getMetadata', () => {
```
muningis, line 4: Perhaps we should have a test, which tests with nested object?

EskiMojo14: not sure what you mean by nested object, sorry

muningis:
```
v.pipe(
  v.object({
    name: v.pipe(v.string(), v.title("The Name")0
  }),
  v.title("An object")
);
```

Or does this just get metadata of a single field?

EskiMojo14: this just gets the metadata of the single schema, not any nested schemas - you'd need to do `v.getMetadata(schema.pipe[0].entries.name)`

library/src/utils/entriesFromObjects/entriesFromObjects.ts
```diff
@@ -80,7 +64,7 @@ type RecursiveMerge<TSchemas extends readonly [Schema, ...Schema[]]> =
           infer TFirstSchema extends Schema,
           ...infer TRestSchemas extends readonly [Schema, ...Schema[]],
         ]
-      ? MergeEntries<TFirstSchema['entries'], RecursiveMerge<TRestSchemas>>
+      ? Overwrite<TFirstSchema['entries'], RecursiveMerge<TRestSchemas>>
```
fabian-hiller, line 67: Are we sure that the new type is equal? If so, should we do a small benchmark to prove the new type is better?

library/src/methods/pipe/pipe.ts
```diff
@@ -24,7 +24,7 @@ export type SchemaWithPipe<
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     ...PipeItem<any, unknown, BaseIssue<unknown>>[],
   ],
-> = Omit<FirstTupleItem<TPipe>, '~standard' | '~run' | '~types'> & {
+> = Omit<FirstTupleItem<TPipe>, 'pipe' | '~standard' | '~run' | '~types'> & {
```
fabian-hiller, line 27: Great catch! We should also fix it for `pipeAsync`.

library/src/utils/entriesFromObjects/entriesFromObjects.ts
EskiMojo14: I'm not sure how to benchmark types - if you do you're welcome to 😅

library/src/methods/pipe/pipe.ts
EskiMojo14: already in this PR :)

library/src/methods/metadata/getTitle/getTitle.ts
```diff
@@ -0,0 +1,30 @@
...
+    : TFirst extends TitleAction<unknown, infer TTitle>
+      ? ExtractTitle<TRest, TTitle>
+      : ExtractTitle<TRest, Acc>
+  : Acc;
+
+export type GetTitle = <const TSchema extends UnknownSchemaWithPipe>(
+  schema: TSchema
+) => ExtractTitle<TSchema['pipe']>;
+
+export const getTitle = _createMetadataMethod(
```
fabian-hiller, line 27: I am not sure if we should use the same Util for `getDescription`, `getTitle` and `getMetadata` as the algorithms should probably be a bit different. `getMetadata` needs to check each `metadata` action and merge their values in a defined order. `getDescription` and `getTitle`, on the other hand, should break as soon as they find the "first" matching action.

EskiMojo14: first or last?
```ts
const EmailSchema = v.pipe(
  v.string(),
  v.email(),
  v.title("Email")
)

const GmailSchema = v.pipe(
  EmailSchema,
  v.title("Gmail")
)
```

fabian-hiller:
As a general feedback. I would not put the metadata method folders in a new folder `/metadata` folder and put them directly in the `/methods` folder.

library/src/methods/metadata/getTitle/getTitle.ts
fabian-hiller: What's the current order in which the metadata values are merged? What's the order in which the title and description are extracted? There are common algorithms for such cases like breadth-first search and depth-first search. We should mention this in the docs as well.

fabian-hiller:
My goal is to finalize and merge this PR tomorrow or Tuesday and release these methods with Valibot v1.1 this week. If it takes us longer than that, I will have to reschedule it for Valibot v1.2.

library/src/methods/metadata/getTitle/getTitle.ts
EskiMojo14: `v.getMetadata(v.pipe(v.string, v.metadata(obj1), v.pipe(v.string(), v.metadata(obj2)), v.metadata(obj3))) === Object.assign({}, obj1, obj2, obj3)`

`v.getTitle(v.pipe(v.string, v.title("1"), v.pipe(v.string(), v.title("2")), v.metadata("3"))) === "3"`

library/src/methods/getTitle/getTitle.ts
```diff
@@ -0,0 +1,30 @@
...
+  ? TFirst extends
+      | SchemaWithPipe<infer TPipe2>
+      | SchemaWithPipeAsync<infer TPipe2>
+    ? ExtractTitle<TRest, ExtractTitle<TPipe2, Acc>>
+    : TFirst extends TitleAction<unknown, infer TTitle>
+      ? ExtractTitle<TRest, TTitle>
+      : ExtractTitle<TRest, Acc>
+  : Acc;
+
+export type GetTitle = <const TSchema extends UnknownSchemaWithPipe>(
```
fabian-hiller, line 23: Do we need the two types `ExtractTitle` and `GetTitle`? Can't `ExtractTitle` access the pipe with `TSchema['pipe']` internally?

library/src/methods/metadata/getTitle/getTitle.ts
EskiMojo14: basically you can think of pipes as being flattened

library/src/methods/getTitle/getTitle.ts
EskiMojo14: ExtractTitle is recursive, so it needs to be able to use itself with just the pipe available

EskiMojo14: I suppose we could inline GetTitle on the variable annotation if you'd prefer though

library/src/methods/getTitle/getTitle.ts
```diff
@@ -0,0 +1,30 @@
...
+  schema: TSchema
+) => ExtractTitle<TSchema['pipe']>;
+
+export const getTitle = _createMetadataMethod(
```
fabian-hiller, line 27: I prefer to use the `function` keyword for all top level functions and use const and arrow function only for internal, e.g. recursive function of a root function. If it is only a few lines of code, it is also ok to not use a util and repeat the same code in `getDescription`. This is just general feedback. I recommend discussing and solving [this commen](https://github.com/fabian-hiller/valibot/pull/1154#discussion_r2062746619)t first before any other comments.

EskiMojo14: having a util seemed useful for ensuring consistency and supporting any future metadata actions

EskiMojo14: it also includes some useful type guarding DX, which you don't normally get when processing unknown pipe items

EskiMojo14: i.e. `action.kind === "metadata" && action.type === "title"` isn't enough to narrow `action` to `TitleAction`, it's still just `BaseMetadata`

library/src/utils/entriesFromObjects/entriesFromObjects.ts
fabian-hiller: Create two files for both types and execute the type with the same sample data. Then run `tsc --noEmit --extendedDiagnostics` for both files.

library/src/methods/metadata/getTitle/getTitle.ts
fabian-hiller: I think for `getMetadata` I would use depth-first search. So whenever there is a nested `pipe`, you search it first before continuing with the next element of the current pipe. Here is some untested sample code.

```ts
function getMetadata(schema) {
  const metadata = {};
  const depthFirstSearch = (currentSchema) => {
    if ('pipe' in currentSchema) {
      for (const item of currentSchema.pipe) {
        if ('pipe' in item) {
          depthFirstSearch(item);
        }
        if (item.type === 'metadata') {
          Object.assign(metadata, item);
        }
      }
    }
  };
  depthFirstSearch(schema);
  return metadata;
}
```

EskiMojo14: isn't that what the current code is doing?

```ts
function getMetadata(
    schema: UnknownSchemaWithPipe,
    _acc: Record<string, unknown> = {}
  ): Record<string, unknown> {
    return schema.pipe.reduce((acc, item) => {
      if (item.kind === 'metadata' && isMetadata(item)) {
        return Object.assign(item, acc);
      } else if ('pipe' in item) {
        return getMetadata(item, acc);
      }
      return acc;
    }, _acc);
  };
```

fabian-hiller: For `getTitle` and `getDescription` I would probably use a reverse breadth-first search that stops after finding the first title or description. Here is some untested sample code.

```ts
function getTitle(schema) {
  const reverseBreadthFirstSearch = (currentSchema): string | undefined => {
    if ('pipe' in currentSchema) {
      const schemas = [];
      for (let index = currentSchema.pipe.length - 1; index >= 0; index--) {
        const item = currentSchema.pipe[index];
        if ('pipe' in item) {
          schemas.push(item);
        }
        if (item.type === 'title') {
          return item.title;
        }
      }
      for (const item of schemas) {
        const title = reverseBreadthFirstSearch(item);
        if (title) return title;
      }
    }
  };
  return reverseBreadthFirstSearch(schema);
}
```

library/src/methods/getTitle/getTitle.ts
fabian-hiller: Feel free to leave it as is. I can review and unify it with our other source code once we think the implementation is in a good state.

library/src/methods/getTitle/getTitle.ts
fabian-hiller: Thanks for your feedback. Feel free to leave it as is. I can review and unify it with our other source code once we think the implementation is in a good state.

library/src/methods/metadata/getTitle/getTitle.ts
fabian-hiller: Would you implement `getTitle` and `getDescription` differently?

EskiMojo14: i guess i can see the benefit of searching backwards for the first instead of searching through all for the last, yeah

fabian-hiller: > isn't that what the current code is doing?

Yes. That looks good. Sorry for not checking your source code carefully before asking.

EskiMojo14: How would you represent the reverse breadth search in Typescript?

EskiMojo14: hmm, maybe 

```ts
export type ExtractTitle<TPipe extends UnknownPipe> = TPipe extends readonly [
  ...infer TRest extends UnknownPipe,
  infer TLast,
]
  ? TLast extends
      | SchemaWithPipe<infer TPipe2>
      | SchemaWithPipeAsync<infer TPipe2>
    ? ExtractTitle<TPipe2> extends string
      ? ExtractTitle<TPipe2>
      : ExtractTitle<TRest>
    : TLast extends TitleAction<unknown, infer TTitle>
      ? TTitle
      : ExtractTitle<TRest>
  : undefined;
```

fabian-hiller: Here is a first draft:

```ts
type Schema =
  | BaseSchema<unknown, unknown, BaseIssue<unknown>>
  | SchemaWithPipe<
      [
        BaseSchema<unknown, unknown, BaseIssue<unknown>>,
        ...(
          | PipeItem<any, unknown, BaseIssue<unknown>>
          | TitleAction<any, string>
          | DescriptionAction<any, string>
        )[],
      ]
    >;

function reverseBreadthFirstSearch(
  rootSchema: Schema,
  type: 'title' | 'description'
): string | undefined {
  if ('pipe' in rootSchema) {
    const nestedSchemas: Schema[] = [];
    for (let index = rootSchema.pipe.length - 1; index >= 0; index--) {
      const item = rootSchema.pipe[index];
      if (item.kind === 'schema' && 'pipe' in item) {
        nestedSchemas.push(item);
      }
      if (item.kind === 'metadata' && item.type === type) {
        return item[type];
      }
    }
    for (const nestedSchema of nestedSchemas) {
      const title = reverseBreadthFirstSearch(nestedSchema, type);
      if (title) return title;
    }
  }
}

function getTitle<TSchema extends Schema>(schema: TSchema): GetTitle<TSchema> {
  return reverseBreadthFirstSearch(schema, 'title') as GetTitle<TSchema>;
}
```

EskiMojo14: I meant in the ExtractTitle type 😛

fabian-hiller: Not sure. I need more time to investigate. If it is not possible or we are not able to solve it right now. I am also ok with marking this function as beta and returning `string | undefined`.

EskiMojo14: it's a little more complex, but I agree that prioritising less nested titles/descriptions makes sense.

fabian-hiller: Should we simplify the return type to `string | undefined` with `@bate` and a `TODO` comment to fix it later an publish it with Valibot v1.1 this week or delay by about a month and release it with v1.2?

EskiMojo14:
Website builds locally, I have no visibility on what's upsetting Vercel 🤔

fabian-hiller:
Don't worry. I will make sure everything is ok before merging. Thank you for updating the PR! I will probably merge it on the weekend and publish it with Valibot v1.1.

fabian-hiller:
I started to review this PR and will probably merge it tomorrow.

fabian-hiller:
Next, I will review the docs and merge the PR.

fabian-hiller (approved):

fabian-hiller:

PR #1155, opened 2025-04-12
fix: getFallbacks use fallback value of an object

### LINKED ISSUE #1150: getFallbacks Fails Wtih Nested Objects
I'm not sure if this is a skill issue or a limitation of valibot but I have just run into some trouble with extracting fallback values and I'm not sure what else to try.
 

```ts
const VSearchParams = v.object({
	units: v.fallback(v.union([v.literal("SI"), v.literal("IP")]), "IP"),
	selectedCoordinate: v.fallback(
		v.object({
			longitude: coerceToNumber,
			latitude: coerceToNumber,
		}),
		{
			longitude: 30.2525,
			latitude: -97.7539,
		},
	),
	timeRange: v.fallback(
		coerceToNumber,
		1,
	),
});

const t = v.getFallbacks(VSearchParams);
// const t: {
//     units: "IP";
//     selectedCoordinate: {
//         longitude: undefined;
//         latitude: undefined;
//     };
//     timeRange: 1;
// }

type t2 = v.InferOutput<typeof VSearchParams>;
// type t2 = {
//     units: "SI" | "IP";
//     selectedCoordinate: {
//         longitude: number;
//         latitude: number;
//     };
//     timeRange: number;
// }

const t3 = v.getFallback(v.fallback(
	v.object({
		longitude: coerceToNumber,
		latitude: coerceToNumber,
	}),
	{
		longitude: 30.2525,
		latitude: -97.7539,
	},
));
// const t3: {
//     readonly longitude: 30.2525;
//     readonly latitude: -97.7539;
// }
```

muningis:
Resolves #1150 

This is, semi breaking change though I would assume, as it changes default behaviour.
Previously, it would check for fallbacks of an object, and ignore if it has a fallback itself, eg.:
```ts
const ObjectWithFallback = v.fallback(
  v.object({ foo: string() }), <-- checks if entries of this object has a fallback
  { foo: "bar" }, <-- completely ignored
);
```

### DIFF
--- a/library/src/methods/getFallbacks/getFallbacks.ts
+++ b/library/src/methods/getFallbacks/getFallbacks.ts
@@ -64,6 +64,12 @@ export function getFallbacks<
         ErrorMessage<TupleWithRestIssue> | undefined
       >,
 >(schema: TSchema): InferFallbacks<TSchema> {
+  
+  // If it's and object schema and it has a fallback, return it
+  if (schema.type === 'object' && 'fallback' in schema) {
+    return schema.fallback as InferFallbacks<TSchema>;
+  }
+
   // If it is an object schema, return fallbacks of entries
   if ('entries' in schema) {
     const object: Record<string, unknown> = {};
--- a/library/src/methods/getFallbacks/getFallbacks.test.ts
+++ b/library/src/methods/getFallbacks/getFallbacks.test.ts
@@ -47,7 +47,7 @@ describe('getFallbacks', () => {
       });
     });
 
-    test('for nested object', () => {
+    test.only('for nested object', () => {
       expect(
         getFallbacks(
           object({
@@ -56,6 +56,7 @@ describe('getFallbacks', () => {
               key2: fallback(number(), () => 123 as const),
               key3: fallback(boolean(), false as const),
             }),
+            objectWithFallback: fallback(object({ foo: string() }), { foo: "bar" }),
             other: string(),
           })
         )
@@ -65,6 +66,7 @@ describe('getFallbacks', () => {
           key2: 123,
           key3: false,
         },
+        objectWithFallback: { foo: "bar" },
         other: undefined,
       });
     });


library/src/methods/getFallbacks/getFallbacks.ts
```diff
@@ -64,6 +64,12 @@ export function getFallbacks<
...
 >(schema: TSchema): InferFallbacks<TSchema> {
+  
+  // If it's and object schema and it has a fallback, return it
+  if (schema.type === 'object' && 'fallback' in schema) {
+    return schema.fallback as InferFallbacks<TSchema>;
+  }
+
   // If it is an object schema, return fallbacks of entries
   if ('entries' in schema) {
     const object: Record<string, unknown> = {};
```
muningis, line 75: Ideally, it would be like this, however `fallback` requires all values to be defined. 

```suggestion
  // If it is an object schema, return fallbacks of entries
  // If an object itself has a fallback, use it
  if ('entries' in schema) {
    const object: Record<string, unknown> = schema.type === 'object' && 'fallback' in schema ? schema.fallback : {};
```

eg.

```ts
const ObjectWithFallbackSchema = fallback(object({
  foo: number(),
  bar: fallback(number(), 5)
}), {
  foo: 3
});
```

I would expect `bar` to not be required, because it itself has a fallback.

muningis: It can be achieve by adjusting `Fallback` type to

```ts
/**
 * Fallback type.
 */
export type Fallback<
  TSchema extends BaseSchema<unknown, unknown, BaseIssue<unknown>>,
  TFallbackSchema = MaybeReadonly<InferOutput<TSchema>>
> =
  | {
    [key in keyof TFallbackSchema as TFallbackSchema[key] extends SchemaWithFallback<any, any> ? never : key]: TFallbackSchema[key];
    } & {
      [key in keyof TFallbackSchema as TFallbackSchema[key] extends SchemaWithFallback<any, any> ? key : never]?: TFallbackSchema[key];
    }
  | ((
      dataset?: OutputDataset<InferOutput<TSchema>, InferIssue<TSchema>>,
      config?: Config<InferIssue<TSchema>>
    ) => ({
      [key in keyof TFallbackSchema]: TFallbackSchema[key] extends SchemaWithFallback<any, any>
      ? TFallbackSchema[key]
      : TFallbackSchema[key];
    } & {
      [key in keyof TFallbackSchema as TFallbackSchema[key] extends SchemaWithFallback<any, any> ? key : never]?: TFallbackSchema[key];
    }));
```

but that's whole other discussion

library/src/methods/getFallbacks/getFallbacks.ts
```diff
@@ -64,6 +64,12 @@ export function getFallbacks<
         ErrorMessage<TupleWithRestIssue> | undefined
       >,
 >(schema: TSchema): InferFallbacks<TSchema> {
+  
+  // If it's and object schema and it has a fallback, return it
+  if (schema.type === 'object' && 'fallback' in schema) {
```
fabian-hiller, line 69: Isn't it enough to just check for `'fallback' in schema`?

muningis: Yeah, could be done that way too.

But probably this if clause can be removed, and do this when retrieving fallbacks of an object -> https://github.com/fabian-hiller/valibot/pull/1155#discussion_r2040726954

muningis:
Good point, definitely need to update the type too. Will do, once decided if fix/change should be implemented/added or not.

fabian-hiller:
If we merge this. Do we consider this a fix or a breaking change and park it in a new `v2` branch? I am not really sure. I really wonder why I implemented it the way I did.

muningis:
IMO it’s breaking change even if it’s a bugfix. Libraries can’t (shouldn’t be able to) afford such changes with minor/patch releases.

PR #1158, opened 2025-04-17
Add `summarize` method to summarize issues in multi-line string

fabian-hiller:


EskiMojo14:
could be interesting to have a similar util over in @standard-schema/utils - there's not really anything that's valibot specific here as far as i can tell

fabian-hiller:
Yes, great idea! I will focus on Valibot v1.1 and maybe v1.2 before reviewing and merging the Standard Schema utils PR.

EltonLobo07 (approved):

fabian-hiller:

vladfaust:
Personally, I don't like the opinionated choice to prepend `× ` to error messages. I have to do `.slice(2)` to get rid of the symbol. But it doesn't work for multiline errors...

fabian-hiller:
Hey @vladfaust if you prefer a custom format, I recommend creating your own `summarize` function. Feel free to copy our `summarize` function and use it as a template. The benefit of a modular design is that you can replace any function with custom code.

Issue #1128, opened 2025-04-04
extend the `multipleOf` action to support bigint inputs


EltonLobo07:
The `multipleOf` action works for number inputs only. It would be better if the action also supported bigint inputs. Specifically, this is desired:
```ts
import * as v from 'valibot';

// this works at the moment
const Schema1 = v.pipe(v.number(), v.multipleOf(7));

// this does not work at the moment (this is the issue) 
const Schema2 = v.pipe(v.bigint(), v.multipleOf(7n));

// these should continue to generate type errors after the issue is fixed
const Schema3 = v.pipe(v.number(), v.multipleOf(7n));
const Schema4 = v.pipe(v.bigint(), v.multipleOf(7));
```
This change will make `multipleOf` action very similar to Zod's `multipleOf` validation method

fabian-hiller:
Sounds good. Do you want to create a PR?

EltonLobo07:
Yes, I'll create a PR.

PR #1164, opened 2025-04-20
Add support for bigints to multipleOf validation action

fabian-hiller:
Solves #1128

### DIFF
--- a/library/src/actions/multipleOf/multipleOf.test.ts
+++ b/library/src/actions/multipleOf/multipleOf.test.ts
@@ -1,11 +1,7 @@
 import { describe, expect, test } from 'vitest';
 import type { NumberIssue } from '../../schemas/index.ts';
 import { expectActionIssue, expectNoActionIssue } from '../../vitest/index.ts';
-import {
-  multipleOf,
-  type MultipleOfAction,
-  type MultipleOfIssue,
-} from './multipleOf.ts';
+import { multipleOf, type MultipleOfAction } from './multipleOf.ts';
 
 describe('multipleOf', () => {
   describe('should return action object', () => {
@@ -45,8 +41,6 @@ describe('multipleOf', () => {
   });
 
   describe('should return dataset without issues', () => {
-    const action = multipleOf(5);
-
     test('for untyped inputs', () => {
       const issues: [NumberIssue] = [
         {
@@ -59,7 +53,7 @@ describe('multipleOf', () => {
         },
       ];
       expect(
-        action['~run']({ typed: false, value: null, issues }, {})
+        multipleOf(5)['~run']({ typed: false, value: null, issues }, {})
       ).toStrictEqual({
         typed: false,
         value: null,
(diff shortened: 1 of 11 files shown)


library/src/actions/multipleOf/multipleOf.ts
```diff
@@ -119,7 +119,8 @@ export function multipleOf(
     requirement,
     message,
     '~run'(dataset, config) {
-      if (dataset.typed && dataset.value % this.requirement !== 0) {
+      // @ts-expect-error
+      if (dataset.typed && dataset.value % this.requirement != 0) {
```
EltonLobo07, line 123: The usage of the `%` operator is unsafe. It can generate a runtime error as shown in the code block below.
```ts
const Schema = v.pipe(
    v.union([v.number(), v.bigint()]), 
    v.multipleOf(5)
);

v.safeParse(Schema, 10n); // Cannot mix BigInt and other types, use explicit conversions
```
We will have to also test if the types of the values match. Maybe something like shown in the code block below?
```ts
dataset.typed &&
(
    typeof dataset.value !== typeof this.requirement ||
    // @ts-expect-error
    dataset.value % this.requirement != 0
)
```

library/src/actions/multipleOf/multipleOf.ts
```diff
@@ -9,8 +9,8 @@ import { _addIssue } from '../../utils/index.ts';
  * Multiple of issue interface.
  */
 export interface MultipleOfIssue<
-  TInput extends number,
-  TRequirement extends number,
+  TInput extends number | bigint,
```
EltonLobo07, line 12: Maybe let's create an internal type for maintainability (Eg: `type Input = number | bigint`)?

library/src/utils/_stringify/_stringify.ts
```diff
@@ -13,6 +13,7 @@ export function _stringify(input: unknown): string {
   if (type === 'string') {
     return `"${input}"`;
   }
+  // TODO: Should we add "n" suffix to bigints?
```
EltonLobo07, line 16: I vote yes. It will make the stringified version of bigint and number type values a bit more clear. That would be a breaking change, right?

library/src/actions/multipleOf/multipleOf.test.ts
EltonLobo07, line 1: Also, do you think adding a few runtime tests related to the case shown in the first code block of [this comment](https://github.com/fabian-hiller/valibot/pull/1164#discussion_r2052292004) a good idea?

library/src/utils/_stringify/_stringify.ts
fabian-hiller: Yes, so the earliest we will change it will be Valibot v2.

library/src/actions/multipleOf/multipleOf.ts
fabian-hiller: How about:
```ts
dataset.typed &&
(typeof dataset.value === typeof this.requirement
  ? dataset.value % this.requirement != 0
  : Number(dataset.value) % Number(this.requirement) !== 0)
```

EltonLobo07: I don't think it's a good idea. The program might end up losing precision when converting bigints to numbers, which might lead to incorrect results. Take a look at the code block below:
```ts
const datasetValue = 18014398509481985n;

console.log(datasetValue % 3n); // 2
console.log(Number(datasetValue) % 3); // 1
```

fabian-hiller: Another idea would be to have extra TypeScript overload function signatures for `number` and `bigint`. That way TS will show an error if you mix them. I will push a PR with this change to demonstrate what I mean.

library/src/actions/multipleOf/multipleOf.test.ts
fabian-hiller: If we keep the additional overload signatures and disallow mixing of the two types, we don't need additional tests.

library/src/actions/multipleOf/multipleOf.ts
EltonLobo07: I'll try to summarize everything I understand:
- We intend to discourage `number | bigint` type inputs because even though there are some bigints that can be converted to numbers, not all bigints can reliably be converted to numbers. So there is no reliable way of having a common requirement value for `number | bigint` type inputs.
- Even though the solution provided [here](https://github.com/fabian-hiller/valibot/pull/1164#discussion_r2052292004) works, it does not provide a good user experience since the schema may not generate a type error, but the action can silently fail because of a type mismatch. So, it's better to flag this as a type error instead of a silent runtime fail.
- If the user really wants to mix `number | bigint` types, the recommended way to do this safely would be:
  ```ts
  const Schema = v.pipe(
    v.union([v.pipe(v.number(), v.integer()), v.bigint()]),
    v.transform((input) => BigInt(input)),
    v.multipleOf(5n)
  );
  ```
Are all of the points valid, and do they match the intentions behind the changes made? If yes, let me know. I'll do a final review (after you update the website) and merge the PR.

EltonLobo07 (approved):

fabian-hiller:

PR #1170, opened 2025-04-25
feat: create cache method for caching schema output by input

### LINKED ISSUE #923: Feature: `memoize` result of asynchronous schema
## Preface - valibot is awesome

Thank you all for developing this library.
It is the most type-safe library, I have ever used.
It comes with great DX & the docs become better and better.

## Use Case

Forms:
- Sometimes we need to validate certain fields asynchronously. 
- It would be nice if the schema could cache certain inputs and the results of the asynchronous validations to not trigger asynchronous operations if not needed.

## New action `memoize`

Wouldn't it be nice if we could `memoize` the executions inside a schema, where it is needed?
We could provide a synchronous and an asynchronous version... 

```ts
memoize
memoizeAsync
```

### Example

```ts
v.pipeAsync(
  v.string(),
  v.memoizeAsync(
    v.customAsync(input => fetch(`/api/email/${input}`).then(() => /* ... */)), { 
    buffer: 10, // default 1 - tells how many inputs with their respective result are cached
    itemExpiryInMs: 30_000  // default 5_000
  })
)
```

EskiMojo14:
fixes #923 

```ts
const cachedSchema = v.cache(baseSchema, { maxSize: 100, duration: 5_000 })
```


library/src/utils/_Cache/_Cache.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * A basic cache with optional max size and expiry.
+ */
+export class _Cache<TKey, TValue> {
+  private maxSize: number;
+  private duration: number | undefined;
+  private cache = new Map<TKey, _CacheEntry<TValue>>();
+  constructor(options?: _CacheOptions) {
+    this.maxSize = options?.maxSize ?? 1;
```
fabian-hiller, line 19: Should `maxSize` be `Infinity` if not set?

library/src/utils/_Cache/_Cache.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.duration = options?.duration;
+  }
+  private scheduleExpiry(key: TKey) {
+    if (!this.duration) return;
+    return setTimeout(() => this.cache.delete(key), this.duration);
```
fabian-hiller, line 24: I wonder if it is more efficient to create a timeout for each entry or to store `Date.now()` in the entry and check it before returning the cached value.

library/src/utils/_Cache/_Cache.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (entry) {
+      // reset expiry
+      clearTimeout(entry.expiryId);
+      entry.expiryId = this.scheduleExpiry(key);
+      return entry.value;
+    }
+  }
+  set(key: TKey, value: TValue): void {
+    if (this.cache.size >= this.maxSize) {
+      this.delete(this.cache.keys().next().value!);
```
fabian-hiller, line 40: Does this delete entries in a specific order or randomly?

fabian-hiller (commented):
Thanks for creating this PR. I will probably plan it for v1.2 or v1.3.

library/src/utils/_Cache/_Cache.ts
EskiMojo14: a default of 1 is pretty common - for example reselect and proxy-memoize

library/src/utils/_Cache/_Cache.ts
EskiMojo14: not sure, but this way means we're not holding the reference for longer than we need

library/src/utils/_Cache/_Cache.ts
EskiMojo14: map entries are always iterated in order of insertion, so the first key is the oldest - though calling .set again will not bring it to the end, since it's updated not inserted (not an issue here since we only set if we don't already have a cached value)

EskiMojo14: we *could* make .set always delete the existing entry first, if we wanted to be extra safe - i don't know how much that would impact performance

library/src/utils/_Cache/_Cache.ts
fabian-hiller: But doesn't this mean that `cache` is pretty useless if you don't specify `maxSize`?

muningis: Not really. If 2nd run is done with same value, it wouldn’t re-evaluate, just use cached result

library/src/utils/_Cache/_Cache.ts
muningis: I would agree on keeping date entry, and invalidate it on accessing. Ensures that it isn’t invalidated too late or too early. Settimeout is flaky

fabian-hiller: True. Do we know how other libraries handle this case? For serverless environments, this could cause problems when executing code after the function has finished, and I am concerned that creating many timeouts will have an impact on performance and/or memory usage. But I could be wrong.

library/src/utils/_Cache/_Cache.ts
fabian-hiller: I think deleting the oldest entry makes sense.

muningis: It could also be configurable - FIFO vs LRU.

fabian-hiller: True, but I like to keep things simple. To get started, we should choose what makes the most sense in terms of overall performance. We can always add a new configuration later if we really need to.

library/src/utils/_Cache/_Cache.ts
fabian-hiller: I am open to discussing this, but right now I would probably choose a default that makes the most sense, so that people only need to define a custom value for special needs.

library/src/utils/_Cache/_Cache.ts
EskiMojo14: we could also define a common interface for what we need, and allow the user to provide a custom constructor if they really want to customise behaviour, similar to lodash.memoize

something like 
```ts
interface CacheOptions {
  maxSize?: number;
  duration?: number;
}
interface CacheInstance<K, V> {
  get(key: K): V | undefined;
  set(key: K, value: V): void;
}
type CacheConstructor = new <K, V>(opts: CacheOptions | undefined) => CacheInstance<K, V>
```

library/src/utils/_Cache/_Cache.ts
EskiMojo14: for memoize libraries i think a lot of people leave it as 1, but i guess it could be argued a schema is more likely to be called with different values

library/src/utils/_Cache/_Cache.ts
EskiMojo14: seems to be a variety of different approaches - removing on access, polling, etc

setTimeout seemed the most intuitive to me but perhaps also the most naive

library/src/utils/_Cache/_Cache.ts
EskiMojo14: image
hmm, maybe not possible

EskiMojo14: oh, or just accepting the instance rather than the constructor 🤦🏻

library/src/utils/_Cache/_Cache.ts
EskiMojo14: changed to 1000

fabian-hiller:
Thank you! If I understand correctly, the cache only works for primitive values or complex values with the same reference. Providing a new object as the `dataset.value` would result in no caching. Is this correct?

EskiMojo14:
> Thank you! If I understand correctly, the cache only works for primitive values or complex values with the same reference. Providing a new object as the `dataset.value` would result in no caching. Is this correct?

correct, the default cache compares by reference - the user is free to provide a cache implementation with more complex comparisons as long as the same get/set interface is there

fabian-hiller:
@coderabbitai can you review this PR?

fabian-hiller:
I think I would like to simplify this PR a little bit. For example, I think I would not allow people to pass a custom cache implementations. I don't like that we have an argument that can either be a config object or a cache instance. I intentionally not do that anywhere else in the library to keep the API and implementation simple. If people actually request a custom cache we can reevaluate it and in the meantime they can build their own cache method if needed. Feel free to give me feedback on this. I can take over this PR if you like.

fabian-hiller:
I implemented a simpler cache2 method. Let me know what you think. I would love to merge this PR before releasing Valibot v1.3.

EskiMojo14:
> I think I would like to simplify this PR a little bit. For example, I think I would not allow people to pass a custom cache implementations. I don't like that we have an argument that can either be a config object or a cache instance. I intentionally not do that anywhere else in the library to keep the API and implementation simple. If people actually request a custom cache we can reevaluate it and in the meantime they can build their own cache method if needed. Feel free to give me feedback on this. I can take over this PR if you like.

sure, up to you - though i don't think supporting custom caches complicated the implementation that much personally

fabian-hiller:
Ok. I will simplify and finalize the implementation. We can always come back to this PR if we want to extend it later on.

fabian-hiller:
@codex review

library/src/methods/cache/_LruCache.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+    if (type === 'object' || type === 'function') {
+      if (input) {
+        this.refIds ??= new WeakMap();
+        let id = this.refIds.get(input as WeakKey);
+        if (!id) {
+          id = ++this.refCount;
+          this.refIds.set(input as WeakKey, id);
+        }
+        return `#${id}`;
```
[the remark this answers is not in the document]
fabian-hiller: True but what can we do about it. If we keep it this way we need to clearly document this.

library/src/methods/cache/cache.ts
```diff
@@ -0,0 +1,85 @@
...
+    get '~standard'() {
+      return _getStandardProps(this);
+    },
+    '~run'(dataset, runConfig) {
+      const key = this.cache.key(dataset.value, runConfig);
+      let outputDataset = this.cache.get(key);
+      if (!outputDataset) {
+        this.cache.set(
+          key,
+          (outputDataset = schema['~run'](dataset, runConfig))
```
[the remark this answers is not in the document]
fabian-hiller: We probably needs to store a copy or return a copy of the dataset to prevent that

library/src/methods/cache/cacheAsync.ts
```diff
@@ -0,0 +1,151 @@
...
+      if (!promise) {
+        activeRuns ??= new Map();
+        promise = Promise.resolve(schema['~run'](dataset, runConfig));
+        activeRuns.set(key, promise);
+      }
+
+      // Await pending promise, cache output and return
+      try {
+        const outputDataset = await promise;
+        this.cache.set(key, outputDataset);
```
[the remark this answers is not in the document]
fabian-hiller: We probably needs to store a copy or return a copy of the dataset to prevent that

fabian-hiller:

PR #1171, opened 2025-04-26
fix: update nanoid to be more accurate

jasperteo:
Looking at my old commit, I realized that the old regex was partially wrong due to ESLint autofix. The old regex was less strict and allowed for certain characters that are invalid in Nano ID.

This PR also standardizes all the types and interface with regards to Nano ID to PascalCase.

### DIFF
--- a/library/src/regex.ts
+++ b/library/src/regex.ts
@@ -135,9 +135,7 @@ export const MAC_REGEX: RegExp =
 /**
  * [Nano ID](https://github.com/ai/nanoid) regex.
  */
-export const NANO_ID_REGEX: RegExp =
-  // eslint-disable-next-line regexp/prefer-w
-  /^[A-Za-z0-9_-]+$/u;
+export const NANO_ID_REGEX: RegExp = /^[\w-]+$/u;
 
 /**
  * [Octal](https://en.wikipedia.org/wiki/Octal) regex.
--- a/library/src/regex.ts
+++ b/library/src/regex.ts
@@ -135,7 +135,9 @@ export const MAC_REGEX: RegExp =
 /**
  * [Nano ID](https://github.com/ai/nanoid) regex.
  */
-export const NANO_ID_REGEX: RegExp = /^[\w-]+$/u;
+export const NANO_ID_REGEX: RegExp =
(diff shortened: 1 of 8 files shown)


library/src/regex.ts
```diff
@@ -135,7 +135,9 @@ export const MAC_REGEX: RegExp =
 /**
  * [Nano ID](https://github.com/ai/nanoid) regex.
  */
-export const NANO_ID_REGEX: RegExp = /^[\w-]+$/u;
+export const NANO_ID_REGEX: RegExp =
+  // eslint-disable-next-line regexp/prefer-w
+  /^[A-Za-z0-9_-]+$/u;
```
fabian-hiller, line 140: `\w` should be equal to `[a-zA-Z0-9_]`. So `[\w-]` should be equal to `[A-Za-z0-9_-]` or am I wrong? Is there an invalid case that is not covered by our unit tests?

fabian-hiller (commented):
Rest looks good. Thank you for this PR!

library/src/regex.ts
jasperteo: Yeah in JavaScript/TypeScript it is the same and will pass all unit tests. Valibot will still fail the invalid characters.

It just happenes when I copied `/^[\w-]+$/u` to use outside JavaScript/TypeScript, it will allow for invalid characters such as `é`, so I thought to update it to a more accurate `/^[A-Za-z0-9_-]+$/u`.

Functionally nothing will change at all.

fabian-hiller: The main reason I prefer `\w` is Valibot's focus in bundle size. So if we can write something with less chars we should consider it. If there is no difference in JS/TS I would probably keep the regex as it is.

fabian-hiller: If we can make a decision and merge this PR tomorrow, I will include this change in the Valibot v1.1 release.

jasperteo: After some thought I think we should go with the old `/^[\w-]+$/u` instead

fabian-hiller (approved):

fabian-hiller:

PR #1195, opened 2025-05-15
Action `isoTimestamp` allows space in UTC offset

frenzzy:
Fixes:
- https://github.com/fabian-hiller/valibot/issues/495

Changes:
- Updated ISO 8601 timestamp regex to allow a single space before timezone offset (e.g., `+00:00`).

Reason:
- PostgreSQL’s `timestamptz` default output includes a space before the timezone offset:
   ```sql
   SELECT NOW()::timestamptz;
   -- → '2025-05-13 14:15:41.123904 +00:00'
   ```
- This aligns with the SQL standard output format, even though it’s not strictly ISO 8601 compliant.
- The change improves compatibility with real PostgreSQL data and prevents false validation errors.


fabian-hiller:
Thank you for creating this PR! Are you aware of any drawbacks? I'm a bit concerned because it doesn't seem to be ISO 8601 compliant.

frenzzy:
Thank you for asking! I agree this tweak isn’t strictly ISO 8601, but I don’t anticipate any serious drawbacks, mainly because it matches common real-world usage. For example, PostgreSQL’s `timestamptz` has for years output a space before the offset by default (following the SQL standard formatting). This means strings like `"2025-05-13 14:15:41.123904 +00:00"` are already out there – even modern JavaScript engines handle that format just fine (`Date.parse()` / `new Date()` don’t error on it in my testing).

Also, our current `ISO_TIMESTAMP_REGEX` isn’t a fully strict ISO 8601 validator anyway – we’ve already allowed a space instead of `"T"` between date and time, and the timezone offset was designed to be optional. In that light, permitting a leading space before the `+/-` offset is consistent with our pragmatic approach. It lets developers validate timestamps exactly as PostgreSQL emits them without having to pre-process the string or write a custom regex.

In short, this change improves compatibility with real data (avoiding false validation errors on perfectly valid SQL-standard timestamps) while staying aligned with the library’s goal of being practical and developer-friendly.

fabian-hiller:
Sorry for the delay! I will try to look into it when working on Valibot v1.3.

fabian-hiller:
@coderabbitai please review this PR

fabian-hiller:
Valibot v1.3 is available. Feel free to submit and expense of $15 for all your contributions for Valibot v1.3 via Open Collective.

PR #1199, opened 2025-05-16
feat: add examples action and getExamples metadata

EskiMojo14:
follows up from #1189 

Currently any value is allowed in the array - it could make sense to limit it to only valid inputs/outputs, though I'm not sure how the action would infer those.

When multiple example actions are in a pipe, the arrays are merged depth first.


fabian-hiller:
I will finlaize this PR

fabian-hiller:

PR #1204, opened 2025-05-25
feat: add guard action for narrowing using type predicates

### LINKED ISSUE #1089: Check action to infer return type from a passed typeguard
Here is a simple example. Let's say we have a typeguard that checks if a string starts with hash symbol: 
```ts
export const checkStartsWithHash = (target: string): target is `#${string}` => target.startsWith('#')
```
What I would expect after passing that typeguard as an argument of `check` action is something like:
```ts
const input: string = '#fff'
const hashString = parse(
  pipe(string(), check(checkStartsWithHash)),
  input
) 
type Test = typeof hashString // `#${string}`
```
But instead of `#${string}` I get the initial input type with no consideration for the passed typeguard.
This could be previously bypassed with type casting: `check<TypeToCast>(...)`, but with the latest update, it doesn't work anymore.

Is it something that you would consider adding? 
Thanks!

EskiMojo14:
fixes #1089


EskiMojo14:
found a fix (involving lots of `any`, unfortunately) - should be ready for review :)

fabian-hiller:
This PR should be ready now. I will merge it soon.

fabian-hiller:

PR #1206, opened 2025-05-27
add zod `tuple` transform

niconiahi:


niconiahi:
I've finished the implementation for `tuple` , although it escapes current normal code flow, it creates an additional logical branch for when the transforming an schema with `property === "tuple"`

I think I'll try another version where I send additional data, as done with `coerce`, maybe this `coerceOption` argument for `toValibotSchemaExp` could become an `options` and grow as needed? 

this is the current state

niconiahi:
I pushed a new version. I think in this version its readability is increased. It uses the most it can, the existing code

PR #1212, opened 2025-05-31
feat: create coercion actions for primitives (and date)

EskiMojo14:
Decisions left:
- should we try/catch around conversions? it's rare but some can throw, like `Number(Symbol())`
- v.toNumber - should we check for NaN after converting, or recommend piping into v.number?
- v.toDate - similarly, should we check the date is valid, or recommend piping into v.date?

### DIFF
--- a/website/src/routes/api/(actions)/toBigint/index.mdx
+++ b/website/src/routes/api/(actions)/toBigint/index.mdx
@@ -52,7 +52,7 @@ The following APIs can be combined with `toBigint`.
 
 ### Schemas
 
-<ApiList items={['any', 'custom', 'number', 'string']} />
+<ApiList items={['any', 'boolean', 'custom', 'number', 'string', 'unknown']} />
 
 ### Methods
 
--- a/website/src/routes/api/(actions)/toBoolean/index.mdx
+++ b/website/src/routes/api/(actions)/toBoolean/index.mdx
@@ -43,7 +43,20 @@ The following APIs can be combined with `toBoolean`.
 
 ### Schemas
 
-<ApiList items={['any', 'custom', 'string']} />
+<ApiList
+  items={[
+    'any',
+    'bigint',
+    'custom',
+    'date',
+    'null',
+    'number',
+    'string',
+    'symbol',
+    'undefined',
+    'unknown',
+  ]}
+/>
 
 ### Methods
 
--- a/website/src/routes/api/(actions)/toDate/index.mdx
+++ b/website/src/routes/api/(actions)/toDate/index.mdx
@@ -52,7 +52,7 @@ The following APIs can be combined with `toDate`.
 
 ### Schemas
 
-<ApiList items={['any', 'custom', 'string']} />
+<ApiList items={['any', 'custom', 'number', 'string', 'unknown']} />
 
 ### Methods
 
--- a/website/src/routes/api/(actions)/toNumber/index.mdx
+++ b/website/src/routes/api/(actions)/toNumber/index.mdx
@@ -52,7 +52,18 @@ The following APIs can be combined with `toNumber`.
 
 ### Schemas
 
-<ApiList items={['any', 'custom', 'string']} />
+<ApiList
+  items={[
+    'any',
+    'bigint',
+    'boolean',
+    'custom',
+    'date',
+    'null',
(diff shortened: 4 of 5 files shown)


library/src/actions/toBigint/toBigint.ts
```diff
@@ -0,0 +1,39 @@
...
+  TInput extends bigint | boolean | number | string,
+>(): ToBigintAction<TInput> {
+  return {
+    kind: 'transformation',
+    type: 'to_bigint',
+    reference: toBigint,
+    async: false,
+    '~run'(dataset) {
+      // @ts-expect-error
+      dataset.value = BigInt(dataset.value);
```
fabian-hiller: I think we need `try/catch` here. An example edge case is `BigInt({})`.

fabian-hiller:
Thank you so much!

> should we try/catch around conversions? it's rare but some can throw, like Number(Symbol())

Yes, we should do that whenever it is necessary to cover all edge cases

> v.toNumber - should we check for NaN after converting, or recommend piping into v.number?
> v.toDate - similarly, should we check the date is valid, or recommend piping into v.date?

I would only convert it. Do you know what Zod is doing here?

EskiMojo14:
with Zod the coercion isn't separate from the schema (it just happens as the first step of the schema if configured), so `z.coerce.number()` would be the same as `v.pipe(v.unknown(), v.toNumber(), v.number())`

library/src/actions/toDate/toDate.ts
```diff
@@ -0,0 +1,65 @@
...
+    reference: toDate,
+    async: false,
+    '~run'(dataset, config) {
+      try {
+        // @ts-expect-error
+        dataset.value = new Date(dataset.value);
+      } catch {
+        _addIssue(this, 'date', dataset, config);
+        // @ts-expect-error
+        dataset.typed = false;
```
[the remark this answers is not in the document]
fabian-hiller: I think right now we want devs to explicitly handle this case with `v.pipe(v.unknown(), v.toDate(), v.date())`.

website/src/routes/api/(actions)/toBigint/properties.ts
```diff
@@ -0,0 +1,63 @@
...
+        {
+          type: 'custom',
+          name: 'TMessage',
+        },
+      ],
+    },
+  },
+};
+
+export default properties;
```
alexilyaev: This `export default` does not align with the rest of the properties files which use `export const properties`.
This also fails because the `mdx` file expects `import { properties } from './properties'`

website/src/routes/api/(actions)/toBigint/index.mdx
```diff
@@ -0,0 +1,63 @@
...
+- `TInput` <Property {...properties.TInput} />
+- `TMessage` <Property {...properties.TMessage} />
+
+## Parameters
+
+- `message` <Property {...properties.message} />
+
+### Explanation
+
+With `toBigint` you can transform the input to a bigint. If the input cannot be transformed, you can use `message` to customize the error message.
```
alexilyaev, line 31: This says the user can pass a `message` but the code at `library/src/actions/toBigint/toBigint.ts` does not actually accept that argument.

website/src/routes/api/(actions)/toBigint/properties.ts
EskiMojo14: good spot, fixed

website/src/routes/api/(actions)/toBigint/index.mdx
EskiMojo14: good catch, added message argument to toBigint

fabian-hiller:
I will start working on this PR to get it over the line

fabian-hiller:
> v.toNumber - should we check for NaN after converting, or recommend piping into v.number?
> v.toDate - similarly, should we check the date is valid, or recommend piping into v.date?

I decided to return an issue for both as I think that this is the happy path for most users.

Feel free to review my changes.

fabian-hiller:
I will finalize and merge this PR in the next days

EskiMojo14:
changes look good, thank you 😄

fabian-hiller:

PR #1220, opened 2025-06-08
fix(to-json-schema): do not mark exactOptional entry as required JSON property

sruenwg:
This PR fixes `convertSchema` to no longer mark `exactOptional` object properties as required properties in the output JSON Schema.
See [playground example](https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgChRJY4BvOGCAKVQgDsAygGMAFgFMQyOAF8ceAoQACJclRgB6TgFoAVrz5bUoicjr16Q-qnjDxkuAF5MAOggUdYoTAAUrenDgAawBGAC4XCDAYYH4ybwxnPgBXEAoxKG8ASkyAGn8ggCZwhLEAD2QvAHkomL44hOsoYD4Acyzc+mlMhgsreCgxVCTSeCcG5GwxAAVkKFQxb1sTHLYuhks+XlIxZ1IINoGhke7zDes4PX4l+ydOHivjSUXH5BOziG3d-e9LwReToA) — `{}` is clearly a valid value according to the schema, but the generated JSON Schema would reject this as it requires the `k2` property.

### DIFF
--- a/packages/to-json-schema/CHANGELOG.md
+++ b/packages/to-json-schema/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to the library will be documented in this file.
 ## vX.X.X (Month DD, YYYY)
 
 - Add support for `examples` action
+- Fix conversion of `exactOptional` object properties (pull request #1220)
 
 ## v1.3.0 (June 01, 2025)
 


fabian-hiller:
Thank you for your contribution! Not sure how we could miss that. I will merge it before releasing the next version of the package.

sruenwg:
Thanks! 👍

PR #1261, opened 2025-07-27
feat: add `or` migration

paoloricciuti:
Currently migrating `or` produces a runtime error

```ts
import * as z from 'zod';

const Schema = z.string().or(z.number())
```

proudces

```ts
import * as v from "valibot";

const Schema = v.string()(v.number())
```

this fixes it migrating or to an union.

### DIFF
--- a/codemod/zod-to-valibot/__testfixtures__/or/input.ts
+++ b/codemod/zod-to-valibot/__testfixtures__/or/input.ts
@@ -1,0 +1,3 @@
+import { z } from "zod";
+
+const Schema1 = z.string().or(z.number());
--- a/codemod/zod-to-valibot/__testfixtures__/or/input.ts
+++ b/codemod/zod-to-valibot/__testfixtures__/or/input.ts
@@ -1,3 +1,4 @@
 import { z } from "zod";
 
 const Schema1 = z.string().or(z.number());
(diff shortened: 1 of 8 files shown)


codemod/zod-to-valibot/__testfixtures__/or/output.ts
```diff
@@ -0,0 +1,3 @@
+import * as v from "valibot";
+
+const Schema1 = v.union(v.string(), v.number());
```
fabian-hiller: The array is missing: v.union([v.string(), v.number()])

codemod/zod-to-valibot/__testfixtures__/or/input.ts
```diff
@@ -0,0 +1,3 @@
+import { z } from "zod";
+
+const Schema1 = z.string().or(z.number());
```
fabian-hiller: Does it work with multiple `.or` in the same chain? We should probably add a test cases for it as well.

codemod/zod-to-valibot/__testfixtures__/or/output.ts
paoloricciuti: Dang...should've actually tested the code and not just rely on me remembering 😄

PR #1262, opened 2025-07-27
fix: properly migrate `strict` and `passthrough`

paoloricciuti:
Currently `strict` migrate to the wrong code as you can see from the test input and output:

input
```ts
import { z } from "zod";

const Schema1 = z.object({key: z.string()}).strict();
const Schema2 = z.object({key: z.string()}, {message: "some message"}).strict();
const Schema3 = z.object({key: z.string()}, {description: "some description"}).strict();
const Schema4 = z.object({key: z.string()});
const Schema5 = Schema4.strict();
```

output
```ts
const Schema1 = v.strictObject({key: v.string()});
const Schema2 = v.strictObject({key: v.string()}, "some message");
const Schema3 = v.pipe(v.object({key: v.string()}), v.description("some description"), v.strict());
const Schema4 = v.object({key: v.string()});
const Schema5 = v.pipe(Schema4, v.strict());
```

`v.strict()` is not a thing.

this is the new output
```ts
const Schema8 = v.object({key: v.string()});
const Schema9 = v.object({key: v.string()}, "some message");
const Schema10 = v.object(
  v.pipe(v.object({key: v.string()}), v.description("some description")).entries
);
const Schema11 = v.strictObject({key: v.string()});
const Schema12 = v.pipe(Schema11, v.strip());
const Schema13 = v.object({key: v.string()});
const Schema14 = v.object(Schema13.entries);
```

### DIFF
--- a/codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
+++ b/codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
@@ -23,7 +23,7 @@ const Schema13 = v.strictObject(v.extend(Schema2, {bar: v.string()}).entries);
 const Schema14 = v.extend(v.object({foo: v.string()}), {bar: v.number()});
 const Schema15 = v.object({foo: v.number()});
 const Schema16 = v.extend(Schema15, {bar: v.string()});
-const Schema17 = v.object(v.extend(v.strictObject({foo: v.string()}), {bar: v.number()}).entries);
+const Schema17 = v.object(v.extend(v.object({foo: v.string()}), {bar: v.number()}).entries);
 const Schema18 = v.object(v.extend(Schema2, {bar: v.string()}).entries);
 
 // catchall
--- a/codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
+++ b/codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
@@ -9,21 +9,21 @@ const Schema3 = v.extend(Schema2, {bar: v.string()});
 const Schema4 = v.extend(v.looseObject({foo: v.string()}), {bar: v.number()});
 const Schema5 = v.looseObject({foo: v.number()});
 const Schema6 = v.extend(Schema5, {bar: v.string()});
-const Schema7 = v.pipe(v.extend(v.object({foo: v.string()}), {bar: v.number()}), v.passthrough());
-const Schema8 = v.pipe(v.extend(Schema2, {bar: v.string()}), v.passthrough());
+const Schema7 = v.looseObject(v.extend(v.looseObject({foo: v.string()}), {bar: v.number()}).entries);
+const Schema8 = v.looseObject(v.extend(Schema2, {bar: v.string()}).entries);
 
 // strict
 const Schema9 = v.extend(v.strictObject({foo: v.string()}), {bar: v.number()});
 const Schema10 = v.strictObject({foo: v.number()});
 const Schema11 = v.extend(Schema10, {bar: v.string()});
-const Schema12 = v.strictObject(v.extend(v.object({foo: v.string()}), {bar: v.number()}).entries);
+const Schema12 = v.strictObject(v.extend(v.strictObject({foo: v.string()}), {bar: v.number()}).entries);
 const Schema13 = v.strictObject(v.extend(Schema2, {bar: v.string()}).entries);
 
 // strip
 const Schema14 = v.extend(v.object({foo: v.string()}), {bar: v.number()});
 const Schema15 = v.object({foo: v.number()});
 const Schema16 = v.extend(Schema15, {bar: v.string()});
-const Schema17 = v.object(v.extend(v.object({foo: v.string()}), {bar: v.number()}).entries);
+const Schema17 = v.object(v.extend(v.strictObject({foo: v.string()}), {bar: v.number()}).entries);
 const Schema18 = v.object(v.extend(Schema2, {bar: v.string()}).entries);
 
 // catchall
--- a/codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
+++ b/codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
@@ -1,71 +1,203 @@
 import * as v from "valibot";
 
 // plain
-const Schema1 = v.extend(v.object({foo: v.string()}), {bar: v.number()});
+const Schema1 = v.object({
+  foo: v.string(),
+  bar: v.number()
+});
 const Schema2 = v.object({foo: v.number()});
-const Schema3 = v.extend(Schema2, {bar: v.string()});
+const Schema3 = v.object({
+  .../*@valibot-migrate we can't detect if Schema2 has a `pipe` operator, if it does you might need to migrate this by hand otherwise it will loose it's pipeline*/
+  Schema2.entries,
+
+  bar: v.string()
+});
 
 // passthrough
-const Schema4 = v.extend(v.looseObject({foo: v.string()}), {bar: v.number()});
+const Schema4 = v.object({
+  ...v.looseObject({foo: v.string()}).entries,
+  bar: v.number()
+});
 const Schema5 = v.looseObject({foo: v.number()});
-const Schema6 = v.extend(Schema5, {bar: v.string()});
-const Schema7 = v.looseObject(v.extend(v.looseObject({foo: v.string()}), {bar: v.number()}).entries);
-const Schema8 = v.looseObject(v.extend(Schema2, {bar: v.string()}).entries);
+const Schema6 = v.object({
+  .../*@valibot-migrate we can't detect if Schema5 has a `pipe` operator, if it does you might need to migrate this by hand otherwise it will loose it's pipeline*/
+  Schema5.entries,
+
+  bar: v.string()
+});
+const Schema7 = v.looseObject({
+  ...v.looseObject({foo: v.string()}).entries,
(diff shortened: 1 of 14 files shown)


codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
```diff
@@ -16,19 +16,15 @@ const Schema8 = v.pipe(v.extend(Schema2, {bar: v.string()}), v.passthrough());
...
 const Schema15 = v.object({foo: v.number()});
 const Schema16 = v.extend(Schema15, {bar: v.string()});
-const Schema17 = v.pipe(
-  v.extend(v.object({foo: v.string()}), {bar: v.number()}),
-  v.strict(),
-  v.strip()
-);
-const Schema18 = v.pipe(v.extend(Schema2, {bar: v.string()}), v.strict(), v.strip());
+const Schema17 = v.object(v.extend(v.object({foo: v.string()}), {bar: v.number()}).entries);
+const Schema18 = v.object(v.extend(Schema2, {bar: v.string()}).entries);
```
fabian-hiller: Something looks wrong here. We don't have an `v.extend()` function. We have [`v.entriesFromObjects()`](https://valibot.dev/api/entriesFromObjects/) but this is more a utility function. We could also just spread the `.entries`. Spreading is probably cleaner for the codemod but I don't mind if it is simpler to implement with `v.entriesFromObjects()`.

codemod/zod-to-valibot/__testfixtures__/object-strict/output.ts
```diff
@@ -2,6 +2,8 @@ import * as v from "valibot";
 
 const Schema1 = v.strictObject({key: v.string()});
 const Schema2 = v.strictObject({key: v.string()}, "some message");
-const Schema3 = v.pipe(v.object({key: v.string()}), v.description("some description"), v.strict());
+const Schema3 = v.strictObject(
+  v.pipe(v.object({key: v.string()}), v.description("some description")).entries
```
fabian-hiller: This does not work. We are losing the `pipe` this way. Better would be:

```ts
v.pipe(v.strictObject({key: v.string()}), v.description("some description"));
```

codemod/zod-to-valibot/__testfixtures__/object-strict/output.ts
```diff
@@ -2,6 +2,8 @@ import * as v from "valibot";
...
+);
 const Schema4 = v.object({key: v.string()});
-const Schema5 = v.pipe(Schema4, v.strict());
\ No newline at end of file
+const Schema5 = v.strictObject(Schema4.entries);
```
fabian-hiller: This works but if people have the `v.object` of `Schema4` wrapped in a `v.pipe` we will lose the pipe. The codemod will never be 100% perfect. So I am ok with merging this.

codemod/zod-to-valibot/__testfixtures__/object-strip/output.ts
```diff
@@ -17,13 +17,10 @@ const Schema7 = v.pipe(Schema6, v.passthrough(), v.strip());
 // strict
 const Schema8 = v.object({key: v.string()});
 const Schema9 = v.object({key: v.string()}, "some message");
-const Schema10 = v.pipe(
-  v.object({key: v.string()}),
-  v.description("some description"),
-  v.strict(),
-  v.strip()
+const Schema10 = v.object(
+  v.pipe(v.object({key: v.string()}), v.description("some description")).entries
```
fabian-hiller, line 21: This should probably be `v.pipe(v.object({key: v.string()}), v.description("some description"))` but I know that it could be complicated to solve it. Basically, depending of the last Zod object modifier (`strict`, `strip`, ...) we need to choose the right Valibot schema (`looseObject`, `object` or `strictObject`).

fabian-hiller (changes requested):

codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
fabian-hiller: `v.entriesFromObjects()` exists for the rare case where you are only interested in the produced TS type but don't use the schema for validation. `v.entriesFromObjects()` is a tree-shaking optimization. If you never use the schema at runtime bundlers like Rollup will remove it.

codemod/zod-to-valibot/__testfixtures__/object-strict/output.ts
fabian-hiller: This is related: https://github.com/fabian-hiller/valibot/pull/1262#discussion_r2234090933

codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
paoloricciuti: This is another broken thing i plan to fix in a different pr...but the test needs to be updated because it's using `strict`

codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
```diff
@@ -9,26 +9,22 @@ const Schema3 = v.extend(Schema2, {bar: v.string()});
...
 const Schema14 = v.extend(v.object({foo: v.string()}), {bar: v.number()});
 const Schema15 = v.object({foo: v.number()});
 const Schema16 = v.extend(Schema15, {bar: v.string()});
-const Schema17 = v.pipe(
-  v.extend(v.object({foo: v.string()}), {bar: v.number()}),
-  v.strict(),
-  v.strip()
-);
-const Schema18 = v.pipe(v.extend(Schema2, {bar: v.string()}), v.strict(), v.strip());
+const Schema17 = v.object(v.extend(v.strictObject({foo: v.string()}), {bar: v.number()}).entries);
```
fabian-hiller: Should we add the @valibot-migrate comment here too where we are accessing `.entries`?

codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
```diff
@@ -9,26 +9,22 @@ const Schema3 = v.extend(Schema2, {bar: v.string()});
...
+const Schema18 = v.object(v.extend(Schema2, {bar: v.string()}).entries);
 
 // catchall
 const Schema19 = v.extend(v.objectWithRest({foo: v.string()}, v.null()), {bar: v.number()});
```
fabian-hiller: The `extend` does not exists in Valibot but I think we fix that in another PR as this was already wrong before.

codemod/zod-to-valibot/__testfixtures__/object-strip/output.ts
```diff
@@ -4,26 +4,30 @@ import * as v from "valibot";
 const Schema1 = v.object({key: v.string()});
 const Schema2 = v.object({key: v.string()}, "some message");
 const Schema3 = v.pipe(
-  v.object({key: v.string()}),
+  v.looseObject({key: v.string()}),
   v.description("some description"),
-  v.passthrough(),
   v.strip()
```
fabian-hiller: This looks wrong. It should use the last call of `passthrough`, `strip` or `strict` and ignore and remove the previous calls to this modifier.

fabian-hiller: So this should output `v.pipe(v.object({key: v.string()}), v.description("some description"))` because `strip` was the last modifier call.

codemod/zod-to-valibot/__testfixtures__/object-strip/output.ts
```diff
@@ -4,26 +4,30 @@ import * as v from "valibot";
...
 
 // strict
 const Schema8 = v.object({key: v.string()});
 const Schema9 = v.object({key: v.string()}, "some message");
 const Schema10 = v.pipe(
-  v.object({key: v.string()}),
+  v.strictObject({key: v.string()}),
   v.description("some description"),
-  v.strict(),
   v.strip()
```
fabian-hiller: Same here

fabian-hiller:
Other then that, I think the PR looks good.

codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
paoloricciuti: Yeah this is again just updating tests to keep things in separate PRs

codemod/zod-to-valibot/__testfixtures__/object-extend/output.ts
paoloricciuti: I think we are gonna handle this when we fix `extend`

codemod/zod-to-valibot/__testfixtures__/object-strip/output.ts
paoloricciuti: Uhhhh didn't notice this...let me check quickly

paoloricciuti: Mmm i don't think this is completely correct tho right? I think `.passthrough().strip()` will nullify `passthrough` but `.strict().strip()` shouldn't no? Because you first go over `strict` and then try to strip (but `strict` will already error before)...not super familiar with how `zod` handles this.

fabian-hiller: I am not totally sure but I think Zod just takes the last modifier and ignores all the other calls before that. I think `strip` is the default behaviour and the others or similar to `looseObject` and `strictObject` in Valibot.

PR #1269, opened 2025-08-05
Accept readonly issues in flatten

IlyaSemenov:
Currently, `flatten(issues)` chokes if the list of issues is readonly. The PR fixes that.

Context: such issues come from [Standard Schema implementation in trpc](https://github.com/trpc/trpc/blob/1876ecd7a151252548f2e1ddcee945dedfca17a1/packages/server/src/vendor/standard-schema-v1/error.ts).


fabian-hiller:
Valibot v1.4.0 is available

PR #1313, opened 2025-09-17
feat: add husky and lint-staged

yslpn:
## feat: add Husky and lint-staged for pre-commit checks

- Add Husky pre-commit hook
- Configure lint-staged to run prettier, eslint and tsc on staged files.
- Dev tooling only; no runtime code changes

Instant feedback. Streamlines contributions by enforcing consistent formatting locally, reducing back-and-forth on formatting fixes

### DIFF
--- a/website/.lintstagedrc.js
+++ b/website/.lintstagedrc.js
@@ -1,5 +1,5 @@
 export default {
   '*.{ts,js,cjs,json,md}': 'prettier --write',
-  'src/**/*.css': 'prettier --write',
+  'src/**/*.{css,mdx}': 'prettier --write',
   'src/**/*.{ts,tsx}': ['prettier --write', 'eslint --fix'],
 };


packages/i18n/.lintstagedrc.js
```diff
@@ -0,0 +1,4 @@
+export default {
+  '*.{ts,js,json,md}': 'prettier --write',
+  'src/**/*.ts': ['prettier --write', () => 'tsc -p tsconfig.json --noEmit'],
```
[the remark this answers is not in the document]
yslpn: It seems like we should check the entire project to determine whether a file has the correct types. Therefore, we don't need to provide a path to a specific file, as lint-staged does. This is recommended by the creators of lint-staged.

https://github.com/lint-staged/lint-staged?tab=readme-ov-file#example-run-tsc-on-changes-to-typescript-files-but-do-not-pass-any-filename-arguments

website/.lintstagedrc.js
```diff
@@ -0,0 +1,5 @@
+export default {
+  '*.{ts,js,cjs,json,md}': 'prettier --write',
+  'src/**/*.css': 'prettier --write',
+  'src/**/*.{ts,tsx}': ['prettier --write', 'eslint --fix'],
```
fabian-hiller, line 4: I think we need to add MDX here

fabian-hiller:
I never used Husky and lint-staged before. Does this run linting and formatting on every commit? If so, I am not sure if I want that. Mostly because it probably takes more than 30 seconds to run everything. Is it smart enough to only run it on files or folders that have changed?

yslpn:
> I never used Husky and lint-staged before. Does this run linting and formatting on every commit? If so, I am not sure if I want that. Mostly because it probably takes more than 30 seconds to run everything. Is it smart enough to only run it on files or folders that have changed?

This only happens automatically for files that will be committed.

The only exception is when we can't check the type validity of a single file. In this case, we run tsc on the project (this is only configured in i18n). For this package and my computer, it's almost instantaneous. 

I really like it when open-source projects have these hooks. Because many people use different editors, different environments, or are simply inattentive or constantly switching between projects. This helps.

website/.lintstagedrc.js
yslpn: Thank you. I added mdx for prettier, eslint is not configured for mdx yet.

PR #1320, opened 2025-09-25
fix: update Guides link in Header component and add source to stringi…

yslpn:
- The Guides link is now active for any guide page.
- Added the missing GitHub link.

### DIFF
--- a/website/src/components/Header.tsx
+++ b/website/src/components/Header.tsx
@@ -139,7 +139,7 @@ export const Header = component$<HeaderProps>(({ searchOpen }) => {
           id="main-menu"
         >
           {[
-            { label: 'Guides', href: '/guides/introduction/' },
+            { label: 'Guides', href: '/guides/' },
             { label: 'API', href: '/api/' },
             { label: 'Blog', href: '/blog/' },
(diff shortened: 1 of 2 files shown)


website/src/components/Header.tsx
```diff
@@ -139,7 +139,7 @@ export const Header = component$<HeaderProps>(({ searchOpen }) => {
           id="main-menu"
         >
           {[
-            { label: 'Guides', href: '/guides/introduction/' },
+            { label: 'Guides', href: '/guides/' },
```
fabian-hiller, line 142: Why would you change this? It should automatically redirect to the introduction page.

fabian-hiller: The only reason I would probably change it if this causes issues when prefetching the route information.

yslpn: We have several guide pages. The navigation link is expected to be active for all pages, not just the first one.

fabian-hiller: Oh... my bad! Thank you!

fabian-hiller:
Thank you for this PR!

Issue #1207, opened 2025-05-28
Inconsistent type inference for optional


anatoly-tenenev:
Code to reproduce:
```typescript
import * as v from 'valibot';

const S1 = v.pipe(
  v.optional(v.number()),
  v.transform(() => 'success')
);

type T1 = v.InferOutput<typeof S1>;

// expected type error when assign undefined
const test1: T1 = undefined;

const S2 = v.object({
  s1: S1,
});

type T2 = v.InferOutput<typeof S2>;

// inconsistent ability to assign undefined
const test2: T2['s1'] = undefined;
```
https://valibot.dev/playground/?code=JYWwDg9gTgLgBAKjgQwM5wG5wGZQiOAcg2QBtgAjCGQgbgCh6BjCAO1XgGUBGOAXkwA6MMDABTABT04QiGBjA2ZCRkGsAriApioEgJR6ANNKEwoydtmggJ+-gD4iqdUyZjUqQnvp6G9GACe4nAAKrwCqgCSrNg6APLqMGCJADyB4hDYcDz2fgD0eXBiAB7iTDBiACZw6WJFUHhQzGwcNe4w3ABcoeFw6qyVYtjArFV+LOxcAEz8shQAVmLlEgDeJqhd2dzGAL6+jLWhMxGC0bFQCUmptZnZU7mMBXAjE6jAHGKs8MgUwOSBNQgKA8wAA5qw+gMhiMqs1Jm0OFNuiEpgBtQgbQgAXVm-UGw1GlQYQA

anatoly-tenenev:
This code fixes the problem
```typescript
v.optional(v.number(), () => undefined)
```
[...]

PR #1321, opened 2025-09-25
Adds example for optional

robations:
## What it does

Adds an example of setting a default of `undefined` for the optional schema, which came from discussion in #1207.

## Open qus

This way of setting undefined might potentially apply to all schemas? Should this also or instead be documented at a higher level?


website/src/routes/api/(schemas)/optional/index.mdx
```diff
@@ -80,6 +80,18 @@ const OptionalNumberSchema = v.optional(v.number());
 const NumberSchema = v.unwrap(OptionalNumberSchema);
 ```
 
+### Default to `undefined`
+
+If you wish optional entries to default to an `undefined` value, you must pass the `default_` parameter as a function returning `undefined`, otherwise this will be seen as no parameter being passed.
```
fabian-hiller: Should we change it to "If you wish missing object entries to default to an `undefined` value, ..."?

website/src/routes/api/(schemas)/optional/index.mdx
```diff
@@ -80,6 +80,18 @@ const OptionalNumberSchema = v.optional(v.number());
...
+
+```ts
+const OptionalEntrySchema = v.object({
+  key: v.optional(v.string(), () => undefined),
+});
+const res = v.parse(OptionalEntrySchema, {});
+// res = { key: undefined };
+```
```
fabian-hiller, line 112: I think I would just write this to be consistent with the other example. But I see why you added it.

```ts
const OptionalEntrySchema = v.object({
  key: v.optional(v.string(), () => undefined),
});
```

fabian-hiller (commented):
Feel free to also add it to [`nullish`](https://valibot.dev/api/nullish/).

Maybe should also add a little info to the default value section of the [optionals guide](https://valibot.dev/guides/optionals/).

PR #1345, opened 2025-11-06
fix(to-json-schema): support integer type for minValue and maxValue actions

patdx:
## Problem

The `minValue` and `maxValue` actions in `@valibot/to-json-schema` only supported `number` type, but JSON Schema supports `minimum` and `maximum` for both `number` and `integer` types. This caused an error when converting schemas that used `v.pipe(v.number(), v.integer(), v.minValue())` or `v.maxValue()`.

**Error message:**

```
The "min_value" action is not supported on type "integer".
```

## Solution

Updated both `min_value` and `max_value` action handlers in `convertAction.ts` to accept both `'number'` and `'integer'` types, matching JSON Schema's support for these constraints.

## Changes

- Updated `max_value` action handler to check for both `number` and `integer` types
- Updated `min_value` action handler to check for both `number` and `integer` types
- Added test cases to cover integer type scenarios for both actions

## Testing

All tests pass, including 4 new test cases:
- `convertAction > should convert min value action for integers`
- `convertAction > should convert max value action for integers`
- `toJsonSchema > should convert schema > for integer schema with min value`
- `toJsonSchema > should convert schema > for integer schema with max value`

Run `pnpm test run` in the `packages/to-json-schema` directory to verify.

### DIFF
--- a/packages/to-json-schema/src/converters/convertAction/convertAction.ts
+++ b/packages/to-json-schema/src/converters/convertAction/convertAction.ts
@@ -246,7 +246,7 @@ export function convertAction(
     }
 
     case 'max_value': {
-      if (jsonSchema.type !== 'number') {
+      if (jsonSchema.type !== 'number' && jsonSchema.type !== 'integer') {
         errors = addError(
           errors,
           `The "max_value" action is not supported on type "${jsonSchema.type}".`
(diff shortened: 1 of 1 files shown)


patdx:
I think this is resolved by https://github.com/open-circle/valibot/pull/1367 now. Did not notice that existing issue when I created this PR. Since it should now be resolved I will close this PR.

PR #1348, opened 2025-11-08
feat: add jwt validation action and docs

yslpn:
## Overview

- rename the proposed `jwt()` action to `jwsCompact()` and scope it to the 3-part JWS compact serialization
- leave room for a future `jweCompact()` action for 5-part compact tokens and keep `jwt()` reserved for broader JWT support
- add runtime and type tests, website docs, and `@valibot/to-json-schema` support for `jwsCompact()`
- document that this is only a lightweight structural check; full JWT validation, signature verification, and claim checks should use dedicated libraries

## Summary by CodeRabbit

* **New Features**
  * Added a JWS Compact validation action to verify JWS compact-serialized strings (three unpadded Base64URL segments) and expose it in the public API.

* **Tests**
  * Added runtime and type-level tests covering valid and invalid JWS compact inputs.

* **Documentation**
  * Added API reference pages, property metadata, examples, and updated API menu entries for the JwsCompact action and related types.

* **Chores**
  * Exported the new action for broader tooling and schema conversion support.

### DIFF
--- a/.agents/skills/repo-website-api-create/SKILL.md
+++ b/.agents/skills/repo-website-api-create/SKILL.md
@@ -117,6 +117,29 @@ export const properties: Record<string, PropertyProps> = {
 };
 ```
 
+### Optional Object Keys
+
+For optional object keys from TypeScript source such as `key?: string`, do not append `?` to the property name in `properties.ts`.
+
+Use the plain key name and represent optionality in the value type with `undefined` as the last union option:
+
+```typescript
+payload: {
+  type: {
+    type: 'object',
+    entries: [
+      {
+        key: 'key',
+        value: {
+          type: 'union',
+          options: ['string', 'undefined'],
+        },
+      },
+    ],
+  },
+},
+```
+
 ### DefinitionData Types
 
 | Type            | Syntax                                                                         |
@@ -130,7 +153,7 @@ export const properties: Record<string, PropertyProps> = {
 | Intersect       | `{ type: 'intersect', options: [type1, type2] }`                               |
 | Array           | `{ type: 'array', item: elementType }`                                         |
 | Tuple           | `{ type: 'tuple', items: [type1, type2] }`                                     |
-| Object          | `{ type: 'object', entries: [{ key: 'name', optional?: true, value: type }] }` |
+| Object          | `{ type: 'object', entries: [{ key: 'name', value: type }] }`                   |
 | Function        | `{ type: 'function', params: [{ name: 'x', type: t }], return: retType }`      |
 | Template        | `{ type: 'template', parts: [{ type: 'string', value: '>=' }, otherType] }`    |
 
--- a/.agents/skills/repo-website-api-update/SKILL.md
+++ b/.agents/skills/repo-website-api-update/SKILL.md
@@ -24,7 +24,7 @@ Update documentation when source code changes:
 
 1. **Read full source file** - Don't just look at diff; understand complete current state
 2. **Identify changes** - Categorize as addition, removal, or modification
-3. **Update `properties.ts`** - Match types exactly to source
+3. **Update `properties.ts`** - Match types exactly to source; for optional object keys, keep the plain key name and use a union with `undefined` as the last option instead of appending `?`
 4. **Update `index.mdx`** - Signature, generics, parameters, examples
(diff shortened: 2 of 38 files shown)


fabian-hiller:
Sorry for my late replay! Thank you for this PR! We really need this action! I will review this PR when working on Valibot v1.3.

fabian-hiller:
Are you already a member of the Valibot Discord?

yslpn:
> Are you already a member of the Valibot Discord?

Yes. Same username as on GitHub.

> We really need this action! I will review this PR when working on Valibot v1.3.

Agreed. I'll be here if any fixes are needed. Or if there are any questions.

fabian-hiller:
@coderabbitai please review this PR

fabian-hiller:
Thank you for this PR. I am trying to get most important PRs merged in the next weeks. One thing I’m not sure about is whether regex is the right boundary for `jwt()`. I would love to discuss this with you.

The current implementation seems reasonable if the intention is just a lightweight structural check for the common 3-part JWT/JWS shape. In that sense, regex is simple, fast, and consistent with several other format validators in the library.

That said, it might be worth aligning on what jwt() is expected to mean:
- If it is meant as a loose “looks like a JWT” check, the regex approach feels fine.
- If it is meant as a stricter JWT-format validator, regex alone may be too narrow in some places and too permissive in others.
  - But if a strict check results in a huge bundle size it is probably also not perfect for Valibot

For example, this currently only allows the 3-part compact form, so it would exclude JWTs represented as JWE compact serialization with 5 parts. On the other hand, it will also accept strings that are just three base64url-like segments even if the decoded header/payload are not valid JSON.

I don’t see this as a blocker by itself, but I think it would be good to decide explicitly which contract we want the jwt() action to have, and then shape the implementation/docs around that.

yslpn:
> Thank you for this PR. I am trying to get most important PRs merged in the next weeks. One thing I’m not sure about is whether regex is the right boundary for `jwt()`. I would love to discuss this with you.
> 
> The current implementation seems reasonable if the intention is just a lightweight structural check for the common 3-part JWT/JWS shape. In that sense, regex is simple, fast, and consistent with several other format validators in the library.
> 
> That said, it might be worth aligning on what jwt() is expected to mean:
> 
> * If it is meant as a loose “looks like a JWT” check, the regex approach feels fine.
> * If it is meant as a stricter JWT-format validator, regex alone may be too narrow in some places and too permissive in others.
>   
>   * But if a strict check results in a huge bundle size it is probably also not perfect for Valibot
> 
> For example, this currently only allows the 3-part compact form, so it would exclude JWTs represented as JWE compact serialization with 5 parts. On the other hand, it will also accept strings that are just three base64url-like segments even if the decoded header/payload are not valid JSON.
> 
> I don’t see this as a blocker by itself, but I think it would be good to decide explicitly which contract we want the jwt() action to have, and then shape the implementation/docs around that.

Thanks, this makes sense to me, and I agree the main question is the contract rather than regex itself.

To make that boundary explicit, I renamed the action from `jwt()` to `jwsCompact()`. My intention here is to keep this as a lightweight structural validator for the 3-part JWS compact serialization, which also covers the common signed JWT shape, without implying broader JWT semantics.

This also leaves room to add `jweCompact()` later for the 5-part compact serialization if we decide to support that shape, while keeping `jwt()` reserved in case we ever want a broader or higher-level API in the future.

I also updated the documentation to describe the current boundaries explicitly: `jwsCompact()` only checks the structural format, does not decode the token, does not verify signatures, and does not validate claims. For full JWT validation and signature verification, I think dedicated libraries are the better fit, so I added a pointer to https://www.jwt.io/libraries?programming_language=javascript as well.

Given that, I think the lightweight `jwsCompact()` direction is a good fit for Valibot.

fabian-hiller:
I will probably merged this now but wanted to ask if the regex can be shorten without changing it meaning or perf. Feel free to check and create a PR if so.

```js
const JWS_COMPACT_REGEX = /^(?:[\w-]{4})*(?:[\w-]{2,4})\.(?:[\w-]{4})*(?:[\w-]{2,4})?\.(?:[\w-]{4})*(?:[\w-]{2,4})?$/u
```

fabian-hiller:

PR #1351, opened 2025-11-11
feat: include `type` in enum

paoloricciuti:
So, all of this sprouted by this PR in the conformance tests for model context protocol: https://github.com/modelcontextprotocol/conformance/pull/14

The official SDK does not support Valibot, but the library I've built (https://github.com/paoloricciuti/tmcp) does, and I was getting a failing test when implementing the `everything-server` for the conformance test using Valibot as the schema.

After investigating the reason why, turns out that `@valibot/to-json-schema` doesn't include the `type` parameter for enums JSON schema, which is required to be `string` for elicitation requests. I kinda fixed in at the adapter layer in TMCP with a somewhat atrocious hack, but I think it would be interesting to actually fix this in `@valibot/to-json-schema` and that's what I did here...since we both `picklist` and `enum` can only accept string or numbers to be valid we can actually check what the options are and include the right `type` into the result JSON-Schema.

As I wrote in the comment, technically we could always return a type as an array and pushing stuff into it, but that would still fail the MCP validation which I think it would be a shame.

WDYT? Should we include this?


fabian-hiller:
Sorry for my late reaction. I forgot to add the PR to one of our milestones. I will look into it for v1.6.

fabian-hiller:
I will address this PR in https://github.com/open-circle/valibot/pull/1430 and merge it there.

PR #1359, opened 2025-11-23
Export raw check and transform types for improved DX

### LINKED ISSUE #1218: Export rawTransform Context
Hi! Thank you for this awesome library 😄 

I'd like to extract my long `rawTransform` into a separate file. Could we export this in the TS definitions?
```ts
interface Context$1<TInput> {
    readonly dataset: TypedDataset<TInput, never>;
    readonly config: Config<RawTransformIssue<TInput>>;
    readonly addIssue: AddIssue<TInput>;
    readonly NEVER: never;
}
```

Same with all the menitioned types and types used by them, like `AddIssue`, `IssueInfo`, etc.

Without it, I can't get proper typing on my extracted function without a lot of repetition. I guess using TS's `Parameters` to "steal" the type also can't work when the function is generic.

fabian-hiller:
Fix #1218


fabian-hiller:

PR #1383, opened 2025-12-30
feat: ISO4217 Currency Code

ysknsid25:
Being able to provide types that comply with ISO4217 makes sense in the following cases.

- When setting sales currency units on an e-commerce site.
- As a guardrail when obtaining exchange rate information on an MCP server.
- etc

I think this feature will save users the pain of defining ISO4217 compliant code for every application.


library/src/actions/isoCurrencyCode/isoCurrencyCode.ts
```diff
@@ -0,0 +1,289 @@
...
+  'XPD',
+  'XPF',
+  'XPT',
+  'XTS',
+  'XXX',
+  'YER',
+  'ZAR',
+  'ZMW',
+  'ZWG',
+] as const;
```
yslpn, line 194: Who created this list? It includes both historical and currently active ones. But not all are active and not all are historical.

For example, there are no active
CHE — WIR Euro (Switzerland)
CHW — WIR Franc (Switzerland)
There are historical HRK — Croatian Kuna
And so on.

> "This action supports the official list of active ISO 4217 currency codes."

The documentation should clearly describe what it checks. The current description is incorrect and too general.

fabian-hiller:
I think this is a great idea, and since our architecture is modular, we can implement it without affecting our users' bundles if it is not used. However, I am still concerned that shipping more actions like this will increase the size of our general bundle in cases where tree-shaking cannot be applied. Perhaps it would be better for users to bring their own currency list and use our `enum` or `picklist` schema instead.

More context here:
- https://valibot.dev/guides/enums/
- https://valibot.dev/api/enum/
- https://valibot.dev/api/picklist/

What do you think?

ysknsid25:
@fabian-hiller 
Thank your for your opinion. I thought it would be a good idea to publish the list as a separate library so that only those who want to use it can use it.

This pull request will be closed 🙏

PR #1384, opened 2025-12-30
feat: ISO3166 Country Code

ysknsid25:
Being able to provide types that comply with ISO3166 makes sense in the following cases.

- When setting language.
- Managing nationality information
- Obtaining some international geographic information
  - For example, weather, disasters, conflict areas, etc.
- As a guardrail when nationality information information on an MCP server.
- etc

I think this feature will save users the pain of defining ISO3166 compliant code for every application.


library/src/actions/isoCountryCode/isoCountryCode.ts
```diff
@@ -0,0 +1,401 @@
...
+}
+
+/**
+ * Creates an [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) country code action.
+ *
+ * @returns An ISO 3166-1 country code action.
+ *
+ * @beta
+ */
+export function isoCountryCode<TInput extends string>(): IsoCountryCodeAction<
```
yslpn, line 1076: What if someone needs validation for ISO_3166-2 or ISO_3166-3?

library/src/actions/isoCountryCode/isoCountryCode.ts
```diff
@@ -0,0 +1,401 @@
...
+  { name: 'Vanuatu', alphaTwo: 'VU', alphaThree: 'VUT' },
+  { name: 'Venezuela, Bolivarian Republic of', alphaTwo: 'VE', alphaThree: 'VEN' },
+  { name: 'Viet Nam', alphaTwo: 'VN', alphaThree: 'VNM' },
+  { name: 'Virgin Islands (British)', alphaTwo: 'VG', alphaThree: 'VGB' },
+  { name: 'Virgin Islands (U.S.)', alphaTwo: 'VI', alphaThree: 'VIR' },
+  { name: 'Wallis and Futuna', alphaTwo: 'WF', alphaThree: 'WLF' },
+  { name: 'Western Sahara', alphaTwo: 'EH', alphaThree: 'ESH' },
+  { name: 'Yemen', alphaTwo: 'YE', alphaThree: 'YEM' },
+  { name: 'Zambia', alphaTwo: 'ZM', alphaThree: 'ZMB' },
+  { name: 'Zimbabwe', alphaTwo: 'ZW', alphaThree: 'ZWE' },
```
yslpn: Storing the name is unnecessary and bloats the bundle.

library/src/actions/isoCountryCode/isoCountryCode.ts
```diff
@@ -0,0 +1,401 @@
...
+] as const;
```
yslpn, line 1011: Please provide the source of your data so we can ensure it is free of errors.

library/src/actions/isoCountryCode/isoCountryCode.ts
```diff
@@ -0,0 +1,401 @@
...
+> {
+  return {
+    kind: 'validation',
+    type: 'iso_country_code',
+    reference: isoCountryCode,
+    async: false,
+    expects: null,
+    requirement(input) {
+      return ISO_3166_1_ENTRIES.some(
+        (entry) => entry.alphaTwo === input || entry.alphaThree === input
```
yslpn, line 1110: My opinion: each check should be a separate action, since some people only need one thing, while others need everything. If someone doesn't need the three-letter check, they'll get an extra code in their application.

ISO 3166-1-alpha-2
ISO 3166-1-alpha-3
ISO 3166-1-numeric

Plus, in the current version, it's unclear why we don't check the numeric value, even though it's part of the standard.

fabian-hiller:
I think this is a great idea, and since our architecture is modular, we can implement it without affecting our users' bundles if it is not used. However, I am still concerned that shipping more actions like this will increase the size of our general bundle in cases where tree-shaking cannot be applied. Perhaps it would be better for users to bring their own country list and use our `enum` or `picklist` schema instead.

More context here:
- https://valibot.dev/guides/enums/
- https://valibot.dev/api/enum/
- https://valibot.dev/api/picklist/

What do you think?

ysknsid25:
@fabian-hiller 
Thank your for your opinion. I thought it would be a good idea to publish the list as a separate library so that only those who want to use it can use it.

This pull request will be closed 🙏

PR #1398, opened 2026-01-25
Add maxDate and minDate action implementations with corresponding tests

sandros94:
As I was working on a side-project for Valibot I realized that we were missing the `minDate` and `maxDate` action, which are quire useful for events that require time-based validation with full type support.

Skipped opening issue as implementation is quite basic and straight forward.

Feel free to close if these are not intended and developers should use `maxValue`/mivValue` instead.


sandros94:
Ok, actually it was my local ts server that was acting a bit strangely. `minValue` and `maxValue` do return correctly Date types.

I should have tested this in the playground first 😬

PR #1421, opened 2026-03-11
fix: prefer hasOwnProperty over hasOwn

alaycock-stripe:
We are using Valibot for a project that requires significant browser backwards compatibility but I am unable to use some schemas because of `Object.hasOwn`. While this is possible to polyfill, this small change would allow us to drop our polyfill, reducing our bundle size. The tradeoff is a small reduction in readability for significantly better browser compatibility.

This change is similar to an [existing polyfill](https://github.com/niksy/babel-plugin-transform-object-hasown/blob/master/index.js#L4-L10), except it relies on types rather than a throwing an exception.

The change should have no runtime impact, all tests continue to pass.


fabian-hiller:
How much of an impact it the bundle size of the polyfill? Another option for you is to copy our `object` schema code (Valibot is modular so you can bring your own schemas and actions) or patch this locally (your package manager probably supports patching a library).

alaycock-stripe:
When using core-js to polyfill several bundles are getting a ~3kb bump. I haven't explored into why the impact is so significant.

I'll look into patching, but was curious if this change would be appreciated upstream. Your tsconfig targets ES2020 but hasOwn landed in ES2022.

fabian-hiller:
> I'll look into patching, but was curious if this change would be appreciated upstream. Your tsconfig targets ES2020 but hasOwn landed in ES2022.

This is a good argument. On the other side hasOwn should have around 97% coverage across browsers and no one has complained so far. I was using hasOwn instead of hasOwnProperty because the MDN docs recommended it.

fabian-hiller:
I think I prefer to wait before making a decision.

alaycock-stripe:
That's fair. We've patched it for our own purposes, so if the change isn't helpful upstream feel free to close. Our use-case is not everybody's case.

krokyze:
This is the only reason that forces me to add a polyfill/patch in my setup (noticed it after connecting Sentry). Would be happy if this lands into main.

fabian-hiller:
Valibot v1.4.0 is available

Issue #1375, opened 2025-12-20
`v.multipleOf` does not handle some non-integer inputs well


esdmr:
It seems that `v.multipleOf` tries to find the remainder of the input divided by the requirement. This might not yield an appropriate result if their division has floating point precision errors:

```js
const Length = v.pipe(v.number(), v.multipleOf(0.01), v.description('in meters, with centimeter precision'));

toJsonSchema(Length) // {"type": "number", "multipleOf": 0.01, "description": ...}

v.parse(Length, 3); // Invalid multiple: Expected %0.01 but received 3
```

fabian-hiller:
Hey, thank you for reaching out! What would you recommend us to change?

esdmr:
Different libraries seem to be handling this case differently

<details><summary>Test case for different libraries</summary>

```json
{
  "private": true,
  "type": "module",
  "dependencies": {
    "ajv": "^8.17.1",
    "arktype": "^2.1.29",
    "joi": "^18.0.2",
    "typebox": "^1.0.69",
    "valibot": "^1.2.0",
    "zod": "^4.2.1"
  },
  "devDependencies": {
    "@types/node": "22"
  }
}
```

```js
import assert from "node:assert/strict";
import { describe, it } from "node:test";
import * as ajv from "ajv";
import * as arktype from "arktype";
import * as joi from "joi";
import * as typeboxType from "typebox";
import * as typeboxCompile from "typebox/compile";
import * as typeboxValue from "typebox/value";
import * as valibot from "valibot";
import * as zod from "zod";
import * as zodV3 from "zod/v3";
import * as zodMini from "zod/mini";

describe("valibot", () => {
  it("handles multipleOf(0.1) correctly", () => {
    const schema = valibot.pipe(valibot.number(), valibot.multipleOf(0.1));
    assert.ok(valibot.safeParse(schema, 3).typed);
    assert.ok(valibot.safeParse(schema, 3.3).typed);
    assert.ok(!valibot.safeParse(schema, 3.03).typed);
  });
});

describe("zod", () => {
  describe('v4', () => {
    it("handles multipleOf(0.1) correctly", () => {
      const schema = zod.number().multipleOf(0.1);
      assert.ok(zod.safeParse(schema, 3).success);
      assert.ok(zod.safeParse(schema, 3.3).success);
      assert.ok(!zod.safeParse(schema, 3.03).success);
    });
  });

  describe("v3", () => {
[...]

PR #1422, opened 2026-03-12
fix(multipleOf): use float-safe modulo to prevent false positives with decimal requirements

Vadaski:
## Problem

`v.multipleOf()` incorrectly rejects valid inputs when the requirement is a decimal number, because JavaScript's `%` operator introduces floating-point precision errors.

```ts
const Length = v.pipe(v.number(), v.multipleOf(0.01));
v.parse(Length, 3); // throws — but 3 is a valid multiple of 0.01
v.parse(Length, 1.5); // throws — but 1.5 is a valid multiple of 0.01
```

The root cause: `3 % 0.01` in JavaScript is `~2.22e-16` instead of `0`.

Reported in https://github.com/fabian-hiller/valibot/issues/1375.

## Fix

Scale both operands to integers before taking the modulo:

1. Compute the maximum decimal places of value and requirement via `_getDecimalPlaces()`
2. Multiply both by `10^decimalPlaces` using `Math.round()` to neutralize any remaining float error at the scale boundary
3. Compare the integer remainder to `0`

The `bigint` path is unchanged in behavior; the explicit `typeof` narrowing now allows `!== 0n` without the previous `@ts-expect-error`.

## Tests added

```ts
// These now pass correctly:
// multipleOf(0.01) on [3, 0.1, 0.03, 1.5]
// multipleOf(0.1)  on [0.3]
// multipleOf(0.25) on [0.75]

// These still correctly fail:
// multipleOf(0.01) on [-Infinity, Infinity, NaN]
```

## Known limitation

For pathological inputs with very high decimal precision combined with very large values (e.g. `multipleOf(1e-17)` on `Number.MAX_SAFE_INTEGER`), the scaled integer can exceed `Number.MAX_SAFE_INTEGER` and lose precision. This matches the inherent limitation of IEEE 754 arithmetic. Practical uses of `multipleOf` with reasonable decimal requirements (like `0.01`, `0.1`, `0.25`) are not affected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multiple-of validation to correctly handle decimal precision (including scientific notation), bigint zero behavior, and special numeric values to avoid incorrect pass/fail results.

* **Tests**
  * Expanded test coverage for decimal and bigint scenarios, covering fractional values, scientific-notation inputs, Infinity/NaN cases, and zero-requirement edge cases.


fabian-hiller:
Thank you for creating a PR. Here some feedback:

1. Performance — recomputing on every call: _getDecimalPlaces(this.requirement) runs inside ~run, meaning it's recalculated on every validation even though the requirement is constant. Should be precomputed once in the factory and closed over.                                                                           
2. Helper function placement: Defining const _getDecimalPlaces at module scope is unusual for this codebase. Actions with helpers use a utils/ subfolder (like isbn/utils/) and the `function` keyword.
3. Missing test edge cases: No tests for values that should fail decimal checks (e.g., multipleOf(0.01) with 0.001), division by zero (multipleOf(0)), or integer requirement with decimal value (this is AI feedback, not sure if it is true and relevant).

fabian-hiller:
**Current `_getDecimalPlaces` allocates heavily:** `toString()` + 2x `split()` creates ~5 heap objects (1 string, 2 arrays, sub-strings) per call. In a hot validation path (e.g., validating array items), this adds GC pressure.

**Alternative A — Optimized string (drop-in replacement):**
Use `indexOf` + `slice` instead of `split` to avoid array allocations (~2 allocations instead of ~5):

```ts
const _getDecimalPlaces = (value: number): number => {
  const s = String(Math.abs(value));
  const dotIndex = s.indexOf('.');
  const eIndex = s.indexOf('e');
  if (dotIndex === -1 && eIndex === -1) return 0;
  const decimals = dotIndex === -1 ? 0 : (eIndex === -1 ? s.length : eIndex) - dotIndex - 1;
  if (eIndex === -1) return decimals;
  return Math.max(0, decimals - Number(s.slice(eIndex + 1)));
};
```

**Alternative B — Epsilon-based (eliminates `_getDecimalPlaces` entirely):**
Zero allocations, pure arithmetic. Replace the entire scaling approach with a remainder + epsilon check:

```ts
const remainder = Math.abs(value % requirement);
const tolerance = Number.EPSILON * Math.max(Math.abs(value), Math.abs(requirement));
if (remainder > tolerance && Math.abs(remainder - Math.abs(requirement)) > tolerance) {
  _addIssue(this, 'multiple', dataset, config);
}
```

The double check is needed because `%` can produce a result near `0` OR near `requirement` due to float rounding. For the motivating case: `3 % 0.01 ≈ 2.22e-16`, and `Number.EPSILON * 3 ≈ 6.66e-16`, so `2.22e-16 < tolerance` → correctly passes.

**Trade-off summary:**

| | Current (PR) | Alt A (indexOf) | Alt B (epsilon) |
|---|---|---|---|
| Allocations/call | ~5 | ~2 | 0 |
| Correctness | Exact (within MAX_SAFE_INTEGER) | Same | Approximate |
| Complexity | Medium | Medium | Low |

This was created by Claude Code 🤖. Let me know if it is wrong or complicates the code.

Vadaski:
Good catches, thank you.

All three points addressed in the latest commit:

- **Helper placement**: `_getDecimalPlaces` is now in `utils/_getDecimalPlaces.ts` using the `function` keyword, mirroring the `isbn/utils/` pattern. Also switched to `indexOf`/`slice` instead of `split` to reduce allocations.
- **Precomputed in factory**: `_getDecimalPlaces(requirement)` is now computed once before `return {}` and closed over — `~run` only calls it for `dataset.value`.
- **Missing tests**: Added three test cases — invalid decimal multiples (`0.001`, `0.015`, `1.234` against `multipleOf(0.01)`), decimal values with an integer requirement (`1.5`, `0.5` against `multipleOf(2)`), and zero requirement (always invalid).

Vadaski:
Latest push addresses two edge cases I missed:

- **bigint `0n` requirement**: `value % 0n` throws a `RangeError`. Now guarded — `multipleOf(0n)` treats all values as invalid, matching the number `0` behavior.
- **Numeric zero with extreme requirements**: `0` is always a valid multiple of any finite nonzero number, so it now short-circuits before the decimal scaling path. The guard explicitly checks `Number.isFinite(requirement)` to avoid silently passing `0` for `NaN`/`Infinity` requirements.

Also added scientific-notation test cases (`1e-7`) to cover the exponent parsing branch in `_getDecimalPlaces`.

HardikDewra (commented):
This is a well-thought-out fix for a real floating point issue. `0.3 % 0.1` famously returns `0.09999999999999998` in JS instead of `0`, and the integer-scaling approach via `_getDecimalPlaces` handles it correctly.

A few observations:

1. The `_getDecimalPlaces` utility handles scientific notation well (e.g. `1e-7`), and the coefficient/exponent decomposition is correct. Good that it uses `Math.abs` to handle negative numbers.

2. The early return for `dataset.value === 0` with a non-zero finite requirement is a nice optimization - zero is always a multiple of anything except zero.

3. The explicit `typeof` branching for `number` vs `bigint` is cleaner than the old single-line with `@ts-expect-error`. Also good that bigint `0n` requirement now correctly issues an error for all inputs (division by zero is undefined).

4. The fallback to the raw `%` operator when `multiplier` overflows to Infinity is sensible - avoids `NaN` results from multiplying by Infinity.

5. Test coverage is thorough - decimals, scientific notation, zero requirements, Infinity/NaN inputs.

fabian-hiller:
Hey, thanks for the iterations — the bigint `0n` guard and added test coverage are great catches.

Before we land it, I'd like your take on the core algorithm. Two things I'm wondering about:

1. `multipleOf(0.1)` on `0.1 + 0.2` (i.e. `0.30000000000000004`) still reports invalid. Since users typically arrive at these values through arithmetic rather than literals, I'm not sure if this fully resolves the bug shape from [#1375](https://github.com/fabian-hiller/valibot/issues/1375).
2. The `MAX_SAFE_INTEGER` limitation triggers whenever the scaled integer overflows — worth thinking about how often that comes up in practice.

For reference, Zod v4 went with a relative-epsilon check ([source](https://github.com/colinhacks/zod/blob/main/packages/zod/src/v4/core/util.ts#L248)):

```ts
const ratio = val / step;
const rounded = Math.round(ratio);
const tolerance = Number.EPSILON * Math.max(Math.abs(ratio), 1);
// valid if Math.abs(ratio - rounded) < tolerance
```

That's one option, but there may be reasons to prefer the scaled approach I'm not seeing. What's your read — stick with the current direction, or explore alternatives? Happy to help either way.

PR #1429, opened 2026-03-17
Add `isPlainObject` util, and use it for `v.record` - also add `v.pojo()` schema

### LINKED ISSUE #1387: `v.record()` accepts arrays at runtime but TypeScript types reject them
## Sample Code

```typescript
import * as v from 'valibot';

const Schema = v.record(v.string(), v.unknown())
type SchemaType = v.InferOutput<typeof Schema>
// ^{ [x: string]: unknown; }

const data = [2, 7, 5, 4] satisfies SchemaType
// ^ Type 'number[]' does not satisfy the expected type '{ [x: string]: unknown; }'.
//   Index signature for type 'string' is missing in type 'number[]'. (1360)

console.log(v.safeParse(Schema, data));
// ^ success
//   Why?
```

https://valibot.dev/playground/

## Expected Behavior

Either:

1. Runtime validation should reject arrays (consistent with TypeScript types), or
2. TypeScript types should accept arrays (consistent with runtime behavior)

## Possible Solutions

1. Add an array check to `v.record()` validation
2. Update TypeScript types to include arrays
3. Document this behavior clearly

Is this intentional behavior, or should it be fixed?

EskiMojo14:
Based on `@httpx/plain-object`
```ts
export function isPlainObject(
  value: unknown
): value is Record<string, unknown> {
  if (value === null || typeof value !== 'object') {
    return false;
  }

  const proto = Object.getPrototypeOf(value);
  return (
    proto === null ||
    proto === Object.prototype ||
    // Required to support node:vm.runInNewContext({})
    Object.getPrototypeOf(proto) === null
  );
}
```

This util correctly identifies POJOs (including null prototypes and POJOs from other realms) vs complex objects like arrays.

fixes #1387 (technically breaking, though arguably a bug fix to align with existing types)

## Summary by CodeRabbit

* **New Features**
  * Added `pojo` schema validation for plain JavaScript objects, including custom error messages.
  * Exposed the new schema and related types through the public API.
  * Added a utility for identifying plain objects, including null-prototype objects.

* **Bug Fixes**
  * Improved `record` and asynchronous record validation to reject arrays and other non-plain objects correctly.

* **Documentation**
  * Added comprehensive `pojo` schema and type reference documentation, examples, and API listings.

* **Tests**
  * Added coverage for `pojo`, plain-object detection, and record validation of arrays.


EskiMojo14:
undecided if we should use `object` or `Record<string, unknown>` for the narrowed type here, open to feedback

library/src/schemas/record/record.ts
```diff
@@ -122,7 +123,7 @@ export function record(
       const input = dataset.value;
 
       // If root type is valid, check nested types
-      if (input && typeof input === 'object') {
+      if (input && isPlainObject(input)) {
```
EskiMojo14, line 126: this could just be `if (isPlainObject(input))` but i think a falsy check will probably fail faster

fabian-hiller:
Do you think there will be a meaningful performance impact?

EskiMojo14:
depends on how slow `Object.getPrototypeOf` is, I guess - primitives will still fail fast, plain objects should be pretty fast since the loop will only happen once, but non-plain objects may be slower (but they should be failing now instead of passing)

EskiMojo14:
it would probably be quicker without the loop (`Object.getPrototypeOf(value) === Object.prototype`), but then you lose correctness for cross-realm objects (VM, iframe, etc)

belgattitude:
BTW you might have a look at this implem: 

https://github.com/belgattitude/httpx/tree/main/packages/plain-object#readme.

100% redux compat, might be 2x-4x times faster depending (see benchmarks). 

````
 RUN  v4.1.4 /home/sebastien/github/httpx/packages/plain-object

 ✓ bench/comparative.bench.ts > Compare calling isPlainObject with 110x mixed types values 6702ms
     name                                                           hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · "@httpx/plain-object": `isPlainObject(v)`            1,301,180.23  0.0007  1.2707  0.0008  0.0007  0.0013  0.0016  0.0090  ±0.71%   650591
   · "is-plain-obj":"4.1.0": 'isPlainObj(v)'              1,170,620.68  0.0007  0.4172  0.0009  0.0008  0.0016  0.0019  0.0119  ±0.52%   585311
   · "@sindresorhus/is":"8.0.0": 'is.plainObject(v)'      1,135,634.79  0.0007  0.7565  0.0009  0.0008  0.0016  0.0019  0.0123  ±0.75%   567818
   · "es-toolkit":"1.45.1": 'isPlainObject(v)'            1,040,985.16  0.0008  0.7910  0.0010  0.0008  0.0018  0.0021  0.0149  ±0.58%   520493
   · "redux":"5.0.1": 'isPlainObject(v)'                    367,333.57  0.0021  8.2123  0.0027  0.0022  0.0052  0.0061  0.0333  ±4.96%   183667
   · "is-plain-object":"5.0.0": 'isPlainObject(v)'          553,579.82  0.0014  7.1030  0.0018  0.0015  0.0040  0.0043  0.0254  ±2.91%   276791
   · "immer/is-plain-object":"4.2.0": 'isPlainObject(v)'    429,868.70  0.0020  1.3350  0.0023  0.0020  0.0050  0.0056  0.0280  ±0.95%   214935
   · lodash-es:"4.18.1": '_.isPlainObject(v)'                16,017.70  0.0475  6.8642  0.0624  0.0663  0.1644  0.2127  0.4020  ±3.10%     8011

 BENCH  Summary
                                                                                                                                                                                    
  "@httpx/plain-object": `isPlainObject(v)` - bench/comparative.bench.ts > Compare calling isPlainObject with 110x mixed types values
    1.11x faster than "is-plain-obj":"4.1.0": 'isPlainObj(v)'
    1.15x faster than "@sindresorhus/is":"8.0.0": 'is.plainObject(v)'
    1.25x faster than "es-toolkit":"1.45.1": 'isPlainObject(v)'
    2.35x faster than "is-plain-object":"5.0.0": 'isPlainObject(v)'
    3.03x faster than "immer/is-plain-object":"4.2.0": 'isPlainObject(v)'
    3.54x faster than "redux":"5.0.1": 'isPlainObject(v)'
    81.23x faster than lodash-es:"4.18.1": '_.isPlainObject(v)'
````

Tested on bun, node, chrome, workers.

```typescript
import type { BasePlainObject, DefaultBasePlainObject } from './internal.types';
import type { PlainObject } from './plain-object.types';

/**
 * Check if a value is a plain object
 *
 * A plain object is a basic JavaScript object, such as {}, { data: [] }, new Object() or Object.create(null).
 *
 * @example
 * ```typescript
 * import { isPlainObject } from '@httpx/plain-object';
 *
 * // ✅👇 True
 *
 * isPlainObject({ });                       // ✅
 * isPlainObject({ key: 'value' });          // ✅
 * isPlainObject({ key: new Date() });       // ✅
 * isPlainObject(new Object());              // ✅
 * isPlainObject(Object.create(null));       // ✅
 * isPlainObject({ nested: { key: true} });  // ✅
 * isPlainObject(new Proxy({}, {}));         // ✅
 * isPlainObject({ [Symbol('tag')]: 'A' });  // ✅
 *
 * // ✅👇 (node context, workers, ...)
 * const runInNewContext = await import('node:vm').then(
 *     (mod) => mod.runInNewContext
 * );
 * isPlainObject(runInNewContext('({})'));   // ✅
 *
 * // ❌👇 False
 *
 * class Test { };
 * isPlainObject(new Test())           // ❌
 * isPlainObject(10);                  // ❌
 * isPlainObject(null);                // ❌
 * isPlainObject('hello');             // ❌
 * isPlainObject([]);                  // ❌
 * isPlainObject(new Date());          // ❌
 * isPlainObject(new Uint8Array([1])); // ❌
 * isPlainObject(Buffer.from('ABC'));  // ❌
 * isPlainObject(Promise.resolve({})); // ❌
 * isPlainObject(Object.create({}));   // ❌
 * isPlainObject(new (class Cls {}));  // ❌
 *
 * // ⚠️ Edge cases
 * //
 * // 👇 globalThis isn't properly portable across all JS environments
 * //
 *
 * isPlainObject(globalThis);          // ✅ with Bun ❌ otherwise (browser, Nodejs, edge, cloudflare)
 *
 * // 👇 Static built-in classes aren't properly checked. This is a trade-off
 * //    to maintain the best performance and size. If you need to check for these,
 * //    use a custom type guard. But in most cases, you won't need to check for these
 * //    as the probability of writing a code that receives these as plain objects is low.
 * //    and probably indicates an issue in your code.
 *
 * isPlainObject(Math);                // ⚠️✅ return true, but Math is not a plain object
 * isPlainObject(JSON);                // ⚠️✅ return true, but JSON is not a plain object
 * isPlainObject(Atomics);             // ⚠️✅ return true, but Atomics is not a plain object
 * isPlainObject(Reflect);             // ⚠️✅ return true, but Reflect is not a plain object
 * ```
 */
export const isPlainObject = <
  TValue extends BasePlainObject = DefaultBasePlainObject,
>(
  v: unknown
): v is TValue extends DefaultBasePlainObject
  ? BasePlainObject
  : PlainObject<TValue> => {
  if (v === null || typeof v !== 'object') {
    return false;
  }

  const proto = Object.getPrototypeOf(v) as typeof Object.prototype | null;
  return (
    proto === null ||
    proto === Object.prototype ||
    // Required to support node:vm.runInNewContext({})
    Object.getPrototypeOf(proto) === null
  );
};
```
Test file:

https://github.com/belgattitude/httpx/blob/main/packages/plain-object/src/__tests__/is-plain-object.test.ts

fabian-hiller:
If we implement this, does parsing env variables still work? I remember that env variables (I think it was Node.js) were not a plain object. We should check that before we continue with this PR.

EskiMojo14:
image

looks like you're right, process.env is not a plain object.

fabian-hiller:
Mhm... I was checking for plain objects using the `Object` constructor if I remember correctly. This broke env variables and some special run times like Vercel Edge Functions.

EskiMojo14:
i guess the question is whether we think anyone is using v.record for env variables etc.

if so, do we make it an option? maybe off by default until v2?

belgattitude:
I don’t think so.

Vercel edge runtime is deprecated for info. Vercel now recommend node
runtime.

By the way if you’re looking for more speed. We can create a conditional
export for browsers without the check for vm.runInNewContext

Much faster. If interested I can provide a bench later

EskiMojo14:
doesn't the same go for objects from other iframes?

belgattitude:
@fabian-hiller 

> Mhm... I was checking for plain objects using the Object constructor if I remember correctly. 
> This broke env variables and some special run times like Vercel Edge Functions.

Just tested `process.env` on few runtimes (except browsers) on the [@httpx/plain-object](https://github.com/belgattitude/httpx/tree/main/packages/plain-object#readme) implementation, the current answer:

```
In bun and cloudflare workers, `isPlainObject(process.env)` returns `true`. 
In node, edge runtime and deno `isPlainObject(process.env)` returns `false`.
```

Sources: [Test file](https://github.com/belgattitude/httpx/blob/573a807468d171096cd61ef0e140917731f51a95/packages/plain-object/src/__tests__/is-plain-object.test.ts#L140) | [PR](https://github.com/belgattitude/httpx/pull/2716) | [Release](https://github.com/belgattitude/httpx/releases/tag/%40httpx%2Fplain-object%402.1.10)

That said IMHO I think it falls in edge case section. It might be possible to create an implementation that uniformize this. Let me know

belgattitude:
@EskiMojo14 

> doesn't the same go for objects from other iframes?

I must add a test with playwright to confirm. Do you want me to try ? I hope nobody uses parent.global anymore but who knows... do you think a test with postMessage would be sufficient ?

For your information with conditional exports (browser without runInNewContext test) The gain is this one:

node 24.15 (with runInNewContext support)

```
"@httpx/plain-object": `isPlainObject(v)`            1,192,608.94  0.0007  0.5203  0.0008  0.0007  0.0014  0.0017  0.0144  ±0.72%   596305
```

conditional export for browsers, simulated on node 24.15 (without runInNewContext support) - 

```
"@httpx/plain-object": `isPlainObject(v)`            1,622,412.64  0.0005  2.5987  0.0006  0.0005  0.0010  0.0012  0.0021  ±1.44%   811207
```

So it's a bit faster, but not that much

belgattitude:
For the little story, I've been curious and tested the zod v4 [isPlainObject implem](https://github.com/colinhacks/zod/blob/b6071fc0ad2be94850ed57e07e418a6f902a9771/packages/zod/src/v4/core/util.ts#L385)

```
 RUN  v4.1.5 /home/sebastien/github/httpx/packages/plain-object
 ✓ bench/comparative.bench.ts > Compare calling isPlainObject with 110x mixed types values 7496ms
     name                                                           hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · "@httpx/plain-object": `isPlainObject(v)`            1,600,433.82  0.0005  0.6249  0.0006  0.0006  0.0011  0.0011  0.0035  ±0.51%   800238
   · "zod":"4.4.3": 'zod.util.isPlainObject(v)'             932,899.41  0.0009  1.9657  0.0011  0.0009  0.0023  0.0026  0.0166  ±1.07%   466451
```

The implementations are almost compatible. On ly 3 tests fails. I'll investigate the reasons.

Basically zod seems to 

- Object.create({}) is considered as true (I don't think it's a good thing)
- new ObjectConstructor() is considered as true (same comment)

And

````typescript
new Proxy(new Date(), {
        get(target, _prop, _receiver) {
          return target;
        },
}),
```` 

is considered true as well (imho oit should be false)

PR here for reference: https://github.com/belgattitude/httpx/pull/2718

PR #1442, opened 2026-03-30
Research how to improve TypeScript type performance

### LINKED ISSUE #1374: slow Code Completion with use `valibot`  schema define
I've raised this issue once in my original TypeScript project.
https://github.com/microsoft/TypeScript/issues/62484

I originally thought that after using `typescript-go`, this issue would become better, but the time is still very slow (or slightly faster, but it's not noticeable with long-term use).
Simply put, when using Valibot to create a large number of definitions, the autocomplete performance becomes very slow after parsing. Here's an example:
https://gist.github.com/wszgrcy/a4b9161695ab5f233fde53cd2994b82a

Or you can also use the code generation script to generate definitions of any specification.

```ts
(() => {
  function createRandomObject(list: number[]) {
    function createObject(depth: number, maxDepth: number) {
      if (depth >= maxDepth) {
        return null;
      }

      const result = {} as any;
      const keyCount = list[depth] || 1;
      for (let i = 0; i < keyCount; i++) {
        let type = 'object';
        const nestedObject = createObject(depth + 1, maxDepth);
        if (nestedObject !== null) {
          let data = Object.keys(nestedObject)
            .map((item) => {
              return `${item}:${nestedObject[item]}`;
            })
            .join(',\n');
          result[`${type}_${i}`] = `v.object({${data}})`;
        } else {
          result[`string_${i}`] = 'v.string()';
        }

        let key = `${type}_${i}`;
        let count = metadataCount;
        while (count) {
          result[key] = `v.pipe(${result[key]},v.title('test'))`;
          count--;
        }
      }

      return result;
    }

    return createObject(0, list.length);
  }
  let metadataCount = 1;
  let list = [20, 5, 4, 3];
  const randomObject = createRandomObject(list);
  let data = Object.keys(randomObject)
    .map((item) => {
      return `${item}:${randomObject[item]}`;
    })
    .join(',');
  let output = `import * as v from 'valibot';
  let Define=v.object({${data}});
  let value=v.parse(Define,undefined)`;
  console.log(output);
})();

```
For example, when I input `value.object_0.object_0.object_0.object_0`, every time there's a `.` followed by a delay before the autocomplete pops up.
And additionally, when I input the second `value.object_0.object_0.object_0.object_0`, it still requires a long waiting time.

I can accept that it's slow to parse initially (after all, the definition is quite complex). But what I can't understand is why subsequent completion is still slow every time I reference it, even though I haven't modified the definition. Doesn't it cache the results?

I don't know what else needs to be provided. If there's anything else needed, please feel free to reply, and I'll do my best to help troubleshoot the issue.

fabian-hiller:
Closes #1374


fabian-hiller:
Valibot v1.4.0 is available

PR #1476, opened 2026-05-24
fix(stringifyJson): preserve dataset.value when JSON.stringify returns undefined

francisjohnjohnston-web:
## Bug

`stringifyJson` unconditionally overwrites `dataset.value = output` even after recording a validation failure, corrupting the dataset with `undefined`.

### Root cause

```ts
// current code:
if (output === undefined) {
  _addIssue(this, 'JSON', dataset, config);
  dataset.typed = false;
}
dataset.value = output;  // ← always runs — sets value to undefined on error
```

`JSON.stringify` returns `undefined` for top-level functions, symbols, and `undefined` values. When this happens:
1. `_addIssue` is correctly called to record the error
2. `dataset.typed = false` is correctly set
3. But `dataset.value = output` then **overwrites the original input with `undefined`**

This is inconsistent with every other error path in Valibot — `parseJson`'s `catch` branch, for example, never touches `dataset.value` on failure. It's also inconsistent with the `catch` branch within `stringifyJson` itself, which doesn't overwrite `dataset.value`.

The consequence: a caller that inspects `dataset.value` after a failed `stringifyJson` sees `undefined` instead of the original invalid input (e.g. a `Symbol`). Error handlers that rely on the original value for logging or downstream processing receive a corrupted value.

## Fix

Guard the assignment so it only runs on success:

```ts
if (output === undefined) {
  _addIssue(this, 'JSON', dataset, config);
  dataset.typed = false;
} else {
  dataset.value = output;  // ← only on success
}
```

## Tests

- Updated the existing `'for function input'` test: was asserting `value: undefined` (the bug), now correctly asserts `value: input` (the original function reference)
- Added `'for symbol input'` regression test that would have caught this bug on introduction

Full suite: **3055 tests across 272 files** pass.

---
## Summary by cubic
Fixes `stringifyJson` to preserve `dataset.value` when `JSON.stringify` returns `undefined` for top-level function, symbol, or undefined inputs. Prevents clobbering the original input and keeps behavior aligned with `parseJson` and other error paths.

- **Bug Fixes**
  - Set `dataset.value = output` only when `output !== undefined`.
  - On failure, add an issue and set `dataset.typed = false` without changing `dataset.value`.

- **Tests**
  - Updated function-input test to expect the original value.
  - Added symbol-input regression test.

<sup>Written for commit bed78486fbd84b0e1a4ea191d5407046972c5628. Summary will update on new commits. <a href="https://cubic.dev/pr/open-circle/valibot/pull/1476?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed JSON stringification to preserve the original value when an input cannot be serialized.
  * Added clearer issue reporting for unsupported symbol inputs.
  * Corrected handling of function inputs so their original value is retained in results.

* **Documentation**
  * Updated the changelog with details of the stringification fix.


yslpn (dismissed):
LGTM

PR #1477, opened 2026-05-24
fix(value/notValue/values/notValues): use Object.is for NaN equality

francisjohnjohnston-web:
## What

Replace the `req <= val && req >= val` equality test with `Object.is(req, val) || (req <= val && req >= val)` in `value`, `notValue`, `values`, and `notValues`.

## Why

The ordering-comparison pattern was introduced to handle `Date` cross-instance comparison (two `new Date(x)` objects with the same time are `===`-unequal but `<=`/`>=`-equal). Unfortunately, all IEEE ordering comparisons with `NaN` return `false`:

```
NaN <= NaN  →  false
NaN >= NaN  →  false
```

This causes four correctness failures when `NaN` is used as a requirement:

| Action | Input | Expected | Actual |
|---|---|---|---|
| `value(NaN)` | `NaN` | **pass** | ❌ rejects |
| `notValue(NaN)` | `NaN` | **reject** | ❌ passes |
| `values([NaN])` | `NaN` | **pass** | ❌ rejects |
| `notValues([NaN])` | `NaN` | **reject** | ❌ passes |

These are reachable through `v.pipe(v.unknown(), v.value(NaN))` or any pipeline whose schema does not pre-reject `NaN` (e.g. `v.number()` filters NaN at schema level, but `v.unknown()` and custom transforms do not).

## Fix

Prepend `Object.is(req, val)` as a short-circuit:

```ts
// Before
this.requirement <= dataset.value && this.requirement >= dataset.value

// After
Object.is(this.requirement, dataset.value) ||
  (this.requirement <= dataset.value && this.requirement >= dataset.value)
```

`Object.is(NaN, NaN)` returns `true`, fixing NaN self-equality. `Object.is` returns `false` for distinct `Date` objects, so the `<=`/`>=` fallback still handles same-time Date comparison.

## Testing

One failing test added per action (4 total). Each test fails before the fix and passes after. Full library suite (272 test files / 3052 tests) remains green.

---
## Summary by cubic
Fix NaN equality in value-based validators by short-circuiting with Object.is. This makes `value`, `notValue`, `values`, and `notValues` handle NaN correctly while keeping Date comparisons working.

- **Bug Fixes**
  - Prepend `Object.is(req, val)` before `<=`/`>=` in `value`, `notValue`, `values`, `notValues`.
  - Correct outcomes: `value(NaN)` passes, `notValue(NaN)` rejects, `values([NaN])` passes, `notValues([NaN])` rejects.
  - Added one test per action; full suite remains green.

<sup>Written for commit b158a027631d923fe0f0eb5009af7f7b5ec85e50. Summary will update on new commits. <a href="https://cubic.dev/pr/open-circle/valibot/pull/1477?utm_source=github">Review in cubic</a></sup>


yslpn:
@francisjohnjohnston-web I like this fix.

I'd move this to util and reuse it. Because in the future, I'll need to reuse this same logic in other places `_merge.ts` or `literal`. So the logic doesn't have to be changed, but stays in the same place with comment. Could you please do this?

=> library/src/utils/ 

Pseudocode:

```ts
/**
 * Compares two values using the SameValueZero algorithm, which treats `NaN`
 * as equal to itself unlike `===`.
 *
 * @param value1 The first value.
 * @param value2 The second value.
 *
 * @returns Whether the values are equal.
 *
 * @internal
 */
// @__NO_SIDE_EFFECTS__
export function _sameValueZero(value1: unknown, value2: unknown): boolean {
  return value1 === value2 || (value1 !== value1 && value2 !== value2);
}
```

PR #1482, opened 2026-05-26
fix: `toJsonSchema` definition key with `/` or `~` not encoded in `$ref`

### LINKED ISSUE #1481: Bug: `toJsonSchema` definition key with `/` or `~` not encoded as JSON Pointer in `$ref`
## Description

According to [RFC 6901 (JSON Pointer)](https://datatracker.ietf.org/doc/html/rfc6901), reference tokens inside a JSON Pointer must escape two characters:

- `~` → `~0`
- `/` → `~1`

When `toJsonSchema()` builds `$ref` URI fragments from the `definitions` record keys, it uses the raw key string without applying these escapes. This produces `$ref` values that JSON Schema validators cannot correctly resolve.

The `$defs` key itself is a plain JSON object key and does not need escaping — only the pointer tokens inside `$ref` do.

## Steps to reproduce

```typescript
import * as v from 'valibot';

const UserSchema = v.object({ name: v.string() });

toJsonSchema(
  v.object({ user: UserSchema }),
  { definitions: { 'Shared/User~': UserSchema } }
);
```

## Actual output

```json
{
  "type": "object",
  "properties": {
    "user": {
      "$ref": "#/$defs/Shared/User~"
    }
  },
  "$defs": {
    "Shared/User~": {}
  }
}
```

The `$ref` pointer token `Shared/User~` is not encoded — `/` should be `~1` and `~` should be `~0`.

## Expected output

```json
{
  "type": "object",
  "properties": {
    "user": {
      "$ref": "#/$defs/Shared~1User~0"
    }
  },
  "$defs": {
    "Shared/User~": {}
  }
}
```

The `$defs` key stays as the original unescaped string. Only the `$ref` pointer path needs encoding.

dinwwwh:
Fixes #1481

---
## Summary by cubic
Fixes incorrect `$ref` generation when definition keys contain "/" or "~" by encoding them as JSON Pointer fragments. Prevents broken references and aligns with RFC 6901 (Fixes #1481).

- **Bug Fixes**
  - Added `getDefinitionRef` to encode definition keys in refs (`~` -> `~0`, `/` -> `~1`).
  - Updated `convertSchema` to use encoded `$ref` for both existing and newly created definitions.
  - Added tests for keys with "/" and "~" and updated CHANGELOG.

<sup>Written for commit b059013aa8d3ac5c7a6c25d970203ddf7c329623. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/open-circle/valibot/pull/1482?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


fabian-hiller:
I will try to get this merged today and publish a new version

fabian-hiller:
v1.7.1 should be available soon

PR #1490, opened 2026-06-04
feat: allow GlobalConfig.lang to be a function

canseyran:
## Motivation
Valibot currently supports setting, deleting and overwriting the global config. On server environments where messages may need to be translated depending on the user language, we need to include the language everytime we call a parse function. Also some frameworks like SvelteKit might internally call those functions, not allowing us to pass the user language.

## Solution
Allow the `lang` property in `GlobalConfig` to also accept a function, which resolves when calling `getGlobalConfig`. By passing a function we can get the request scoped language and it also would make integrating libraries like ParaglideJS easier, since they uses AsyncLocalStorage on the server for storing/retrieving the locale. 

```
import * as v from "valibot";
import { getLocale } from "$lib/paraglide/runtime";

v.setGlobalConfig({ lang: getLocale });
```

---
## Summary by cubic
Allow `GlobalConfig.lang` to be a function for request-scoped language resolution on the server. This enables dynamic locales in frameworks that can’t pass `lang` per call.

- **New Features**
  - `GlobalConfig.lang` accepts a string or `() => string`.
  - `getGlobalConfig` invokes the callback on each call; local `lang` still overrides.
  - Updated types and website docs (API `GlobalConfig` properties, i18n guide with a callback example).
  - Tests cover callback resolution, per-call invocation, and local override.

<sup>Written for commit 90aae3d876699493a565662a97fc04992c7310a0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/open-circle/valibot/pull/1490?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Language configuration now supports callback functions for dynamic language selection, enabling runtime language switching alongside static string values.

* **Documentation**
  * Updated the internationalization guide with an example demonstrating dynamic language configuration using callback functions.


fabian-hiller:
Hey, thank you for creating this PR! Just a quick question. Why is passing a function preferred over calling `v.setGlobalConfig` when you resolved the language. You can call `v.setGlobalConfig` multiple times.

If there a good reasons for using a function, I wonder if we should only allow a function for the global config or also for the local config.

canseyran:
> Hey, thank you for creating this PR! Just a quick question. Why is passing a function preferred over calling `v.setGlobalConfig` when you resolved the language. You can call `v.setGlobalConfig` multiple times.
> 
> If there a good reasons for using a function, I wonder if we should only allow a function for the global config or also for the local config.

Thanks for checking out the PR! The issue is that everytime you call `v.setGlobalConfig` you overwrite the previous language, this is fine for stateful environments like the browser, where you have may only have one language which is tied to the whole application. On stateless environments like the server, calling `v.setGlobalConfig` multiple times would overwrite the config for all current requests, causing a race condition. So users will very likely end up with a wrong translation, caused by another request with a different user language.

We could solve this by passing a function which resolves when `v.getGlobalConfig` is called either directly or by calling any parser function like `v.safeParse` and then get the language from the request context. Something like:

````
// Needs to be called only once on app init
v.setGlobalConfig({lang: () => getCurrentRequestContext().lang})
````


You mentioned passing a function for the local config aswell. I am not entirely sure how the configs are resolved within the whole code base, but as far as I can see the local config is resolved immediately when calling the function.

`````
v.safeParse(schema, data, { lang: getLocale()}

v.safeParse(schema, data, { lang: getLocale})
`````

Those two calls would up end up always having the same result, assuming the code snippet runs synchronously. But I might be wrong, if the local config is resolved differently in any part of the library, it might solve a problem, otherwise there is no practical difference or advantage between those two function calls.

canseyran:
I also found this issue on the sveltekit repository, where some people face the same issue. https://github.com/sveltejs/kit/issues/15137

fabian-hiller:
Thank you for the additional context! It helps me make a better decision. Why do you prefer calling `v.setGlobalConfig` in your stateless environment over setting the language explicitly when calling `v.safeParse`?

canseyran:
The main reason is that we can not always influence how and when a schema is validated, because libraries or frameworks might call standard schema validation internally with no way of hooking into it. Sveltekit remote functions take in a standard schema and internally call `schema["~standard"].validate(arg)`.


````
export const getPost = query(v.string(), async (slug) => {
	const [post] = await db.sql`
		SELECT * FROM post
		WHERE slug = ${slug}
	`;

	if (!post) error(404, 'Not found');
	return post;
});
````

https://github.com/sveltejs/kit/blob/65d3344d23dce05073b43ceb81aabe7e2a5ddf0c/packages/kit/src/runtime/app/server/remote/shared.js#L34

Although the responsibility could be shifted towards the frameworks, I think having the option to set the language dynamically is the easiest and most elegant solution.

PR #1493, opened 2026-06-08
fix(to-json-schema): encode JSON Pointer special chars in $ref tokens

### LINKED ISSUE #1481: Bug: `toJsonSchema` definition key with `/` or `~` not encoded as JSON Pointer in `$ref`

chatman-media:
## Bug

When `toJsonSchema()` is called with a `definitions` object whose keys contain `/` or `~`, the generated `$ref` URI fragments are invalid per [RFC 6901 (JSON Pointer)](https://datatracker.ietf.org/doc/html/rfc6901).

For example:

```ts
const UserSchema = v.object({ name: v.string() });
toJsonSchema(
  v.object({ user: UserSchema }),
  { definitions: { 'Shared/User~': UserSchema } }
);
```

**Actual** (broken):
```json
{ "user": { "$ref": "#/$defs/Shared/User~" } }
```

**Expected** (correct per RFC 6901):
```json
{ "user": { "$ref": "#/$defs/Shared~1User~0" } }
```

RFC 6901 requires two escapes inside pointer reference tokens:
- `~` → `~0`
- `/` → `~1`

The `$defs` object key itself is an ordinary JSON key and is **not** encoded — only the pointer path inside `$ref` needs encoding.

## Root cause

Both `$ref`-construction sites in `convertSchema.ts` used the raw `referenceId` string (e.g. `` `#/$defs/${referenceId}` ``) without applying the RFC 6901 escapes.

## Fix

- Adds `encodePointerToken(token: string): string` utility in `packages/to-json-schema/src/utils/encodePointerToken.ts`
- Applies it at both `$ref` construction sites in `convertSchema.ts`
- Adds a regression test covering `/` and `~` in definition keys

## Test evidence

```
Test Files  6 passed (6)
     Tests  222 passed (222)
Type Errors  no errors
```

Closes #1481

---
## Summary by cubic
Fixes invalid `$ref` fragments in `to-json-schema` when definition keys contain `/` or `~`. JSON Pointer tokens are now encoded per RFC 6901 so generated refs are valid.

- **Bug Fixes**
  - Add `encodePointerToken()` to escape `~` → `~0` and `/` → `~1` in `$ref` tokens.
  - Apply encoding at both `$ref` construction sites in `convertSchema.ts`.
  - Do not encode `$defs` keys; only the `$ref` pointer path is encoded.
  - Add a regression test for keys containing `/` and `~`.

<sup>Written for commit b3c824c01989d2a52db8c0f2a850bb78645619e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/open-circle/valibot/pull/1493?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/to-json-schema/src/converters/convertSchema/convertSchema.ts
+++ b/packages/to-json-schema/src/converters/convertSchema/convertSchema.ts
@@ -4,7 +4,11 @@ import type {
   ConversionContext,
   JsonSchema,
 } from '../../types/index.ts';
-import { addError, handleError } from '../../utils/index.ts';
+import {
+  addError,
+  encodePointerToken,
+  handleError,
+} from '../../utils/index.ts';
 import { convertAction } from '../convertAction/index.ts';
 
 /**
@@ -148,7 +152,7 @@ export function convertSchema(
     // If schema is in reference map use reference and skip conversion
     const referenceId = context.referenceMap.get(valibotSchema);
     if (referenceId) {
-      jsonSchema.$ref = `#/$defs/${referenceId}`;
+      jsonSchema.$ref = `#/$defs/${encodePointerToken(referenceId)}`;
       if (config?.overrideRef) {
         const refOverride = config.overrideRef({
           ...context,
@@ -631,7 +635,7 @@ export function convertSchema(
       }
 
       // Add reference to JSON Schema object
-      jsonSchema.$ref = `#/$defs/${referenceId}`;
+      jsonSchema.$ref = `#/$defs/${encodePointerToken(referenceId)}`;
 
       // Override reference, if necessary
       if (config?.overrideRef) {
--- a/packages/to-json-schema/src/functions/toJsonSchema/toJsonSchema.test.ts
+++ b/packages/to-json-schema/src/functions/toJsonSchema/toJsonSchema.test.ts
@@ -95,6 +95,29 @@ describe('toJsonSchema', () => {
       }
     });
 
+    test('for definition key containing JSON Pointer special characters', () => {
+      const UserSchema = v.object({ name: v.string() });
+      expect(
+        toJsonSchema(v.object({ user: UserSchema }), {
+          definitions: { 'Shared/User~': UserSchema },
+        })
+      ).toStrictEqual({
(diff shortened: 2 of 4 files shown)


chatman-media:
Closing — #1482 by @dinwwwh already covers this exact fix. Sorry for the noise!

fabian-hiller:
Thank you! I will merge #1482 soon.

fabian-hiller:
v1.7.1 should be available soon

PR #1517, opened 2026-06-21
fix(_merge): use SameValueZero for equality to handle NaN correctly

LeSingh1:
## Summary

The `_merge` internal utility (used by `intersect`) compares two values for equality using `===`. This causes two correctness bugs:

**Bug 1 — NaN merges fail incorrectly**

`NaN === NaN` is `false` in JavaScript, so `_merge(NaN, NaN)` returns `{ issue: true }` instead of `{ value: NaN }`. Any `intersect` schema whose options both produce `NaN` for the same key (e.g. via a `transform` action) incorrectly reports a merge-conflict error instead of succeeding.

```ts
const schema = intersect([
  object({ x: pipe(number(), transform(() => NaN)) }),
  object({ x: pipe(number(), transform(() => NaN)) }),
]);
schema['~run']({ value: { x: 1 } }, {});
// Before: { typed: false, issues: [{ type: 'intersect', … }] }  ← wrong
// After:  { typed: true,  value: { x: NaN } }                  ← correct
```

**Bug 2 — Invalid Date merges fail incorrectly**

`+new Date(NaN)` is `NaN`, so `_merge(new Date(NaN), new Date(NaN))` also returns `{ issue: true }`. Two invalid dates should be considered equal and merge successfully.

**Bonus — -0 / +0 distinction**

Switching to `Object.is` also makes `-0` and `+0` distinct (they are equal under `===` but different under `Object.is`). This is consistent with the fix already applied to the `value` / `notValue` / `values` / `notValues` actions.

## Changes

- `library/src/schemas/intersect/utils/_merge/_merge.ts` — replace `value1 === value2` with `Object.is(value1, value2)` and `+value1 === +value2` with `Object.is(+value1, +value2)`.
- `library/src/schemas/intersect/utils/_merge/_merge.test.ts` — add test cases for NaN primitives, invalid Date objects, and mismatched -0/+0.

## Test plan

- [ ] `pnpm -C library test` passes (all 33 intersect tests green, 3 new cases added)
- [ ] No other test files changed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved merging of special numeric values, including treating `NaN` values and signed zeros as equivalent.
  * Improved merging of dates with invalid (`NaN`) timestamps and timestamps of `-0` or `+0`.

* **Tests**
  * Added coverage for merging `NaN`, invalid dates, and dates with signed-zero timestamps.


fabian-hiller:
Thank you for this PR. How did you find out about this behavior? Did you run into this issue?

> is consistent with the fix already applied to the value / notValue / values / notValues actions.

Not sure if this is correct. I think -0 and +0 should be treated as the same value.

yslpn:
@LeSingh1 Sorry for delay. Before merging this, we need to replace `Object.is(value1, value2)` with the SameValueZero equivalent (for example, `value1 === value2 || (Number.isNaN(value1) && Number.isNaN(value2))`, similarly for the Date branch), so as not to break -0/+0 and stay in the same semantic field with the other two PRs.

LeSingh1:
@yslpn Thanks — done in 46b8eec.

`Object.is` is now replaced with SameValueZero, so `NaN` still merges with itself but `-0` and `+0` stay interchangeable. That also resolves @fabian-hiller's point above that `-0` and `+0` should be treated as the same value; the original `Object.is` approach did regress that, and the PR title/description have been updated accordingly.

Rather than inlining the comparison twice, I pulled it into a `_sameValueZero` helper so the primitive branch and the `Date` branch can't drift apart:

```ts
function _sameValueZero(value1: unknown, value2: unknown): boolean {
  return (
    value1 === value2 ||
    (typeof value1 === 'number' &&
      typeof value2 === 'number' &&
      Number.isNaN(value1) &&
      Number.isNaN(value2))
  );
}
```

The explicit `typeof` guards are there because the outer `typeof value1 === typeof value2` check doesn't narrow `value2` for TypeScript. Happy to inline it in both places instead if you'd prefer to keep the diff to the single condition.

Test coverage now includes signed zeros in both branches (`_merge(-0, 0)` and `_merge(new Date(-0), new Date(0))`) alongside the existing `NaN` and invalid-`Date` cases.

One unrelated thing I fixed while here: the invalid-date test had an operator-precedence bug — `+(result as { value: unknown }).value as number` applied the unary plus to an `unknown` before the cast, which failed `tsc`. It's now `+(result as { value: Date }).value`. `pnpm lint` and the full library suite (3060 tests) pass locally.

PR #1521, opened 2026-06-27
fix: cache word-count segmenter for non-primitive locales

chatman-media:
The internal \`_getWordCount\` util caches the \`Intl.Segmenter\` in a \`Map\` keyed by the raw \`locales\` argument:

\`\`\`ts
if (!store.get(locales)) {
  store.set(locales, new Intl.Segmenter(locales, { granularity: 'word' }));
}
\`\`\`

But \`Intl.LocalesArgument\` is commonly a **non-primitive** value — an array of locales (e.g. \`['en-US', 'de']\`) or an \`Intl.Locale\` object. Those are compared by reference in a \`Map\`, so a fresh-but-equal argument on each call always misses the cache and a brand-new \`Intl.Segmenter\` is constructed on **every** \`words\` / \`minWords\` / \`maxWords\` / \`notWords\` validation. The cache only ever hits for plain string locales.

### Fix
Key the cache by \`String(locales)\` (a stable BCP-47 string for strings, arrays and \`Intl.Locale\` objects) so equal locale arguments reuse the same segmenter. Counting behavior is unchanged.

### Test
Added a test that spies on \`Intl.Segmenter\` and passes the same locale as both an array and an \`Intl.Locale\` across four calls. Before the fix the constructor runs 4 times; after, exactly once.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved word counting for actions that validate word counts by reusing the same segmentation instance for equivalent locale inputs, reducing inconsistent results across repeated validations.
* **Tests**
  * Updated the word-count test suite to restore mocks after each test.
  * Added coverage to ensure locale-equivalent inputs (arrays and locale objects) reuse cached segmentation so it’s created only once.

### DIFF
--- a/library/CHANGELOG.md
+++ b/library/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the library will be documented in this file.
 
+## vX.X.X (Month DD, YYYY)
+
+- Fix word count actions to cache the `Intl.Segmenter` for non-primitive locales, preventing it from being recreated on every `words`, `minWords`, `maxWords` and `notWords` validation (pull request #1521)
+
 ## v1.4.1 (May 24, 2026)
 
 - Fix `intersect` schema to infer correct input and output types for non-tuple array options instead of `never` (pull request #1478)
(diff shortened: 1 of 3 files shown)


fabian-hiller:
Thank you!

PR #1523, opened 2026-06-29
fix: handle object keys that collide with Object.prototype in object schemas

mahirhir:
## Summary

`strictObject`, `looseObject`, `objectWithRest` and their async variants check whether an input key is a defined schema entry with `key in this.entries`. The `in` operator also matches inherited `Object.prototype` members, so an input key named like a prototype member is treated as a defined entry.

This is the same class of bug as the `flatten` and `intersect` fixes in #1522, here in the entry-membership check.

## Effect

For an input key that collides with a prototype member (`toString`, `valueOf`, `constructor`, `__proto__`, ...):

- `strictObject` / `strictObjectAsync` fail to reject it as an unknown key, even though it is not a defined entry. A `__proto__` or `constructor` key passes a strict schema silently.
- `looseObject` / `looseObjectAsync` drop it instead of passing it through.
- `objectWithRest` / `objectWithRestAsync` skip it instead of validating it against the `rest` schema.

Before the fix:

```ts
v.parse(v.strictObject({ name: v.string() }), { name: 'a', toString: 'b' });
// returns { name: 'a' }; the unknown `toString` key is not reported
```

## Fix

Replace `key in this.entries` with `Object.prototype.hasOwnProperty.call(this.entries, key)`, the same own-property check already used by `_isValidObjectKey` for input keys and by the #1522 fixes. A schema that intentionally defines an entry named like a prototype member (`v.object({ toString: v.string() })`) still works, since `hasOwnProperty` reports developer-defined own entries.

## Tests

Added a test to each affected schema: the prototype-named key is rejected for strict, passed through for loose, and validated against `rest` for object-with-rest. Lint, Prettier and the affected Vitest suites pass for the changed files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected object schema handling for keys that overlap with built-in object properties.
  - Declared keys such as `toString`, `valueOf`, `hasOwnProperty`, `constructor`, and `prototype` are now validated correctly.
  - Unknown keys are classified consistently, while unsafe keys such as `__proto__` remain excluded from parsed output.
  - Applied the fix across strict, loose, rest-aware, and asynchronous object schemas.

- **Tests**
  - Added coverage for prototype-colliding keys and unsafe key handling.

### DIFF
--- a/library/CHANGELOG.md
+++ b/library/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to the library will be documented in this file.
 
+## vX.X.X (Month DD, YYYY)
+
+- Fix `strictObject` and `strictObjectAsync` to reject unknown input keys that collide with `Object.prototype` members like `toString` instead of accepting them silently (pull request #1523)
+- Fix `looseObject`, `looseObjectAsync`, `objectWithRest` and `objectWithRestAsync` to pass through or apply `rest` to input keys that collide with `Object.prototype` members instead of skipping them (pull request #1523)
+
 ## v1.4.2 (June 28, 2026)
 
 - Fix word count actions to cache the `Intl.Segmenter` for non-primitive locales, preventing it from being recreated on every `words`, `minWords`, `maxWords` and `notWords` validation (pull request #1521)
--- a/library/CHANGELOG.md
+++ b/library/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to the library will be documented in this file.
 - Fix `literal` schema and `value`, `values`, `notValue` and `notValues` actions to treat `NaN` as equal to itself (pull request #1573)
 - Fix `intersect` schema to merge matching `NaN` values and invalid dates (pull request #1573)
 - Fix `cache` and `cacheAsync` methods to clone the issues of a cached dataset, preventing parent schemas from adding their path item to the same issue on every cache hit (pull request #1620)
+- Fix `strictObject`, `looseObject`, `objectWithRest` and their async variants to correctly handle unknown input keys that collide with `Object.prototype` members (pull request #1523)
 
 ## v1.4.2 (June 28, 2026)
 
--- a/library/src/schemas/looseObject/looseObject.test.ts
+++ b/library/src/schemas/looseObject/looseObject.test.ts
@@ -672,4 +672,16 @@ describe('looseObject', () => {
       } satisfies FailureDataset<InferIssue<typeof schema>>);
     });
   });
+
+  describe('should pass through keys colliding with the object prototype', () => {
+    const schema = looseObject({ key: string() });
+
+    test('for unknown key named like an object prototype member', () => {
+      const input = { key: 'foo', toString: 'bar' };
+      expect(schema['~run']({ value: input }, {})).toStrictEqual({
+        typed: true,
+        value: { key: 'foo', toString: 'bar' },
+      });
+    });
+  });
 });
--- a/library/src/schemas/looseObject/looseObject.test.ts
+++ b/library/src/schemas/looseObject/looseObject.test.ts
@@ -672,4 +672,46 @@ describe('looseObject', () => {
       } satisfies FailureDataset<InferIssue<typeof schema>>);
     });
   });
+
+  describe('should pass through keys colliding with the object prototype', () => {
+    const schema = looseObject({ key: string() });
+
+    test.each(['toString', 'valueOf', 'hasOwnProperty'])(
+      'for unknown %s key',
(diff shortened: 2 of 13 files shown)


fabian-hiller:
Thank you for this PR! I plan to take a closer look before v1.5.

yslpn:
Hi @mahirhir Thanks for your input. Are you interested in further developing this PR? I see there are some comments from AI review. It seems like the problem is real.

What do you think about this?

And you have conflicts

mahirhir:
Yes, still interested. I will rebase onto current main and address the cubic findings; the prototype-inheritance one is real (the presence checks use `key in input`, so a schema entry named like an Object.prototype member is treated as present when the input only inherits it) and an own-property check is the right shape for the fix. Will push the rebase with that and a regression test.

mahirhir:
Rebased onto current main and pushed (7657fa50). The only conflict was the unreleased CHANGELOG section, resolved by appending the two #1523 entries after the existing ones.

One correction to my last comment: I wrote that the presence checks use `key in input`. They actually used `key in this.entries`, the input side was already guarded by `_isValidObjectKey`. The fix replaces that entries check with `Object.prototype.hasOwnProperty.call(this.entries, key)` in the six sync/async variants of strictObject, looseObject and objectWithRest, each with a regression test.

Locally: vitest on those three schema folders passes 6 files / 198 tests, and eslint and prettier are clean on the changed files.

yslpn:
> Rebased onto current main and pushed ([7657fa5](https://github.com/open-circle/valibot/commit/7657fa50c121e4f6c287cf16826d4fe154a2c1fb)). The only conflict was the unreleased CHANGELOG section, resolved by appending the two #1523 entries after the existing ones.
> 
> One correction to my last comment: I wrote that the presence checks use `key in input`. They actually used `key in this.entries`, the input side was already guarded by `_isValidObjectKey`. The fix replaces that entries check with `Object.prototype.hasOwnProperty.call(this.entries, key)` in the six sync/async variants of strictObject, looseObject and objectWithRest, each with a regression test.
> 
> Locally: vitest on those three schema folders passes 6 files / 198 tests, and eslint and prettier are clean on the changed files.

@mahirhir  Could you please resolve the comments from the AI ​​code review or fix them if there are any issues? This will speed up the review. I'll take a closer look after that. Thank you

mahirhir:
@yslpn Addressed the AI review findings in `2d4b2a44`. cubic was right and I was wrong: in my earlier comment I said the input side was already guarded by `_isValidObjectKey`. It is not — `_isValidObjectKey` only guards the unknown-key/rest loops, and the defined-entry presence checks really did use `key in input`. Sorry for the misdirection.

I reproduced it before touching anything:

```
object({ toString: string() })      parsing {}  ->  issue { type: 'string', received: 'Function' }
strictObject({ toString: optional(string(), 'x') }) parsing {}  ->  fails instead of applying 'x'
```

So a schema entry named like an `Object.prototype` member was validated against the *inherited* function rather than reported missing, and a default was never applied.

Scope is wider than the two files cubic flagged. `grep -n "key in input"` found the same pattern in **8** schemas, including `object` and `objectAsync`, which this PR had not touched. Fixing only `looseObject` and `strictObject` would have left `object` — the most used one — broken, so all eight are covered now.

Changes:

- Both presence checks per schema now use one hoisted `isKeyPresent` own-property check. Hoisting keeps it to a single lookup per entry rather than the two the previous code did, which seemed worth it given the hot-path work in the v1.4.0 notes.
- In the missing-required-key branch the issue path value was `input[key]`. That branch is only reached when the key is *not* an own property, so the value could only ever be an inherited one; it is now `undefined`, which is what the existing missing-key tests already expect.

Verification:

- 16 new regression tests (2 per schema, sync and async). I checked they discriminate: reverting the source change fails all 16, restoring it passes them.
- Full suite: 278 files / 3130 tests pass. `tsc --noEmit`, eslint, and prettier are all clean.

One thing I deliberately did **not** do: CodeRabbit also suggested replacing `for (const key in this.entries)` with own-entry enumeration. That is a separate concern about inherited *schema* keys rather than inherited *input* keys, and the schema object is author-controlled, so I left it out to keep this PR to one defect. Happy to add it here or in a follow-up if you would rather have it.

fabian-hiller:
Updated this PR on `main` and narrowed it for v1.5 to inline own-property checks for schema-entry membership. Broader inherited-input changes are deferred for separate compatibility and performance review. Expanded regression coverage across all six affected implementations and added the changelog entry. All 4,667 library tests, type checks, lint, and formatting pass. We are continuing here instead of #1536, which addresses the same lookup bug.

PR #1525, opened 2026-06-29
feat: add `rfcTime` (RFC 3339 time) validation action

tungyhardDevOps:
Closes #1496

## Summary

Adds an `rfcTime` validation action for [RFC 3339 full-time](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6) strings — seconds, optional fractional seconds, and a mandatory `Z` (case-insensitive) or `±hh:mm` time offset. This complements the existing `isoTime` (`hh:mm`) and `isoTimeSecond` (`hh:mm:ss`) actions, which validate local wall-clock times with no offset, and improves interop with OpenAPI `format: time` and other RFC 3339 APIs.

## Accepts

`00:00:00Z`, `14:30:45Z`, `14:30:45z`, `23:59:59-05:00`, `14:30:45+02:00`, `14:30:45.123Z`, `14:30:45.123456789+14:00`

## Rejects

`14:30:45` (no offset), `14:30Z` (no seconds), `24:00:00Z`, `14:30:45+99:00` (offset out of range), `14:30:45+0200` (missing colon), `14:30:45 Z` (space)

## Implementation

- New `RFC_3339_TIME_REGEX` in `library/src/regex.ts`.
- New action `library/src/actions/rfcTime/`, modelled on `isoTime` (same overloads, issue/action interfaces, `@__NO_SIDE_EFFECTS__` annotation).
- Exported from `library/src/actions/index.ts`.
- Unit tests (`rfcTime.test.ts`) + type tests (`rfcTime.test-d.ts`) covering the accept/reject sets above.

Note: leap second `:60` is intentionally excluded to match `isoTimeSecond`'s behaviour — happy to include it if preferred. Also open to a different action name (e.g. keeping it in the `iso*` family) if that fits the library's conventions better.

## Summary by CodeRabbit

* **New Features**
  * Added RFC 3339 full-time validation with required seconds, optional fractional seconds, timezone offsets, and leap-second handling.
  * Exposed the validation action through the main actions entry point and a dedicated directory entry.
  * Added reusable patterns for RFC 3339 times and KSUIDs.
* **Documentation**
  * Added API documentation and usage guidance for RFC time validation and related types.
* **Tests**
  * Added runtime and type-level coverage for valid and invalid times, messages, dataset passthrough, validation issues, and inferred types.

PR #1526, opened 2026-06-29
feat: add `rfcDuration` (RFC 3339 duration) validation action

### LINKED ISSUE #1497: Feature Request: RFC 3339 `duration` Validator

tungyhardDevOps:
Closes #1497

## Summary

Adds an `rfcDuration` validation action for [RFC 3339 duration](https://datatracker.ietf.org/doc/html/rfc3339#appendix-A) strings (`PnYnMnDTnHnMnS` and the week form `PnW`), so users can validate OpenAPI `format: duration` fields without hand-rolling a regex.

## Accepts

`P1D`, `PT5H30M`, `P1Y2M3DT4H5M6S`, `PT0S`, `P1W`, `P1Y`, `PT5H`, `P1DT2H`, `P3DT4H5M6S`

## Rejects

`P` / `PT` (no components), `1Y` (no `P` prefix), `P1H` (time part outside the `T` section), `P1D2M` (wrong order), `P1WT5H` (week form is exclusive), `P1.5Y` (fractional), `P-1D` (negative)

## Implementation

- New `RFC_3339_DURATION_REGEX` in `library/src/regex.ts` — requires at least one component, enforces `Y M D T H M S` order, and treats `W` (week) as exclusive.
- New action `library/src/actions/rfcDuration/`, modelled on the existing `iso*`/`rfcEmail` actions (same overloads, issue/action interfaces, `@__NO_SIDE_EFFECTS__`).
- Exported from `library/src/actions/index.ts`.
- Unit tests (`rfcDuration.test.ts`) + type tests (`rfcDuration.test-d.ts`).

Notes (happy to adjust to the library's preference):
- Fractional seconds are excluded to match RFC 3339 Appendix A (integer values only).
- The time section accepts any in-order subset of `H`/`M`/`S` (common ISO-8601 behaviour); RFC 3339's strict ABNF nests these — can tighten if preferred.

## Summary by CodeRabbit

* **New Features**
  * Added an RFC 3339 duration validation action with optional typed error messaging.
  * Re-exported the duration validator from the main actions entry point for easier access.
  * Exposed an RFC 3339 duration matching pattern for consumers.
* **Bug Fixes**
  * Expanded duration validation to consistently flag malformed inputs (empty/whitespace, missing parts, wrong ordering, fractional/negative values, and week-form restrictions).
* **Tests**
  * Added compile-time and runtime tests covering action typing and validation behavior.

### DIFF
--- a/library/src/actions/index.ts
+++ b/library/src/actions/index.ts
@@ -90,6 +90,7 @@ export * from './readonly/index.ts';
 export * from './reduceItems/index.ts';
 export * from './regex/index.ts';
 export * from './returns/index.ts';
+export * from './rfcDuration/index.ts';
 export * from './rfcEmail/index.ts';
 export * from './safeInteger/index.ts';
 export * from './size/index.ts';
--- a/library/src/actions/rfcDuration/index.ts
+++ b/library/src/actions/rfcDuration/index.ts
@@ -1,0 +1,1 @@
+export * from './rfcDuration.ts';
--- a/library/src/actions/rfcDuration/rfcDuration.test-d.ts
+++ b/library/src/actions/rfcDuration/rfcDuration.test-d.ts
@@ -1,0 +1,49 @@
+import { describe, expectTypeOf, test } from 'vitest';
+import type { InferInput, InferIssue, InferOutput } from '../../types/index.ts';
+import {
+  rfcDuration,
+  type RfcDurationAction,
+  type RfcDurationIssue,
+} from './rfcDuration.ts';
+
+describe('rfcDuration', () => {
+  describe('should return action object', () => {
+    test('with undefined message', () => {
+      type Action = RfcDurationAction<string, undefined>;
+      expectTypeOf(rfcDuration<string>()).toEqualTypeOf<Action>();
+      expectTypeOf(
+        rfcDuration<string, undefined>(undefined)
+      ).toEqualTypeOf<Action>();
+    });
+
+    test('with string message', () => {
+      expectTypeOf(rfcDuration<string, 'message'>('message')).toEqualTypeOf<
+        RfcDurationAction<string, 'message'>
+      >();
+    });
+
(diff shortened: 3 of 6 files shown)


yslpn:
What do you think about this implementation? https://github.com/open-circle/valibot/pull/1520

tungyhardDevOps:
Flagging that this overlaps with #1520 — both close #1497. To avoid duplicated effort, here's how this PR is scoped, in case it helps decide direction:

- **Naming:** follows valibot's existing `rfcEmail` convention (`rfcTime` / `rfcDuration`) — keeps the `rfc<Thing>` pattern rather than embedding the spec number in the name.
- **Pairing:** companion to #1525 (`rfcTime`), so the two land as one consistent RFC 3339 family.
- **Regex:** derived from the RFC 3339 Appendix A `duration` ABNF (order enforced, `W` exclusive, ≥1 component). I verified it accepts/rejects identically to the proven [ajv-formats](https://github.com/ajv-validator/ajv-formats) `duration` regex across ~47k generated inputs, so behaviour matches the established reference.
- **Tests:** full unit coverage plus `.test-d.ts` type-level tests (20 passing, `tsc` typecheck clean).

Happy to align naming or defer to #1520 — whatever you prefer. Just wanted to give the full picture. 🙏

fabian-hiller:
Thank you! I will try to take a closer look before v1.5.

yslpn:
Thanks all for the PRs (#1520, #1526, #1540). We're moving forward with #1520 first. If it stalls, we'll fall back to #1526 or #1540 instead of blocking the issue.

PR #1527, opened 2026-06-30
perf: Benchmarks and optimizations

kibertoad:
This change adds a runtime benchmark suite and three hot-path optimizations, each proven by before/after measurements. All 4486 tests pass with no behavior or type changes.

## Why these paths

Profiling the schema-execution hot paths and comparing against zod showed three places where valibot did linear work that could be made constant-time, without touching its functional, tree-shakeable design or resorting to `eval`/JIT.

## 1. `variant` / `variantAsync` discriminator-map dispatch

Before, validating a discriminated union scanned the options one by one, running each option's discriminator sub-schema against the input until one matched. Cost grew with the number of options: the last option of ten was ~13x slower than the first.

Now a `Map<discriminatorValue, option>` is built lazily on the first run, so dispatch is a single `Map.get` regardless of option count. The map is disabled (falling back to the original scan) for nested variants, non-enumerable discriminators, colliding values, or `NaN`, so error reporting and every edge case stay identical. Result: dispatch is position-independent (up to ~13x on the worst
case, ~8x for the async schema), and even the first-option case is faster because the fast path skips the per-call closure and `Set` allocations of the scan.

## 2. `enum` / `picklist` set membership

Before, membership used `Array.includes`, an O(n) scan, so checking a value near the end of a large allow-list was ~3x slower than one near the start. Now a lazily-cached `Set` makes `has` O(1) and position-independent. `Set.has` uses the same SameValueZero comparison as `Array.includes`, so behavior (including `NaN`) is unchanged.

## 3. `object` family loop

The entry loop evaluated `key in input` twice per present key (once in the condition, once in the value ternary). Hoisting it to a single check gives a small, consistent win on the object happy path with no behavior change. Applied to `object`, `strictObject`, `looseObject`, and `objectWithRest`.

## Design notes

- Precomputed state (the discriminator map, the option set) is built lazily on first run, so schema creation stays cheap and tree-shaking is unaffected.
- The fast paths only change the success-dispatch case; any ambiguity routes to the original logic, which is why the full test suite passes unchanged.

## Benchmarks

1. `variant` and `variantAsync` discriminator-map fast path (O(1) dispatch instead of scanning every option).
2. `object` family: hoist the redundant `key in input` re-check in the entry loop.
3. `enum` / `picklist` lazily-cached `Set` membership instead of `Array.includes`.

### Baseline vs. optimized (speedup = optimized / baseline)

| Benchmark | Case | Baseline hz | Optimized hz | Speedup |
| --- | --- | ---: | ---: | ---: |
| variant (10 options) | hit first option | 1,674,293 | 3,252,833 | 1.94x |
| variant (10 options) | hit middle option | 340,541 | 3,119,117 | 9.16x |
| variant (10 options) | hit last option | 236,108 | 3,195,553 | 13.54x |
| variant (10 options) | miss | 252,662 | 240,129 | 0.95x (unchanged*) |
| object | all keys present | 1,669,082 | 2,216,805 | 1.33x |
| object | missing required key | 986,683 | 1,259,883 | 1.28x |
| enum (200) | hit first | 6,662,908 | 7,738,684 | 1.16x |
| enum (200) | hit last | 2,453,115 | 6,979,579 | 2.85x |
| enum (200) | miss | 1,797,455 | 2,289,587 | 1.27x |
| picklist (200) | hit first | 6,732,057 | 6,836,693 | 1.02x |
| picklist (200) | hit last | 1,962,357 | 7,151,161 | 3.64x |
| picklist (200) | miss | 1,764,085 | 2,363,121 | 1.34x |
| array (100 objects) | valid | 91,765 | 102,279 | 1.11x (noise, untouched) |
| string pipe | valid | 5,086,026 | 4,827,814 | 0.95x (noise, untouched) |

\* The `variant` miss case is unchanged by design: a non-matching discriminator
falls through to the original slow path so the discriminator issue and message
are byte-identical to before.

## Summary by CodeRabbit

* **Performance**
  * Improved enum and picklist validation performance with cached lookups.
  * Accelerated synchronous and asynchronous variant parsing with optimized discriminator dispatch.
  * Streamlined key-presence handling during object, strict object, loose object, and object-with-rest parsing.
* **Benchmarking**
  * Added runtime benchmarks covering arrays, objects, enums, picklists, variants, and string pipelines.
  * Added a command for running benchmarks locally.
* **Documentation**
  * Added benchmarking documentation and contributor guidance for recording performance results.


library/src/schemas/picklist/picklist.ts
```diff
@@ -101,8 +101,14 @@ export function picklist(
...
     '~run'(dataset, config) {
+      // Lazily cache the options as a set for O(1) membership checks. This is
+      // faster than `Array.includes` for large option lists and uses
+      // SameValueZero comparison, so behavior is identical.
       // @ts-expect-error
-      if (this.options.includes(dataset.value)) {
+      const optionsSet: Set<unknown> = (this._optionsSet ??= new Set(
+        this.options
+      ));
+      if (optionsSet.has(dataset.value)) {
```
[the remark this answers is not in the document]
kibertoad: expects is already snapshotted at creation, so mutating options afterward was already unsupported before this PR.

kibertoad:
@fabian-hiller any chance you could take a look at this?

fabian-hiller:
Thank you for this PR. I will try to take a closer look and get this merged before v1.5.

library/src/schemas/enum/enum.ts
```diff
@@ -144,8 +144,18 @@ export function enum_(
...
+      // Lazily cache the options as a set for O(1) membership checks. This is
+      // faster than `Array.includes` for large enums and uses SameValueZero
+      // comparison, so behavior is identical. The set is built once and assumes
+      // `options` is not mutated after schema creation, which matches the
+      // existing contract: `expects` is already precomputed from `options`
+      // above, so a post-creation mutation would desync the error message
+      // regardless.
       // @ts-expect-error
-      if (this.options.includes(dataset.value)) {
+      const optionsSet: Set<unknown> = (this._optionsSet ??= new Set(
```
yslpn, line 155: What do you think about not writing the cache to the object's observable properties?

1) This contradicts the intuition that parsing doesn't change the object's properties.
2) This changes the object's observable form; the property becomes public and enumerable.
3) For some extreme cases, the behavior might change, like Object.freeze(v.enum(...))

The cache can be stored at the module level.

library/src/schemas/variant/variant.ts
```diff
@@ -19,6 +19,67 @@ import type {
...
+ *
+ * @param key The discriminator key.
+ * @param options The variant options.
+ *
+ * @returns The discriminator map or `null`.
+ *
+ * @internal
+ */
+// @__NO_SIDE_EFFECTS__
+function _buildDiscriminatorMap(
```
yslpn, line 39: 1) You're duplicating `_buildDiscriminatorMap` between the sync and async variants. Move the common synchronous logic into a utility and cover it with tests.

2) I'm not sure the map code is a good place to store any functions at all. Look at how it's done elsewhere in the library; they're either separate files or utils.

BENCHMARKING.md
```diff
@@ -0,0 +1,63 @@
+# Benchmarking
+
+Valibot has a runtime benchmark suite for catching performance regressions and proving
```
yslpn, line 3: Benchmarks currently measure performance, but they don't automatically catch regressions: the vitest bench simply prints numbers. A true regression guard requires a CI baseline/comparison service. 

Any ideas on how to add this?

library/src/schemas/variant/variant.ts
```diff
@@ -117,6 +178,34 @@ export function variant(
...
+        if (discriminatorMap === undefined) {
+          discriminatorMap = _buildDiscriminatorMap(this.key, this.options);
+          // @ts-expect-error
+          this._discriminatorMap = discriminatorMap;
+        }
+
+        // Fast path: dispatch directly to the single option whose discriminator
+        // matches the input. On a miss, fall through to the slow path so the
+        // discriminator issue and message are produced identically.
+        if (discriminatorMap && this.key in input) {
```
yslpn, line 198: Add functional fast path tests. Variant dispatching is currently changing, but there are no tests for literal/enum/picklist discriminators, collisions, NaN, missing key, and parity sync/async in diff.

kibertoad: Sorry for the late reaction, it's been a while, and I missed the notifications among all the github noise I'm getting. I will address all the feedback shortly

library/src/schemas/enum/enum.ts
kibertoad: Agreed, fixed in 182f109d. The set now lives in the factory closure instead of a property on the schema object, so parsing no longer changes the schema's observable shape and `Object.freeze(v.enum(...))` keeps working. Same for `picklist`. I preferred the closure over a module-level `WeakMap` because it avoids the extra lookup on every `~run` call.

Added a `should not mutate the schema object` test to both `enum.test.ts` and `picklist.test.ts` that freezes the schema and asserts `Reflect.ownKeys` is unchanged after parsing.

library/src/schemas/variant/variant.ts
kibertoad: Both done in 182f109d.

The helper now lives in `library/src/utils/_buildDiscriminatorMap/` following the existing `utils/_*` layout (`_buildDiscriminatorMap.ts`, `_buildDiscriminatorMap.test.ts`, `index.ts`), and is generic over the sync and async option types so `variant` and `variantAsync` share one implementation. It has 10 tests of its own covering literal, enum and picklist discriminators, mixed sync/async options, nested variants, missing keys, non-enumerable discriminators and collisions.

The variant schema files now only hold the closure-level cache and the dispatch itself.

library/src/schemas/variant/variant.ts
kibertoad: Added in 182f109d, as a `should dispatch via discriminator fast path` block in both `variant.test.ts` and `variantAsync.test.ts`:

- literal discriminators, asserting that only the option the discriminator points at runs (its own issues come back, not another option's)
- enum and picklist discriminators
- colliding values, which disable the map so every matching option is still tried
- `NaN`
- missing discriminator key and present-but-`undefined` discriminator
- non-enumerable discriminators (`string()`), which disable the map
- schema object is not mutated by parsing (frozen schema, `Reflect.ownKeys` unchanged)
- sync/async parity: the same options run through `variant` and `variantAsync` over 10 inputs (hits, body failures, unknown discriminator, undefined discriminator, `{}`, a string, `null`) and the datasets must be identical

One thing the tests turned up: the `NaN` bail-out was based on a wrong premise. `literal` compares with `_isSameValueZero`, not `===`, which is exactly the comparison `Map` keys use, and `enum`/`picklist` membership is SameValueZero too. So `NaN` (and `-0`) dispatch through the map exactly as they validate, and the guard is gone. The tests above pin that down.

BENCHMARKING.md
kibertoad: Right, the suite measures and reports, it does not gate. A real guard needs a stored baseline and a comparison step in CI, which is a separate piece of infrastructure (and #1609 is already looking at CI benchmark comparison). I would rather not grow this PR into that, so I am leaving it out of scope here and keeping this PR to the suite plus the three optimizations.

yslpn:
@kibertoad I like the ideas. But the PR is doing too much.

It adds benchmarks and includes fixes for `variant`, `object family`, and `enum` / `picklist`.

It would be great to split this into several PRs:

1) Fix for `object family`
2) Fast path for `variant`
3) `Set` support for `enum` / `picklist`

or more easy

1) Fix for `object family` + `Set` support for `enum` / `picklist`
2) Fast path for `variant`

I think the addition of benchmarks can be discussed here: https://github.com/open-circle/valibot/pull/1609. I’d appreciate it if you could review it.

What do you think?

kibertoad:
@yslpn Sure, I can split the optimization part and drop the benchmark part.

kibertoad:
@yslpn Done, split into three PRs off current `main`, benchmarks dropped:

1. #1639 — `object` family key presence check
2. #1640 — `Set` membership for `enum` / `picklist`
3. #1641 — discriminator map fast path for `variant` / `variantAsync`

They touch disjoint files, so they can be reviewed and merged in any order. Each one carries the tests from this PR that belong to it. I'll leave this PR open until they land, then close it. Will take a look at #1609 for the benchmarks.

yslpn:
Thank you so much! I will take a look

PR #1529, opened 2026-06-30
fix(literal): use Object.is so literal(NaN) accepts NaN

spokodev:
## Problem

`literal(NaN)` is an unsatisfiable schema: it rejects every input, including the `NaN` it was constructed from.

```ts
import * as v from 'valibot';

v.parse(v.literal(NaN), NaN); // throws: Invalid type: Expected NaN but received NaN
v.is(v.literal(NaN), NaN);    // false
```

`NaN` is a valid `number`, and `Literal = bigint | boolean | number | string | symbol`, so `literal(NaN)` is a fully type-checked construction. The runtime can never succeed because the value check uses `===`, and `NaN === NaN` is `false`.

The sibling `picklist` schema disagrees, because it compares with `Array.includes` (SameValueZero):

```ts
v.is(v.picklist([NaN]), NaN); // true
```

So `literal` and `picklist` give opposite answers for the same value, which points to `literal`'s `===` being an oversight rather than a deliberate choice.

## Fix

`library/src/schemas/literal/literal.ts` compares `dataset.value === this.literal`. Switch that single comparison to `Object.is`:

```ts
// before
if (dataset.value === this.literal) {
// after
if (Object.is(dataset.value, this.literal)) {
```

`Object.is(NaN, NaN)` is `true`, so `literal(NaN)` now accepts `NaN`.

This matches the equality migration already in flight: #1517 moves `_merge` (used by `intersect`) to `Object.is`, and #1477 moves `value` / `notValue` / `values` / `notValues` to `Object.is`. Neither of those touches the `literal` schema, so this closes the same NaN gap one layer up.

## Trade-off

Like #1517 and #1477, `Object.is` also distinguishes `-0` from `+0`. After this change `literal(0)` no longer accepts `-0` (under `===` they were treated as equal). This is consistent with the direction of those PRs. If matching `picklist`'s SameValueZero (which keeps `-0`/`+0` loosely equal) is preferred instead, I can switch to that.

## Tests

Added one case to `library/src/schemas/literal/literal.test.ts`:

```ts
test('for valid NaN literal', () => {
  expectNoSchemaIssue(literal(NaN), [NaN]);
});
```

It fails on `main` (`literal(NaN)` returns an issue) and passes with the fix. Full `library/` suite stays green: 272 files / 3057 tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `literal` schema matching for special numeric values: `literal(NaN)` now correctly matches datasets containing `NaN`, and signed zero is handled so `-0` and `+0` are treated as equivalent.
* **Tests**
  * Added coverage to confirm datasets with `NaN` produce no schema issues when validating against a matching `literal(NaN)` schema.


spokodev:
Quick clarification on the automated note above: this uses **SameValueZero**, not `Object.is` —

```js
dataset.value === this.literal ||
(dataset.value !== dataset.value && this.literal !== this.literal)
```

So `-0` and `+0` still compare as equal (no signed-zero behaviour change) and it stays consistent with `picklist`; the only new behaviour is `NaN` matching `NaN`, which is the fix. The flagged signed-zero concern therefore doesn't apply.

yslpn:
@spokodev Thanks for your input. Please use Prettier for your changes

I believe the fix is ​​correct; we'll discuss it soon wit Fabian

spokodev:
Ran Prettier — only a stray blank line in the condition changed. No rush on the review.

yslpn (dismissed):

PR #1534, opened 2026-07-08
Improve perf by changing ~standard implementation

fabian-hiller:
Improve schema init performance: eager `~standard` assignment

Replaces the lazy `get '~standard'()` accessor on every schema factory with eager assignment of the Standard Schema props as a plain data property (`_addStandardProp`), removing `_getStandardProps` and its `_standardCache` WeakMap.

**Why:** the getter made schema *construction* slow on every build — an accessor in an object literal disables V8's fast-literal path, and `pipe()`'s spread of `pipe[0]` invoked the getter (plus a WeakMap get/set) on each call, all to produce a value only read when someone actually uses Standard Schema. Allocating the props eagerly is cheaper than installing an accessor.

**Changes:**
- New `_addStandardProp` util; `_getStandardProps` + WeakMap removed.
- 71 schema factories wrap their literal in `_addStandardProp(...)`.
- 14 spread-based methods (`pipe`, `omit`, `partial`, `fallback`, …) re-wrap their result — **required for correctness**, since a spread copies the source's `validate` binding. Regression tests added (e.g. `pipe(string(), maxLength(3))['~standard'].validate('too long')` must fail).

**Compat:** no public API change; `~standard` stays an own enumerable property (now writable data instead of a getter).

**Results:** ~3× faster construction locally (~6.8 → ~2.1 µs); bundle 85.32 → 82.80 kB. Lint + full suite (3065 tests) pass.

## Summary by CodeRabbit

* **New Features**
  * Standard Schema validation is now consistently available across built-in schemas and schema modifiers.
  * Validation metadata is initialized eagerly and remains stable when accessed repeatedly.
  * Standard validation correctly reflects pipelines, transformations, partial schemas, omitted fields, and required fields.

* **Bug Fixes**
  * Improved validation results and issue reporting for composed and transformed schemas.

* **Documentation**
  * Updated advanced guides with current Standard Schema integration examples.


library/src/utils/_standardSchema/_standardSchema.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @internal
+ */
+export function _standardSchema<
+  TSchema extends
+    | BaseSchema<unknown, unknown, BaseIssue<unknown>>
+    | BaseSchemaAsync<unknown, unknown, BaseIssue<unknown>>,
+>(schema: Omit<TSchema, '~standard'>): TSchema {
+  // @ts-expect-error
+  schema['~standard'] = {
```
yslpn, line 26: Hi Fabian. Good optimization.

But we have schemes that never use `~standard` at all. For example, `parse`/`saveParse`, which is where they start to pay this price.

It seems we can get the best of both worlds. Use the classic pattern: self-overwriting lazy getter. Then scheme creation is zero cost, and the first call is slightly slower (in some cases with ~standart), while all next calls are fast.

Let's try something like this, it should work.

```ts
export function _standardSchema<
  TSchema extends
    | BaseSchema<unknown, unknown, BaseIssue<unknown>>
    | BaseSchemaAsync<unknown, unknown, BaseIssue<unknown>>,
>(schema: Omit<TSchema, '~standard'>): TSchema {
  let cached: StandardProps<unknown, unknown>;
  Object.defineProperty(schema, '~standard', {
    get() {
      return (cached ??= {
        version: 1,
        vendor: 'valibot',
        validate: (value: unknown) =>
          schema['~run']({ value }, getGlobalConfig()) as
            | StandardResult<unknown>
            | Promise<StandardResult<unknown>>,
      });
    },
    enumerable: true,
    configurable: true,
  });
  return schema as TSchema;
}
```

yslpn: I noticed that this won't work lazily. Because when creating the schema, we used a spread operator, which immediately triggers the getter. Making this work would require refactoring, which would slow down execution even more. So it's not worth it. Let's forget about it.

library/src/schemas/array/array.ts
```diff
@@ -76,18 +76,27 @@ export function array(
   BaseSchema<unknown, unknown, BaseIssue<unknown>>,
   ErrorMessage<ArrayIssue> | undefined
 > {
-  return {
+  return _standardSchema<
+    ArraySchema<
+      BaseSchema<unknown, unknown, BaseIssue<unknown>>,
+      ErrorMessage<ArrayIssue> | undefined
+    >
+  >({
```
yslpn: Redundant explicit generic argument `_standardSchema<...>(...)`, repeated in ~90 places and duplicating the function's return type annotation.

library/src/methods/partial/partial.test.ts
```diff
@@ -171,6 +171,15 @@ describe('partial', () => {
         }
       });
     });
+
+    test('should validate against the partialed entries via "~standard"', () => {
```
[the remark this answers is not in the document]
fabian-hiller: Yes, we need to fix this

library/src/methods/pipe/pipeAsync.ts
```diff
@@ -3118,5 +3118,8 @@ export function pipeAsync<
       // @ts-expect-error
       return dataset as OutputDataset<unknown, BaseIssue<unknown>>;
     },
-  };
+  } as unknown as Omit<
+    SchemaWithPipeAsync<readonly [TSchema, ...TItems]>,
+    '~standard'
+  >);
```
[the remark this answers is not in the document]
fabian-hiller: Why is the type cast needed? It was not needed before. We need to check why the cast was added and if there is a better way to implement it.

library/src/methods/fallback/fallbackAsync.ts
```diff
@@ -89,21 +89,28 @@ export function fallbackAsync<
...
+              config
+            ),
           }
         : outputDataset;
     },
-  };
+  } as unknown as Omit<
+    SchemaWithFallbackAsync<TSchema, TFallback>,
+    '~standard'
+  >);
```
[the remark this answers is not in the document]
fabian-hiller: Same here

library/src/schemas/array/array.ts
fabian-hiller: Yes, we should remove it if it is not needed.

fabian-hiller: Done! Thanks!

PR #1536, opened 2026-07-09
fix(object): use own-property check so prototype keys aren't matched as entries (#1523)

sanjibani:
Fixes #1523.

## Bug

\`strictObject\`, \`looseObject\`, and \`objectWithRest\` (and their async variants) checked whether an input key was a defined schema entry with \`key in this.entries\`. The \`in\` operator also matches inherited \`Object.prototype\` members, so an input key named like a prototype member was treated as a defined entry.

Effect:

- \`strictObject\` / \`strictObjectAsync\` failed to reject \`__proto__\`, \`constructor\`, \`toString\`, \`valueOf\`, and friends as unknown keys. A \`__proto__\` or \`constructor\` key passed a strict schema silently.
- \`looseObject\` / \`looseObjectAsync\` dropped such keys instead of passing them through.
- \`objectWithRest\` / \`objectWithRestAsync\` skipped such keys instead of validating them against the \`rest\` schema.

Before the fix:

\`\`\`ts
v.parse(v.strictObject({ name: v.string() }), { name: 'a', toString: 'b' });
// returns { name: 'a' }; the unknown \`toString\` key is not reported
\`\`\`

## Fix

Replace each \`key in this.entries\` membership check with a new \`_hasOwnProperty(this.entries, key)\` helper - the same \`Object.prototype.hasOwnProperty.call\` pattern already used by \`_isValidObjectKey\` for input keys and by the #1522 fixes. A schema that intentionally defines an entry named like a prototype member (\`v.object({ toString: v.string() })\`) still works, since the own-property check is precise, not name-based.

The new helper lives at \`library/src/utils/_hasOwnProperty/\` and is re-exported from the utils barrel alongside the existing \`_isValidObjectKey\` so the two prototype-aware checks sit side by side.

## Tests

- New \`_hasOwnProperty\` unit suite (3 tests) covering own properties, inherited \`Object.prototype\` members, and own properties that share a name with a prototype member.
- New \`strictObject\` regression test asserting that an input carrying \`toString\` / \`constructor\` / \`__proto__\` is now rejected as an unknown key. The pre-fix code returned a clean parse for these inputs.

8 files changed, +96/-8.

## Summary by CodeRabbit

* **New Features**
  * Added an internal own-property utility and re-exported it for shared schema use.

* **Bug Fixes**
  * Improved object key handling to treat only direct (own) properties as defined, avoiding prototype-inherited keys being accepted as valid.
  * Tightened strict object validation so prototype-collision keys are correctly rejected as unknown.

* **Tests**
  * Added regression tests covering own-property behavior and strict/edge-case prototype-colliding keys.

### DIFF
--- a/library/src/schemas/looseObject/looseObject.ts
+++ b/library/src/schemas/looseObject/looseObject.ts
@@ -12,6 +12,7 @@ import type {
 import {
   _addIssue,
   _getStandardProps,
+  _hasOwnProperty,
   _isValidObjectKey,
 } from '../../utils/index.ts';
 import type { LooseObjectIssue } from './types.ts';
@@ -205,7 +206,7 @@ export function looseObject(
         // Hint: We exclude specific keys for security reasons
         if (!dataset.issues || !config.abortEarly) {
           for (const key in input) {
-            if (_isValidObjectKey(input, key) && !(key in this.entries)) {
+            if (_isValidObjectKey(input, key) && !_hasOwnProperty(this.entries, key)) {
               // @ts-expect-error
               dataset.value[key] = input[key];
             }
--- a/library/src/schemas/looseObject/looseObjectAsync.ts
+++ b/library/src/schemas/looseObject/looseObjectAsync.ts
@@ -12,6 +12,7 @@ import type {
 import {
   _addIssue,
   _getStandardProps,
+  _hasOwnProperty,
   _isValidObjectKey,
 } from '../../utils/index.ts';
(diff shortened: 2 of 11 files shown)


fabian-hiller:
This addresses the same schema-entry lookup bug as #1523. We are continuing with #1523 for v1.5, using inline own-property checks, regression coverage across all six affected implementations, and a changelog entry. The fix preserves existing inherited-input behavior and does not introduce an additional helper. Thanks for contributing!

PR #1542, opened 2026-07-16
fix(to-valibot): support z.brand() in codemod (#1501)

### LINKED ISSUE #1501: Codemod breaks z.brand
```ts
import { z } from "zod/v4";

export const StationNumberSchema = z.string().brand("StationNumber");
```

The result is

```ts
import * as v from "valibot";

export const StationNumberSchema = v.string()("StationNumber");
```

The expected result is 

```ts
import * as v from "valibot";

const StationNumberSchema = v.pipe(v.string(), v.brand("StationNumber"));
```

```
> pnpx @valibot/zod-to-valibot --version
jscodeshift: 17.3.0
 - babel: 7.29.7
 - babylon: 7.29.7
 - flow: 0.318.0
 - recast: 0.23.11
```

xianjianlf2:
## Overview

The `@valibot/zod-to-valibot` codemod had no handler for Zod's `.brand()` method. Because the method went unrecognized, it was left in place as a call on the transformed schema, producing invalid output like `v.string()("Name")` (a schema object called as a function).

Fixes #1501

## Fix

Added a `brand` method handler that maps `schema.brand(name)` into a pipe action, mirroring how the existing `transform` method is handled. It uses the shared `addToPipe` helper so the brand is appended to any existing pipe:

```
z.string().brand("StationNumber")   ->   v.pipe(v.string(), v.brand("StationNumber"))
```

The handler is wired into `ZOD_METHODS` (constants), `methods/index.ts`, and the schemas-and-links dispatcher.

## Tests

Added a `brand` test fixture (`__testfixtures__/brand/`) covering a basic brand, a brand after a validator chain, a brand on a number schema, and a brand on a linked schema, and registered it in `test-setup.test.ts`. The full codemod suite passes (72/72).

## Summary by CodeRabbit

* **New Features**
  * Added support for converting Zod branded schemas into equivalent Valibot branded schemas.
  * Supports brands on strings, UUID-validated values, constrained numbers, linked schemas, and type-only string literals.
  * Brand names can be inferred from generic type parameters when explicit arguments are not provided.
  * Branding is preserved through existing schema validation pipelines.
* **Tests**
  * Added coverage for multiple branded schema patterns and included the new brand conversion scenario in the transformation test suite.


xianjianlf2:
Addressed the review feedback and rebased the branch onto the latest `main` in `b8efed5`.

Changes:
- Added `.ts` extensions for the new brand transformer exports/imports. In `methods/index.ts` this has to be `./brand/index.ts` because `brand` is a directory, not a sibling `brand.ts` file.
- Added JSDoc and `// @__NO_SIDE_EFFECTS__` for `transformBrand`.
- Preserved Zod's type-only string-literal brand form, so `z.string().brand<"UserId">()` now transforms to `v.pipe(v.string(), v.brand("UserId"))` instead of `v.brand()`.
- Added a brand fixture for that type-only case.

Local checks:
- `pnpm --filter @valibot/zod-to-valibot test` (72 tests passed)
- `pnpm --filter @valibot/zod-to-valibot build`
- `git diff --check`

PR #1550, opened 2026-07-19
Add MCP server to valibot.dev

fabian-hiller:
## Summary

Adds a Model Context Protocol server at `https://valibot.dev/mcp` so coding agents can search and read the documentation through first-class tools instead of crawling pages. This addresses the last substantive gap in the [Mintlify agent score](https://www.mintlify.com/score/valibot) and [isitagentready](https://isitagentready.com/valibot.dev) audits.

### MCP server (`website/worker/mcp.ts`)

- Stateless Streamable HTTP transport with plain JSON responses — no sessions, no auth, no external dependencies. Hand-rolled JSON-RPC layer (~600 lines) covering `initialize`, `ping`, `tools/list`, `tools/call`, notifications, batches, CORS for browser-based clients, and protocol version negotiation (2025-06-18 / 2025-03-26).
- Three tools: `search_docs` (ranked search over all 794 pages with API-name-first scoring), `get_doc` (full page Markdown; accepts `api/string`, full URLs, or bare names like `minLength`, with typo suggestions on misses), and `list_docs` (grouped overview).
- All tool inputs are validated with Valibot itself, and the advertised JSON Schemas are generated from those same schemas via `@valibot/to-json-schema` (`ignoreActions: ['trim']`) — validation and declaration cannot drift.
- Search runs against a build-time index (`search-index.json`, generated in `scripts/llms.ts` from the same source as the `.md` mirrors) cached in isolate memory with precomputed lowercase fields. Deliberately not Algolia: results are always consistent with the deployed `.md` files, and the endpoint has no third-party runtime dependency.

### Discovery

- MCP server card at `/.well-known/mcp/server-card.json` (SEP-1649), generated at build time from constants shared with the Worker (`worker/mcp-info.ts`) so the card and the live handshake cannot diverge.
- The server is listed in the API catalog linkset, `llms.txt`, and the visually hidden agent hint on every HTML page.

### Guide rename

- `guides/llms-txt` → `guides/coding-agents`: a general page covering the MCP server (with Claude Code and Cursor setup snippets), the LLMs.txt routes, Markdown content negotiation, and the agent skill. Moved after the installation guide in the menu. Old URLs (HTML and `.md`) redirect with 301.

### Structure and robustness

- The Worker is now split into focused modules: `index.ts` (routing), `markdown.ts` (content negotiation), `headers.ts` (merge-aware header utilities), `docs.ts` (shared area/slug knowledge and the `SearchEntry` type used by both the Worker and the generator), `mcp-info.ts` (shared MCP constants).
- Review findings fixed along the way: 304 revalidation responses keep their negotiation headers, JSON-RPC edge cases (`id: null` requests, empty batches, id preservation in batch errors), `/mcp/` trailing-slash support, `#anchor`/`?query` tolerance in `get_doc` paths, whitespace-only queries rejected via `v.trim()`, `$schema` stripped from tool schemas, and the typo-suggestion scan capped (`maxLength` on inputs plus a prefix ceiling).

## Test plan

- `pnpm lint`, `pnpm format.check`, and `pnpm build` pass
- JSON-RPC test suite against `wrangler dev`: handshake, all three tools, invalid arguments, unknown tool/method, parse errors, empty batch, `id: null`, trailing slash, CORS preflight, 405 on GET
- End-to-end session with the official `@modelcontextprotocol/sdk` client: connect, list tools, call all three tools
- Markdown negotiation, `.md` caching headers (ETag/304), API catalog, and redirect regressions all re-verified

## Summary by CodeRabbit

* **New Features**
  * Added an MCP endpoint (`/mcp`) for searching, listing, and retrieving documentation, including agent discovery server-card data.
  * Generated a `search-index.json` powering documentation search and lookups.
  * Added the new “Coding agents” guide with setup instructions and Markdown-accessible documentation routes.
  * Published MCP discovery metadata under `.well-known` and added an entry to the API catalog.
* **Documentation**
  * Updated docs navigation/help text and redirected legacy `llms-txt` URLs to the “Coding agents” guide (including `.md` variants).
  * Removed the standalone `llms-txt` guide route.


website/worker/mcp-info.ts
```diff
@@ -0,0 +1,24 @@
...
+// Info about our MCP server shared with clients and our MCP server card
+export const SERVER_INFO = {
+  name: 'valibot-docs',
+  title: 'Valibot Documentation',
+  version: '1.0.0',
+} as const;
+
+// Instructions to help clients use our MCP server effectively
+export const INSTRUCTIONS =
+  'Valibot is the modular and type safe schema library for validating structural data. Its API is divided between schemas (data types like v.string and v.object), methods (operate on schemas like v.parse and v.pipe) and actions (validations and transformations like v.email and v.minLength that are used inside v.pipe). Import everything with `import * as v from \'valibot\'`. Do not confuse Valibot with Zod: Valibot uses functions and pipelines instead of chained methods. Use search_docs to find relevant pages, get_doc to read a page and list_docs for a complete overview. For migration from Zod, read the page at path "guides/migrate-from-zod".';
```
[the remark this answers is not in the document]
fabian-hiller: Fixed in eedf6ce1. Repo usage was checked first: the hyphenated form is the majority (14 vs 7) and matches the published SKILL.md, so the new MCP strings now use "type-safe". The sitewide tagline still uses the unhyphenated form and is out of scope for this PR.

website/scripts/agents.ts
```diff
@@ -70,3 +76,32 @@ fs.writeFileSync(
...
+    {
+      serverInfo: SERVER_INFO,
+      protocolVersion: LATEST_VERSION,
+      transport: {
+        type: 'streamable-http',
+        endpoint: 'https://valibot.dev/mcp',
+      },
+      capabilities: CAPABILITIES,
+      description:
+        'Search and read the documentation of Valibot, the modular and type safe schema library for validating structural data.',
```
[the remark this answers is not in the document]
fabian-hiller: Fixed in eedf6ce1 together with the instructions string.

website/public/_redirects
```diff
@@ -6,3 +6,6 @@
 /guides/errors/ /guides/issues/ 302
 /guides/migrate-from-v0.30.0 /guides/migrate-to-v0.31.0/ 301
 /guides/migrate-from-v0.30.0/ /guides/migrate-to-v0.31.0/ 301
+/guides/llms-txt /guides/coding-agents/ 301
+/guides/llms-txt/ /guides/coding-agents/ 301
+/guides/llms-txt.md /guides/coding-agents.md 301
```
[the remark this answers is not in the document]
fabian-hiller: Fixed in eedf6ce1. Verified that vercel.json actively maintains a redirects array (it already contains the migrate-from-v0.30.0 rename in the same pattern) and added the three llms-txt entries (/guides/llms-txt, /guides/llms-txt/, /guides/llms-txt.md).

website/scripts/agents.ts
```diff
@@ -70,3 +76,32 @@ fs.writeFileSync(
...
+  fs.mkdirSync(mcpDir, { recursive: true });
+}
+
+// Write MCP server card to public directory for agent discovery (SEP-1649)
+// https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127
+fs.writeFileSync(
+  path.join(mcpDir, 'server-card.json'),
+  `${JSON.stringify(
+    {
+      serverInfo: SERVER_INFO,
```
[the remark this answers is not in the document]
fabian-hiller: Partially valid, fixed as a superset in eedf6ce1. Verified against both consumers: the SEP-1649 proposal moved to the Extensions track with the schema still in flux, and the latest draft removed the $schema field this comment asks for. The card now carries the draft-shape fields (name, version, description, websiteUrl, repository, remotes with supportedProtocolVersions) in addition to the original serverInfo/transport/capabilities shape, which is exactly what current discovery checkers validate. JSON consumers ignore unknown fields, so both shapes are served; the card will be trimmed once the extension spec stabilizes.

website/public/_redirects
```diff
@@ -6,3 +6,6 @@
 /guides/errors/ /guides/issues/ 302
 /guides/migrate-from-v0.30.0 /guides/migrate-to-v0.31.0/ 301
 /guides/migrate-from-v0.30.0/ /guides/migrate-to-v0.31.0/ 301
+/guides/llms-txt /guides/coding-agents/ 301
```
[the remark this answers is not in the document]
fabian-hiller: Fixed in eedf6ce1 — see the identical thread from Codex; the three llms-txt redirects were added to vercel.json.

website/worker/markdown.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Returns the quality value of a media type within an `Accept` header.
+ *
+ * @param accept The value of the `Accept` header.
+ * @param mediaType The media type to look for.
+ *
+ * @returns The quality value of the media type.
+ */
+function getQuality(accept: string, mediaType: string): number {
```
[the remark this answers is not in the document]
fabian-hiller: Skipped, this is intentional. Counting text/* or */* as a match for Markdown would flip clients that send only wildcards — most notably curl's default Accept: */* — from HTML to Markdown. HTML is the default representation of these URLs, so Markdown is only served on an explicit text/markdown entry with a quality at or above text/html. All realistic client and agent headers were tested and negotiate correctly.

website/worker/mcp.ts
```diff
@@ -0,0 +1,610 @@
...
+      typeof message !== 'object' ||
+      message === null ||
+      (message as JsonRpcMessage).jsonrpc !== '2.0' ||
+      typeof (message as JsonRpcMessage).method !== 'string'
+    ) {
+      // Preserve ID of invalid message if available so that the client can
+      // correlate the error with its request
+      const errorId =
+        typeof message === 'object' && message !== null && 'id' in message
+          ? (message as JsonRpcMessage).id
```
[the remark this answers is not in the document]
fabian-hiller: Fixed in eedf6ce1, and the verified hole was slightly larger than described: IDs of an invalid type also flowed into successful responses, not just error responses. The structural check now validates the ID type; string and number IDs are preserved, anything else gets id: null. Verified at runtime with boolean and object IDs plus an ID-preservation regression test.

website/worker/mcp.ts
```diff
@@ -0,0 +1,610 @@
+/**
```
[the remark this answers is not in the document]
fabian-hiller: Fixed in eedf6ce1. Verified against the specification, which indeed requires a 400 response for an invalid or unsupported MCP-Protocol-Version ("it MUST respond with 400 Bad Request"). Requests with an unsupported version now receive 400; a missing header still proceeds per the spec's backwards-compatibility rule. Verified at runtime and with the official MCP SDK client, which sends the header on subsequent requests.

website/worker/mcp.ts
```diff
@@ -0,0 +1,610 @@
...
+    return jsonResponse(createError(null, -32700, 'Parse error'), 400);
+  }
+
+  // Reject empty batches as required by the JSON-RPC specification
+  if (Array.isArray(body) && !body.length) {
+    return jsonResponse(createError(null, -32600, 'Invalid request'), 400);
+  }
+
+  // Process single message or batch of messages
+  const messages = Array.isArray(body) ? body : [body];
```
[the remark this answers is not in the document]
fabian-hiller: Skipped. The single-message-per-POST rule is 2025-06-18 text, but this server also negotiates protocol version 2025-03-26, where JSON-RPC batching was part of the Streamable HTTP transport. Rejecting arrays would break spec-compliant 2025-03-26 clients, while 2025-06-18 clients never send arrays and therefore never receive a batch-shaped response.

website/worker/mcp-info.ts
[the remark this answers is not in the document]
fabian-hiller: Fixed in eedf6ce1 — see the identical thread above.

website/worker/mcp.ts
```diff
@@ -0,0 +1,634 @@
...
+ *
+ * @param schema The Valibot schema to convert.
+ *
+ * @returns A JSON Schema object.
+ */
+function toInputSchema(schema: v.GenericSchema): Record<string, unknown> {
+  // Ignore the `trim` transformation as it cannot be converted to JSON
+  // Schema but still runs when the arguments are validated with Valibot
+  const jsonSchema: Record<string, unknown> = {
+    ...toJsonSchema(schema, { ignoreActions: ['trim'] }),
```
yslpn: This is a dangerous place. If, for example, somobody add `tolowercase` in the future and we forget to add it here, the entire site will crash. If I understand correctly.

Maybe add a test to ensure `toInputSchema` won't fail for all schema types. This would be caught during CI/CD rather than after deployment.

Or, more simply, add `errorMode: 'warn'` to `toJsonSchema`

website/worker/mcp.ts
```diff
@@ -0,0 +1,634 @@
...
+  ) {
+    suggestions = searchDocs(index, { query: name.slice(0, length), limit: 3 })
+      .map(({ entry }) => `${entry.area}/${entry.name}`)
+      .join('", "');
+  }
+  return {
+    content: [
+      {
+        type: 'text',
+        text: `No documentation page found for "${input.path}".${suggestions ? ` Did you mean: "${suggestions}"?` : ' Use search_docs or list_docs to find available pages.'}`,
```
yslpn, line 376: The "did you mean" loop doesn't catch typos except at the end of a word, and it's very computationally expensive.

A linear loop through the entire index, with several .includes() checks per entry. In the worst case (a long name with no matches at all), the loop performs up to ~16,000 comparisons per failed query, which ultimately won't find anything for a typical non-terminal typo.

The solution is either to add real fuzzy matching or to do something like `searchDocs(index, { query: name, limit: 3 })`

website/worker/markdown.ts
```diff
@@ -0,0 +1,108 @@
...
+    return withHeaders(response, {
+      'Content-Location': path,
+      Vary: 'Accept',
+    });
+  }
+
+  // Pass redirects (e.g. from our `_redirects` file) through to the client
+  // so that it can request the Markdown file at its new location
+  if (response.status >= 300 && response.status < 400) {
+    return response;
```
yslpn: A cache that stores this 301 at a URL without taking Accept into account may later issue a "markdown" redirect to a regular HTML browser, or vice versa.

```ts
return withHeaders(response, { Vary: 'Accept' });
```

website/vercel.json
yslpn: lost `npm run agents`?

or better create new npm script `prebuild`/`generate` and reuse it

website/worker/mcp.ts
```diff
@@ -0,0 +1,634 @@
...
+          type: 'text',
+          text: `Invalid path "${input.path}". Use a path like "api/string" or "guides/quick-start" or a bare API name like "minLength".`,
+        },
+      ],
+      isError: true,
+    };
+  }
+
+  // Return content of first Markdown file that exists
+  for (const filePath of filePaths) {
```
yslpn: serial fetch instead of parallel

```ts
const responses = await Promise.all(
  filePaths.map((filePath) =>
    env.ASSETS.fetch(new URL(filePath, requestUrl).href)
  )
);
for (let i = 0; i < responses.length; i++) {
  if (responses[i].ok) {
    return { content: [{ type: 'text', text: await responses[i].text() }] };
  }
}
```

website/scripts/agents.ts
```diff
@@ -70,3 +77,49 @@ fs.writeFileSync(
...
+      version: SERVER_INFO.version,
+      websiteUrl: 'https://valibot.dev/',
+      repository: {
+        url: 'https://github.com/open-circle/valibot',
+        source: 'github',
+      },
+      remotes: [
+        {
+          type: 'streamable-http',
+          url: 'https://valibot.dev/mcp',
```
yslpn: maybe you can add this

```ts
export const MCP_URL = 'https://valibot.dev/mcp';
export const TRANSPORT_TYPE = 'streamable-http';
```

to `website/worker/mcp-info.ts`

and reuse it

website/worker/mcp.ts
fabian-hiller: Agree. I will change it to warn.

website/worker/mcp.ts
fabian-hiller: Thank you! I will look into it.

website/vercel.json
fabian-hiller: I plan to delete the Vercel config. But agree we should sync it until we actually delete it.

website/worker/mcp.ts
```diff
@@ -0,0 +1,634 @@
...
+ * @param input The validated tool input.
+ *
+ * @returns The scored search results.
+ */
+function searchDocs(
+  index: IndexedEntry[],
+  input: v.InferOutput<typeof SearchDocsSchema>
+): { entry: IndexedEntry; score: number }[] {
+  // Split query into lowercase search terms
+  const terms = input.query.toLowerCase().split(/\s+/).filter(Boolean);
```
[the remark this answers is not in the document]
fabian-hiller: Fixed in fbc7a737. Search terms now strip a leading `v.` prefix, so queries like `v.string` and `v.minLength` match the bare API names in the index.

website/worker/mcp.ts
fabian-hiller: Fixed in fbc7a737 with `errorMode: 'warn'`. `ignoreActions: ['trim']` is kept so the known case stays silent while an unknown future action logs a warning instead of crashing the Worker at startup.

website/worker/mcp.ts
fabian-hiller: Fixed in fbc7a737. The prefix loop is replaced with a single pass that computes the Levenshtein distance against all page names (with a length pre-filter), so mid-word typos like "mnLength" now get suggestions and the worst case is one pass over the index.

website/worker/markdown.ts
fabian-hiller: Fixed in fbc7a737 exactly as suggested. The redirect target differs between the Markdown and HTML representation, so caches must vary on `Accept`.

website/vercel.json
fabian-hiller: Fixed in fbc7a737. `npm run agents` was added to the build command in the same position as in `build.cloudflare` to keep both in sync until the Vercel config is removed.

website/worker/mcp.ts
fabian-hiller: Fixed in fbc7a737. The candidate files are now fetched with `Promise.all` while the array order still decides the area priority.

website/scripts/agents.ts
fabian-hiller: Fixed in fbc7a737. `MCP_URL` and `TRANSPORT_TYPE` now live in `mcp-info.ts` and are reused in `agents.ts`; the generated server card is byte-identical.

website/worker/mcp.ts
```diff
@@ -0,0 +1,693 @@
...
+  // the API area first as most bare lookups are API references.
+  let areas: DocArea[];
+  let pageName: string;
+  const areaMatch = AREA_PATH_REGEX.exec(cleanPath);
+  if (areaMatch) {
+    areas = [areaMatch[1] as DocArea];
+    pageName = areaMatch[2];
+  } else if (/^[\w.-]+$/.test(cleanPath)) {
+    areas = ['api', 'guides'];
+    pageName = cleanPath;
```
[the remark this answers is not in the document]
fabian-hiller: Fixed in 659cb679. The `v.` prefix is now stripped case-insensitively from the page name after the area split, so `v.string`, `api/v.minLength` and even `V.MINLENGTH` (combined with the case-insensitive resolution) return the page directly.

PR #1562, opened 2026-07-22
Fix critical tar vulnerabilities via pnpm overrides

fabian-hiller:
## Summary

The dependency scan reported two critical findings, both for the transitive `tar` package:

| Finding | Path | Status |
| --- | --- | --- |
| `tar@6.2.1` | `vercel` → `@vercel/fun` | Fixed → `7.5.21` |
| `tar@7.5.2` | `vercel` → `…` → `@mapbox/node-pre-gyp` | Fixed → `7.5.21` |

All known tar advisories (11 in OSV, including the critical decompression DoS GHSA-23hp-3jrh-7fpw) are only fixed in **tar 7.5.19+** — there is no patched 6.x release. Both major lines are therefore pinned to `tar@7.5.21` via `overrides` in `pnpm-workspace.yaml`.

## Why overrides instead of upgrading vercel

The scanner suggested upgrading `vercel` to 50.3.1, but that was verified to be insufficient:

- `vercel@50.3.1` still ships `@vercel/fun@1.2.1`, which pins `tar@6.2.1` exactly.
- Even the latest `vercel@56.5.0` ships `@vercel/fun@1.3.0`, which pins `tar@7.5.7` — still below the 7.5.19 fix threshold. No vercel release resolves tar to a patched version.
- Additionally, every vercel major ≥ 51 depends on the npm package `sandbox` (Vercel's Sandbox CLI, which reuses an old npm package name). Two 2020 advisories for the *original* `sandbox` package (GHSA-fm4j-4xhm-xpwx, GHSA-gc25-3vc5-2jf9) have an unbounded affected range, so scanners flag Vercel's unrelated 3.x code as a critical false positive with no fixed version and no way to suppress it. Their metadata (`last_known_affected_version_range: < 1.0.0`) confirms only the old package is affected. Staying on `vercel@^48.10.10` avoids introducing that permanent false positive into the Security workflow.

## Safety of the tar 6 → 7 override

`@vercel/fun` is the only consumer of the tar 6 line. Its usage is limited to `tar.extract({ strip, C })`, which is unchanged in tar 7 — confirmed by diffing `@vercel/fun` 1.2.0 (tar 6) against 1.3.0 (tar 7): every tar call site is identical.

## Verification

- `pnpm why -r tar` shows a single `tar@7.5.21` across both paths; `tar@6.2.1` and `tar@7.5.2` are gone from the lockfile.
- Rescan (CVE Lite CLI 1.27.0, same as the Security workflow) reports **0 critical** findings and exits 0 with `--fail-on critical`.
- `vercel --version` (48.10.10) runs correctly with the overridden tar.

Out of scope: the remaining high/medium/low findings (e.g. `sharp`, `undici`, `vite`) are untouched, per the request to fix criticals only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated underlying archive handling to address security vulnerabilities and improve application reliability.
  * No user-facing functionality or interface changes.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,6 +6,8 @@ settings:
 
 overrides:
   shell-quote@1: 1.8.4
+  tar@6: 7.5.21
+  tar@7: 7.5.21
 
 patchedDependencies:
   '@builder.io/qwik-city@1.19.2': 2195348d10017bf69653482d128f793158a82c1d0eeb0e3fd909db4c503493a1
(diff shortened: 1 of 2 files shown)


yslpn:
@fabian-hiller сan we get rid of vercel if we use cloudflare?

fabian-hiller:
@yslpn yes, feel free to prepare a PR

PR #1569, opened 2026-07-29
chore: reschedule and add npm target

ysknsid25:
rescheduled depenadbot interval monthly to weekly.
It is best to update library versions frequently.
I’ll be reviewing them on my own initiative every week.

I have also included npm packages in the scope for Dependabot.
I will be checking these on a weekly basis as well.

Regarding major version updates, automatic updates are too risky, so we have decided to verify them ourselves.

## Summary by CodeRabbit

* **Chores**
  * Automated dependency updates are now scheduled weekly for both npm packages and GitHub Actions (with GitHub Actions set to a Saturday time).
  * npm updates are scoped to key project directories, grouped to limit open pull requests, and use a longer default cooldown with a `dependabot` commit prefix.
  * Major semver updates for general dependencies (`*`) are ignored for safer automated rollouts.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,6 +20,10 @@ updates:
       npm-deps:
         patterns:
           - "*"
+    ignore:
+      - dependency-name: "*"
+        update-types:
+          - "version-update:semver-major"
   - package-ecosystem: github-actions
     directory: /
     schedule:
@@ -34,4 +38,8 @@ updates:
       github-actions:
         patterns:
           - '*'
+    ignore:
+      - dependency-name: "*"
+        update-types:
+          - "version-update:semver-major"
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,6 +18,9 @@ updates:
       prefix: "dependabot"
     groups:
       npm-deps:
+        update-types:
+          - "patch"
+          - "minor"
         patterns:
           - "*"
     ignore:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,20 +9,15 @@ updates:
       - "/codemod/*"
     schedule:
       interval: weekly
-      day: "saturday"
-      time: "09:00"
     open-pull-requests-limit: 5
     cooldown:
       default-days: 7
     commit-message:
       prefix: "dependabot"
     groups:
-      patch-updates:
-        update-types:
-          - "patch"
-      minor-updates:
-        update-types:
-          - "minor"
+      npm-deps:
+        patterns:
+          - "*"
   - package-ecosystem: github-actions
     directory: /
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,8 @@ updates:
       - "/codemod/*"
     schedule:
       interval: weekly
+      day: "saturday"
+      time: "09:00"
     open-pull-requests-limit: 5
     cooldown:
       default-days: 7
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,12 @@
 version: 2
 updates:
   - package-ecosystem: npm
-    directory: /
+    directories:
+      - "/"
+      - "/library"
+      - "/packages/*"
+      - "/website"
+      - "/codemod/*"
     schedule:
       interval: weekly
       day: "saturday"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -39,6 +39,9 @@ updates:
       prefix: "dependabot"
     groups:
       github-actions:
+        update-types:
+          - "patch"
+          - "minor"
         patterns:
           - '*'
     ignore:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,29 @@
 version: 2
 updates:
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: weekly
+      day: "saturday"
+      time: "09:00"
+    open-pull-requests-limit: 5
+    cooldown:
+      default-days: 7
+    commit-message:
+      prefix: "dependabot"
+    groups:
+      patch-updates:
+        update-types:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,9 +1,29 @@
 version: 2
 updates:
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: weekly
+      day: "saturday"
+      time: "09:00"
```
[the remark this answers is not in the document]
ysknsid25: That is not a problem. Libraries whose cooldown period has expired become eligible for updates every other week. It is better to perform version upgrades frequently.

.github/dependabot.yml
```diff
@@ -1,9 +1,34 @@
...
+      - "/"
+      - "/library"
+      - "/packages/*"
+      - "/website"
+      - "/codemod/*"
+    schedule:
+      interval: weekly
+      day: "saturday"
+      time: "09:00"
+    open-pull-requests-limit: 5
```
yslpn, line 14: I propose grouping all updates into two PRs:

one
`github-actions`

the other
`npm-deps` or another name

This will make it easier to review and test.

.github/dependabot.yml
```diff
@@ -1,9 +1,34 @@
...
+    directories:
+      - "/"
+      - "/library"
+      - "/packages/*"
+      - "/website"
+      - "/codemod/*"
+    schedule:
+      interval: weekly
+      day: "saturday"
+      time: "09:00"
```
yslpn, line 13: It's best not to overcomplicate the configuration and not add any specific time. Use the default 

`interval: weekly.`That's enough. 

Then we'll get a PR immediately after merging these changes, without having to wait a week.

.github/dependabot.yml
```diff
@@ -1,9 +1,34 @@
...
+    open-pull-requests-limit: 5
+    cooldown:
+      default-days: 7
+    commit-message:
+      prefix: "dependabot"
+    groups:
+      patch-updates:
+        update-types:
+          - "patch"
+      minor-updates:
```
yslpn: It is not worth splitting into minor updates and patch updates, as these will be conflicting PRs.

.github/dependabot.yml
ysknsid25: Honestly, I feel whether this level of configuration is considered complex depends on the person.

I think it is more unpredictable when we don't know when it'll be generated a PR. Once the schedule is clear, it’s easier to make checking it a regular weekly routine. I think it is easier to plan lunch around a pizza that arrives at a set time than around one that is delivered once a week but at an unpredictable time.

Also, I memory getting a PR right after merging isn't actually related to whether day/time are specified or not. GitHub triggers an immediate check as soon as dependabot.yml itself is changed and merged, regardless of the schedule configuration — that's what gives us the "PR right after merging" behavior, not the absence of day/time.

I didn't quite understand the complexities involved in specifying days and times, or the disadvantages that entails, so I would like you to explain it to me in detail.

.github/dependabot.yml
ysknsid25: Thank you for your comment. I fixed 👍

.github/dependabot.yml
ysknsid25: I fixed too.

.github/dependabot.yml
```diff
@@ -1,15 +1,51 @@
...
+      - "/"
+      - "/library"
+      - "/packages/*"
+      - "/website"
+      - "/codemod/*"
+    schedule:
+      interval: weekly
+      day: "saturday"
+      time: "09:00"
+    open-pull-requests-limit: 5
```
yslpn, line 14: Also, consider removing `open-pull-requests-limit: 5` ; I don't think they'll accumulate. Or set the limit to 1. For github-actions too.

.github/dependabot.yml
yslpn: There's no need to plan anything here. Therefore, I would omit any unnecessary configuration lines that don't add any benefit. This is my personal opinion. If you want, feel free to leave them in.

.github/dependabot.yml
ysknsid25: No, we should not remove `open-pull-requests-limit`. Two things to note here:

- In a multi-directory setup, pull requests corresponding to the number of directories can effectively be open simultaneously.
- [Security update pull requests are not subject to this limit and do not count toward it. There is no limit on the number of open pull requests for security updates.](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#open-pull-requests-limit-:~:text=Security%20update%20pull%20requests%20are%20not%20subject%20to%20this%20limit%20and%20do%20not%20count%20toward%20it.%20There%20is%20no%20limit%20on%20the%20number%20of%20open%20pull%20requests%20for%20security%20updates.)

While there is a question of whether "5" is appropriate, that can be adjusted during actual operation.

yslpn: We have one group for all directories. There will be one PR. Right?

ysknsid25: Yes, plus security update are excluded from limits.

In any case, there is not always just only one.

yslpn: Let's delete this field then.

ysknsid25: I'd rather keep the limit as a guardrail than remove it. I've seen plenty of projects get flooded with Dependabot PRs that end up ignored precisely because there was no cap — removing it because the count is "unpredictable" seems backwards to me; unpredictability is the reason to keep a ceiling, not drop it.

There's no need to generate a massive number of alerts all at once; we can process what comes up on a weekly basis. And security updates are independent of both the schedule and this limit either way:

- High-priority (security) items get a PR the moment they're detected, regardless of the limit.
- Everything else, we work through steadily on the weekly cadence.

 I don't think we need to clutter the PR list with updates we can't act on immediately anyway.

yslpn: okay

yslpn (approved):

PR #1573, opened 2026-08-02
fix: use SameValueZero equality for special numeric values

yslpn:
# Fix SameValueZero equality for special numeric values

## Summary

This change makes value equality consistent across `literal`, `intersect`,
`value`, `notValue`, `values`, and `notValues`.

## Problem

JavaScript strict equality and ordering comparisons do not consider `NaN` equal
to itself. As a result, several APIs rejected a value that was identical to its
configured requirement:

- `literal(NaN)` rejected `NaN`.
- `value(NaN)` and `values([NaN])` rejected `NaN`.
- `notValue(NaN)` and `notValues([NaN])` accepted `NaN`.
- `intersect` reported a conflict when two branches produced `NaN` or invalid
  `Date` values for the same entry.

## Changes

- Adds the internal `_sameValueZero` utility.
- Uses it in `literal` and `_merge`.
- Uses it before the existing ordering fallback in `value`, `notValue`,
  `values`, and `notValues`.
- Preserves the existing behavior for `-0` and `+0`.
- Preserves cross-instance comparison of valid `Date` values in the value
  actions.

## Tests

Adds regression coverage for:

- `NaN` matching itself in every affected API.
- `-0` and `+0` remaining interchangeable.
- Merging two invalid `Date` objects.

## Verification

- `pnpm -C library test` — 520 test files and 4514 tests passed.
- `pnpm -C library lint` — passed without errors.
- `pnpm -C library format.check` — passed.

## Acknowledgements

This consolidated fix builds on the work and investigation in:

Closes #1477
Closes #1517
Closes #1529
Closes #1535

Thank you all for identifying the edge cases, proposing fixes, and adding
regression coverage that informed this unified change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of special numeric values, including `NaN` and positive or negative zero.
  * Corrected literal, value, and collection comparisons for invalid dates and other edge cases.
  * Preserved special values correctly when combining data.

* **Tests**
  * Added coverage for `NaN`, signed zero, invalid dates, and value-matching behavior across validation and comparison scenarios.


fabian-hiller:
- Is this a breaking change for some of the modified functions? Can this be released as part of the next v1 release.
- I recommend `_sameValueZero` follows the `_is` prefix that we use for functions that return a boolean.

library/src/actions/notValue/notValue.ts
```diff
@@ -117,8 +117,9 @@ export function notValue(
     '~run'(dataset, config) {
       if (
         dataset.typed &&
-        this.requirement <= dataset.value &&
-        this.requirement >= dataset.value
+        (_sameValueZero(this.requirement, dataset.value) ||
+          (this.requirement <= dataset.value &&
+            this.requirement >= dataset.value))
```
ysknsid25: 🙏 **This is a matter of preference, so I'll leave the decision up to you** 🙏

Since the comparison `from <= value <= to` appears in multiple places, it might be worth creating a helper function named `between(value, from, to)` to handle this more declaratively.
※The name is just a placeholder.

Additionally, as there seem to be many instances where `_sameValueZero` and `between` can be treated as a single combined condition, it might be best to define a more abstract function—representing `_sameValueZero || between`—and place it in its own file.

ysknsid25 (commented):
Thank you for your work!
I left a comment that was more like a personal impression 🙏

yslpn:
> * Is this a breaking change for some of the modified functions? Can this be released as part of the next v1 release.

I consider it a bug fix rather than an API-breaking change. With strict equality, `NaN` could never match itself, making `value(NaN)` impossible to satisfy and inverting the expected behavior of the negative actions. The change also aligns these APIs with existing `picklist`, which already uses `SameValueZero` semantics via `Array.prototype.includes`. There are no API or type-level changes. I am comfortable releasing it in the next v1 release. 1.5 for example.

> * I recommend `_sameValueZero` follows the `_is` prefix that we use for functions that return a boolean.

Done

library/src/actions/notValue/notValue.ts
yslpn: Yes, thanks

ysknsid25 (dismissed):

fabian-hiller (approved):

PR #1577, opened 2026-08-05
Replace `CLAUDE.md` symlink with `@AGENTS.md` import

### LINKED ISSUE #1576: Revert `CLAUDE.md` symlink for Windows users
https://github.com/open-circle/valibot/commit/3865efe6bb4a8591c77fe40405db325cf64f5d63 - this is an old one, but symlink does not work reliably in Windows which might hurt the DX with LLMs.

> For those reasons, Git for Windows disables support for symbolic links by default (it will still read them when it encounters them). You can enable support via the `core.symlinks` config variable, e.g. when cloning:
> 
> ```sh
> git clone -c core.symlinks=true <URL>
> ```
>
> https://gitforwindows.org/symbolic-links.html
 
> On Windows, creating a symlink requires Administrator privileges or Developer Mode, so use the `@AGENTS.md` import instead.
> 
> https://code.claude.com/docs/en/memory#agents-md

Workaround is to create `./claude/CLAUDE.md` with the following content:

```md
See @../AGENTS.md
```

lazerg:
Git for Windows does not create symlinks unless the clone sets `core.symlinks=true` and the user has Administrator rights or Developer Mode enabled. Without that, `CLAUDE.md` lands as a regular file whose entire content is the string `AGENTS.md`, so Claude Code loads that as the project instructions instead of the real ones.

The Claude Code memory docs recommend the `@AGENTS.md` import for this case, so this replaces the symlink with a one-line file. Nothing changes on macOS and Linux, where both forms resolve to the same content.

Fixes #1576

## Summary by CodeRabbit

* **Documentation**
  * Updated project guidance to remove an outdated reference.
  * No changes were made to application functionality or the end-user experience.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,1 +1,1 @@
-AGENTS.md
+@AGENTS.md


lazerg:
Hey @fabian-hiller, could you take a look at replacing the CLAUDE.md symlink with an @AGENTS.md import? Only bot reviews so far.

PR #1589, opened 2026-08-11
ci: harden GitHub Actions with zizmor

yslpn:
# GitHub Actions security hardening

1. **Switch from Pinact to [zizmor](https://github.com/zizmorcore/zizmor).** Pinact primarily checks and updates SHA references for actions, while zizmor also analyzes injection risks, permissions, credentials, and other common GitHub Actions vulnerabilities. Zizmor is therefore better suited as a security gate in CI, so Pinact was removed from CI.

2. **`persist-credentials: false`.** Zizmor's `artipacked` audit checks checkout steps for this setting and reports a finding when it is missing. We set it explicitly so the GitHub token is not stored in the local Git configuration after checkout, reducing the risk of accidental leakage through subsequent commands or artifacts.

3. **Remove the npm installation step.** Installing `npm@11.11.1` with `npm install --global` bypassed the lockfile and triggered zizmor's `adhoc-packages` finding. The npm version bundled with Node.js 24 is now used instead ([v11.17.0](https://nodejs.org/en/download/archive/v24))

4. **Pin GitHub Actions to commit SHAs.** All external actions are pinned to specific commit hashes instead of mutable tags. This protects CI from tag replacement and reduces supply-chain attack risks.

Reference: [The secure way to release an npm package](https://evilmartians.com/chronicles/the-secure-way-to-release-an-npm-package)

Same for Formisch https://github.com/open-circle/valibot/pull/1589

## Summary by CodeRabbit

* **Security**
  * Improved automation security by disabling credential persistence during repository checkouts.
  * Pinned security-related automation tools to specific versions for more predictable execution.
  * Added automated workflow security analysis for code changes and updates to the main branch.

* **Chores**
  * Simplified the development environment setup by removing an unnecessary global package installation.
  * Removed an obsolete workflow that validated action version pinning.


.github/workflows/check-workflows.yml
```diff
@@ -0,0 +1,24 @@
+name: Lint CI workflows
+
+on:
+  push:
+    branches: [main]
```
ysknsid25: It doesn't need to run all the time; I think it only needs to execute when there are changes to `./github`.

.github/workflows/check-workflows.yml
```diff
@@ -0,0 +1,24 @@
...
+  pull_request:
+    branches: ['**']
```
ysknsid25: Same as above

.github/workflows/check-workflows.yml
yslpn: Got it, thanks

.github/workflows/check-workflows.yml
yslpn: Fixed

ysknsid25 (dismissed):
LGTM👌
I thought I'd like to use zizmor in my other works.

.github/workflows/check-workflows.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@6fc4b006235f201fdab3722e17240ab420d580e5 # v0.6.1
```
ysknsid25: latest version v0.6.2 published 2weeks ago.
FYI: https://github.com/zizmorcore/zizmor-action/releases/tag/v0.6.2

ysknsid25 (commented):
Sorry, I just noticed a nitpicky detail; could you check it?

.github/workflows/check-workflows.yml
```diff
@@ -0,0 +1,28 @@
+name: Lint CI workflows
+
+on:
+  push:
+    branches: [main]
+    paths:
```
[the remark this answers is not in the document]
yslpn: @ysknsid25  What do you think about this? I think we should make this workflow mandatory in the future, but then that would block PRs without changes to .github/**

Besides, the official snippet recommends

```yaml
on:
  push:
    branches: ["main"]
  pull_request:
    branches: ["**"]
```

https://docs.zizmor.sh/integrations/#via-zizmorcorezizmor-action

.github/workflows/check-workflows.yml
yslpn: updated

.github/workflows/check-workflows.yml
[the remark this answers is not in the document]
ysknsid25: I see. If it's a limitation imposed by `zizmor`, then it can't be helped; I think it's fine to remove `./github`.

My initial reason for wanting to include it was simply to avoid having it run pointlessly on the vast majority of PRs—since only a small fraction actually modify GitHub Actions workflows—but I don't have any strong objections against it 🙏

ysknsid25 (dismissed):

ysknsid25 (approved):

fabian-hiller (approved):

PR #1592, opened 2026-08-13
docs: add shorn to ecosystem utilities

ChiChuRita:
Adds shorn to the Utilities category, alphabetically between `fastify-type-provider-valibot` and `valibot-env`, and adds myself to the page contributors.

shorn uses a Valibot schema as a binary wire format. An object with a string and a non-negative integer field is 35 bytes of minified JSON and 8 bytes through shorn. Field names and type tags stay off the wire because the schema already carries them, so there is no IDL and nothing to generate.

For validation it reads Standard Schema, and for structure it reads Standard JSON Schema, so Valibot works by passing `toStandardJsonSchema(schema)` from `@valibot/to-json-schema` alongside the schema. There is no Valibot-specific code path in the library. Golden test vectors pin the exact bytes across Valibot, Zod and ArkType, so equivalent schemas in any of the three encode identically.

Docs: https://shorn.dev
Repo: https://github.com/ChiChuRita/shorn
npm: `@chichurita/shorn`

Worth flagging: shorn is at 0.1.0 and describes itself as experimental. The wire format is not frozen until 1.0. If the page is meant for stable libraries only, I am happy to close this and open it again later.

## Summary by CodeRabbit

* **Documentation**
  * Added the `shorn` utility to the Ecosystem list.
  * Updated the page contributor information.

### DIFF
--- a/website/src/routes/guides/(get-started)/ecosystem/index.mdx
+++ b/website/src/routes/guides/(get-started)/ecosystem/index.mdx
@@ -31,6 +31,7 @@ contributors:
   - haihv
   - nykula
   - yamcodes
+  - ChiChuRita
 ---
 
 # Ecosystem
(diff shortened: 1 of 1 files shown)


yslpn:
Great idea for a library, I see no problem adding it.

@fabian-hiller Would you mind adding a link to the new experimental library?

fabian-hiller:
What do you mean by
> adding a link to the new experimental library

PR #1595, opened 2026-08-14
Fix invalid JSON Schema for value actions on unsupported types

### LINKED ISSUE #1594: to-json-schema: errorMode 'ignore'/'warn' emits invalid JSON Schema for string minValue/maxValue/gtValue/ltValue
### Bug

With `errorMode: 'ignore'` or `'warn'`, converting a `string` (or any non-number) schema that carries `minValue`/`maxValue`/`gtValue`/`ltValue` produces an invalid JSON Schema: the string requirement is written into `minimum`/`maximum`/`exclusiveMinimum`/`exclusiveMaximum`, which must be numbers.

### Repro (published `@valibot/to-json-schema` 1.7.1, `valibot` 1.4.2)

```js
import * as v from 'valibot';

toJsonSchema(v.pipe(v.string(), v.minValue('m')), { errorMode: 'ignore' });
// => { type: 'string', minimum: 'm', $schema: '.../draft-07/schema#' }
```

`v.pipe(v.string(), v.minValue('m'))` is a valid Valibot schema (lexicographic string comparison). Ajv rejects the emitted schema: `schema is invalid: data/minimum must be number`. The same happens with `maxValue` (`maximum`), `gtValue` (`exclusiveMinimum`) and `ltValue` (`exclusiveMaximum`), in both `'ignore'` and `'warn'` modes.

### Expected vs actual

The README says non-throw modes ignore unsupported actions (e.g. `v.pipe(v.string(), v.creditCard())` with `errorMode: 'ignore'` → `{ type: 'string' }`, which I confirmed). So the expected output here is `{ type: 'string' }`. The actual output injects a non-numeric `minimum`, which violates the JSON Schema meta-schema.

### Cause

In `packages/to-json-schema/src/converters/convertAction/convertAction.ts`, the `min_value`/`max_value`/`gt_value`/`lt_value` cases record the "not supported on type" error but then still assign the numeric keyword unconditionally (e.g. `jsonSchema.minimum = valibotAction.requirement as number`). That assignment should be skipped when the type isn't `number`/`integer`, like the existing OpenAPI and `includes` branches already do.

lazerg:
With `errorMode` set to `ignore` or `warn`, `minValue`, `maxValue`, `gtValue` and `ltValue` recorded the "not supported on type" error but still wrote their keyword, so a string schema ended up with something like `minimum: 'm'`, which no JSON Schema validator accepts. The four cases now stop after the error like the OpenAPI 3.0 branch already does, so the unsupported action is skipped and the rest of the schema stays valid.

Fixes #1594

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of incompatible value constraints during schema generation.
  * Invalid greater-than, less-than, minimum, and maximum constraints are now reported in warning mode without adding unsupported numeric restrictions.
  * Existing schema types are preserved when constraints cannot be applied.
  * Ignore mode returns unchanged schemas without warnings for invalid constraints.
  * Prevented incompatible numeric constraints from altering generated schemas or introducing misleading validation behavior.


lazerg:
Added ignore mode tests for gtValue, ltValue, maxValue, and minValue in ec4fc805. Each asserts the schema stays unchanged and no error is thrown.

yslpn:
Hi. Thanks for the contribution! LGTM.

yslpn (dismissed):

PR #1604, opened 2026-08-25
fix(to-json-schema): derive lazy reference IDs from the conversion context

dylanpulver:
Generated reference IDs for `lazy` schemas currently depend on previous conversions in the process and can overwrite user-provided definitions. For example, a generated string definition can reuse `'0'` even when that name already belongs to a number schema, changing what existing references validate.

This change gives each conversion a `ReferenceMap` with its own counter and reserved-ID set. Its `set()` method records every ID, including definition aliases and references added or replaced by override hooks. `createId()` skips reserved IDs and existing definition keys without rescanning the map. Repeated conversions produce stable references, and nested conversions have independent state.

Provided names are reserved before conversion without inserting unfinished definition placeholders into the context visible to callbacks. The implementation uses TypeScript private fields and an explicit no-argument constructor for the existing ES2020 build target.

`ConversionContext.referenceMap` is now typed as `ReferenceMap`, which retains the Map API. Callback consumers continue to use the same map operations; externally constructed contexts containing a plain Map need a type adjustment. The changelog records this type change and the reference-ID fix.

Validation:

- 248 package tests pass; `ReferenceMap` has 100% statement, branch, and function coverage.
- Regression tests cover repeated and nested conversions, numeric collisions, definition aliases, reference additions/replacements by override hooks, definitions added without map entries, and callback visibility of converted definitions.
- TypeScript, Deno, ESLint, Prettier, and package build pass.
- Additional built-output checks cover all three schema targets, and all four JavaScript outputs parse as ES2020.
- Local Node 24 benchmark: 3,200 lazy references with override hooks take about 1 ms per conversion, compared with about 38 ms for the previous PR revision that rescanned the map.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed generated reference IDs for lazy and recursive schemas to avoid collisions with existing definitions.
  - Ensured repeated conversions produce consistent JSON Schema output and references.
  - Corrected generated `$defs` and `$ref` values for recursive schemas without user-provided definitions.
  - Improved reference generation and deduplication when schemas are reused.
  - Rejected non-finite numeric literals such as `NaN` and infinite values.
  - Numeric constraints now skip unsupported types in warning and ignore modes.

- **Tests**
  - Added coverage for stable references, collision avoidance, schema deduplication, repeated conversions, and non-finite literal validation.

### DIFF
--- a/packages/to-json-schema/CHANGELOG.md
+++ b/packages/to-json-schema/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to the library will be documented in this file.
 - Add passthrough of other `metadata` action properties to support custom annotations and standard keywords like `format`, which take precedence over generated properties (pull request #1591)
 - Fix `examples` property type of `JsonSchema` to always be an array (pull request #1607)
 - Fix JSON compatibility validation of `literal` schemas to reject `NaN` and infinite numbers by default (pull request #1573)
+- Fix generation of reference IDs for `lazy` schemas to depend on the conversion context instead of a module scoped counter (pull request #1604)
 
 ## v1.7.1 (June 08, 2026)
 
--- a/packages/to-json-schema/CHANGELOG.md
+++ b/packages/to-json-schema/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to the library will be documented in this file.
 - Fix JSON compatibility validation of `literal` schemas to reject `NaN` and infinite numbers by default (pull request #1573)
 - Fix `minValue`, `maxValue`, `gtValue` and `ltValue` actions to skip numeric constraints on unsupported types in `warn` and `ignore` error modes (pull request #1595)
 - Fix generation of reference IDs for `lazy` schemas to produce consistent output and avoid collisions with existing definitions (pull request #1604)
+- Change `ConversionContext.referenceMap` type from `Map` to its `ReferenceMap` subclass (pull request #1604)
 
 ## v1.7.1 (June 08, 2026)
 
--- a/packages/to-json-schema/src/converters/convertSchema/convertSchema.test.ts
+++ b/packages/to-json-schema/src/converters/convertSchema/convertSchema.test.ts
@@ -1449,6 +1449,42 @@ describe('convertSchema', () => {
       ).toStrictEqual({ $ref: '#/$defs/1' });
     });
 
+    test('should skip taken keys for every generated reference ID', () => {
+      const takenSchema = v.number();
+      const wrappedSchemas = [v.string(), v.boolean(), v.null()];
+      const context = createContext({
+        definitions: { '1': { type: 'number' } },
+        referenceMap: new Map().set(takenSchema, '1'),
+      });
+
+      // Hint: Reference IDs are allocated one after the other within a single
+      // context, so every allocation must skip the taken `'1'` key, not just
+      // the first one.
+      for (const [index, wrappedSchema] of wrappedSchemas.entries()) {
+        expect(
+          convertSchema(
+            {},
+            v.lazy(() => wrappedSchema),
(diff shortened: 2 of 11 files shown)


dylanpulver:
Confirmed and fixed in `b3a432f4`.

I measured it before changing anything. Schema: an object with N entries, each a `v.lazy()` returning a distinct schema, so each allocates its own reference ID. Times are per conversion, averaged over repeated runs after warm-up.

| distinct lazy refs | `main` | this branch (before) | this branch (after) |
|---|---|---|---|
| 100 | 0.052 ms | 0.263 ms | 0.044 ms |
| 200 | 0.103 ms | 0.744 ms | 0.089 ms |
| 400 | 0.288 ms | 2.851 ms | 0.272 ms |
| 800 | 0.521 ms | 11.295 ms | 0.382 ms |
| 1600 | 1.317 ms | 44.418 ms | 0.950 ms |

Doubling the reference count roughly quadrupled the time before the fix (~3.9x per doubling) versus roughly doubling on `main`, and at 1600 references the branch was 34x slower than `main`. So this was a regression the PR introduced, not a pre-existing cost, which is why I treated it as in scope.

The fix caches the used-ID scan per conversion context in a `WeakMap` and resumes the search where the previous allocation stopped. That is safe because both entry points (`toJsonSchema`, `toJsonSchemaDefs`) register every provided definition in the reference map before any conversion starts, and each generated ID is marked as used, so allocated IDs strictly increase. `ConversionContext` is unchanged, so nothing new leaks into the `overrideRef` payload.

Added `should skip taken keys for every generated reference ID`, which pins the generated IDs rather than timing: with `'1'` pre-taken, three successive allocations must yield `'0'`, `'2'`, `'3'`. Replacing the search with a plain cached counter — the obvious wrong version of this optimization — fails it:

```
AssertionError: expected { '$ref': '#/$defs/1' } to strictly equal { '$ref': '#/$defs/2' }
```

`to-json-schema` is 230/230 on the branch and 227/227 on `main`, no failure-set change. `eslint`, `tsc --noEmit`, `deno check ./src/index.ts` and `prettier --check` are all clean.

fabian-hiller:
@codex review

fabian-hiller:
I refined the implementation to keep reference allocation fast while preserving collision handling.

- Replaced the external WeakMap state and repeated reference-map scans with a small `ReferenceMap` subclass. Its `set()` method reserves IDs immediately, so override hooks can add or replace references without requiring a full scan before each allocation. In the local Node 24 benchmark with 3,200 lazy references, conversions with hooks dropped from about 38 ms to about 1 ms compared with the previous PR revision.
- Kept the allocator simple: two TypeScript `private` fields (`usedIds` and `count`) and a `createId()` method. TypeScript private fields match the cache implementation and avoid native private-field transformation overhead with our ES2020 build target.
- Changed `ConversionContext.referenceMap` to `ReferenceMap` and updated the test setup accordingly, removing the plain-Map fallback. Normal callback use retains the Map API; externally constructed contexts using a plain Map will need a type adjustment. This type change is documented in the changelog.
- Removed the definition placeholders. Every ID is already reserved by `set()`, including aliases, and placeholders incorrectly exposed unfinished `{}` schemas to override callbacks. The explicit no-argument constructor also prevents entries from reaching `Map` before the tracking fields initialize.
- Expanded regression coverage for repeated and nested conversions, reserved IDs, definition aliases, references added or replaced by overrides, definitions added without map entries, and callbacks seeing only previously converted definitions.

Validation: all 248 package tests pass, with 100% statement, branch, and function coverage for `ReferenceMap`. TypeScript, Deno, ESLint, formatting, and the package build pass. Additional built-output checks cover all three schema targets, and all four JavaScript outputs parse as ES2020.

PR #1608, opened 2026-09-01
perf(url): use URL.canParse for validation

yslpn:
# Optimize URL validation

## Summary

- Use `URL.canParse` for URL validation when the API is available.
- Keep the existing `URL` constructor fallback for runtimes that do not support `URL.canParse`.
- Update the API hint to reflect the parser-based validation.
- Document the URL parser behavior on the website API page.

## Motivation

`URL.canParse` checks whether an input is a valid URL without constructing a `URL` object. This reduces the overhead of the `url` validation action while preserving compatibility through the existing fallback.

## Benchmark

Local Vitest benchmark results. The benchmark is intentionally not included in this PR.

Headline results:

| Scenario      |           Baseline |          Optimized | Improvement |
| ------------- | -----------------: | -----------------: | ----------: |
| Valid input   | 2,311,817.38 ops/s | 5,642,020.36 ops/s |   **2.44×** |
| Invalid input |   286,371.33 ops/s | 4,202,686.83 ops/s |  **14.68×** |

Full output:

| Scenario      | Variant   |           hz |    min |    max |   mean |    p75 |    p99 |   p995 |   p999 |    rme |   samples |
| ------------- | --------- | -----------: | -----: | -----: | -----: | -----: | -----: | -----: | -----: | -----: | --------: |
| Valid input   | Optimized | 5,642,020.36 | 0.0001 | 0.2400 | 0.0002 | 0.0002 | 0.0002 | 0.0003 | 0.0003 | ±0.33% | 2,821,011 |
| Valid input   | Baseline  | 2,311,817.38 | 0.0003 | 0.1271 | 0.0004 | 0.0005 | 0.0006 | 0.0006 | 0.0006 | ±0.20% | 1,155,909 |
| Invalid input | Optimized | 4,202,686.83 | 0.0001 | 0.1798 | 0.0002 | 0.0003 | 0.0003 | 0.0003 | 0.0007 | ±0.27% | 2,101,344 |
| Invalid input | Baseline  |   286,371.33 | 0.0030 | 0.2254 | 0.0035 | 0.0036 | 0.0046 | 0.0047 | 0.0101 | ±0.24% |   143,186 |

## Scope

The code change is limited to `library/src/actions/url/url.ts`, with the matching API documentation update in `website/src/routes/api/(actions)/url/index.mdx`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL validation using the platform’s URL parser when available, with a compatibility fallback.
  * Ensured validation safely handles environments without URL parsing support.
  * Updated validation hints to accurately describe URL parsing behavior.

* **Documentation**
  * Added guidance that validation may accept parser-valid values such as `abc:1234`.
  * Clarified that parser-based validation may not identify every semantically invalid URL.


library/src/actions/url/url.ts
```diff
@@ -96,6 +96,12 @@ export function url(
     async: false,
     expects: null,
     requirement(input) {
+      // `URL.canParse` has been widely available across browsers since December 2023,
+      // but the library targets ES2020. Feature-detect it to avoid constructing a URL object.
+      // https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static
+      if (typeof URL !== 'undefined' && typeof URL.canParse === 'function') {
+        return URL.canParse(input);
+      }
```
ysknsid25, line 104: How about adding test code like following:

```typescript
  describe('should detect URL.canParse support', () => {
    afterEach(() => {
      vi.unstubAllGlobals();
      vi.restoreAllMocks();
    });

    test('for environment with URL.canParse', () => {
      const canParseSpy = vi.spyOn(URL, 'canParse');
      const action = url();
      expect(action.requirement('https://example.com')).toBe(true);
      expect(action.requirement('example.com')).toBe(false);
      expect(canParseSpy).toHaveBeenCalledTimes(2);
    });

    test('for environment without URL.canParse', () => {
      const constructorSpy = vi.fn();
      class MockURL extends globalThis.URL {
        constructor(input: string) {
          constructorSpy(input);
          super(input);
        }
      }
      Object.defineProperty(MockURL, 'canParse', { value: undefined });
      vi.stubGlobal('URL', MockURL);
      expect(typeof URL.canParse).toBe('undefined');
      const action = url();
      expect(action.requirement('https://example.com')).toBe(true);
      expect(action.requirement('example.com')).toBe(false);
      expect(constructorSpy).toHaveBeenCalledTimes(2);
    });

    test('for environment without URL', () => {
      vi.stubGlobal('URL', undefined);
      expect(typeof URL).toBe('undefined');
      const action = url();
      expect(action.requirement('https://example.com')).toBe(false);
    });
  });
```

In that case, wouldn't you be able to delete the comment ? -> L99-101

ysknsid25 (commented):
I like this change 💓
I left a suggestion to improve this PR.

library/src/actions/url/url.ts
```diff
@@ -96,6 +96,12 @@ export function url(
     async: false,
     expects: null,
     requirement(input) {
+      // `URL.canParse` has been widely available across browsers since December 2023,
+      // but the library targets ES2020. Feature-detect it to avoid constructing a URL object.
+      // https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static
+      if (typeof URL !== 'undefined' && typeof URL.canParse === 'function') {
```
fabian-hiller, line 102: Should we simplify it to `URL && typeof URL.canParse === 'function'`?

yslpn: @fabian-hiller If the global URL is missing, the expression `URL && ...` will throw a `ReferenceError`.

library/src/actions/url/url.ts
yslpn: Thanks. I've added the tests.

ysknsid25: Thank you for your good work 👍
btw, Do we still need comments? Even if we do keep it, wouldn't it be better to leave it in the test code?

yslpn: It is better to leave comments next to the code; this makes the code easier to read.

ysknsid25 (dismissed):

fabian-hiller:
Simplified feature detection to `if (URL.canParse)` inside the existing `try/catch`. Standard runtimes expose a function or leave it undefined, so the shorter check is sufficient. The constructor fallback is preserved, and a missing `URL` global still returns `false` through the catch. The original implementation already threw in that case, at `new URL(input)`.

Expanded the tests to check both parser paths against the same inputs, including custom schemes, Unicode hosts, relative paths, and malformed URLs, and added a changelog entry. All 18 URL tests and type checks pass, along with lint and formatting. An 11-round Node 24.15.0 microbenchmark showed no meaningful slowdown versus the original PR implementation.

PR #1609, opened 2026-09-02
perf: add benchmark suite and CI comparison

yslpn:
# Add benchmark suite and CI comparison

## Summary

- Add a private `@valibot/bench` workspace powered by Vitest.
- Add focused benchmarks for arrays, objects, enums, picklists, unions, URLs, and variants, plus a representative product-validation workload.
- Add local commands for running benchmarks and creating or comparing baselines.
- Add CI checks for benchmark formatting, linting, and representative benchmark comparisons against the PR base.
- Update the pnpm workspace and lockfile.

This PR adds benchmarking infrastructure only; it does not change library behavior.

## CI note

The first CI run for this PR will fail in the benchmark comparison job because the base checkout does not contain the new `bench` workspace yet. After this PR is merged, the benchmark workspace will be present in the base branch and comparisons for subsequent PRs should work normally.

## Acknowledgements

This work builds on the groundwork from [PR #1527](https://github.com/open-circle/valibot/pull/1527). Thanks to [Igor Savin (@kibertoad)](https://github.com/kibertoad) for the original work and for providing the foundation used to create this PR.

## Summary by CodeRabbit

- **New Features**
  - Added benchmark coverage for common validation scenarios, including arrays, objects, enums, picklists, unions, URLs, variants, and product workflows.
  - Added commands to run benchmarks, create local performance baselines, and compare results.

- **Documentation**
  - Added guidance for organizing workloads, naming benchmarks, and comparing local baselines.

- **Chores**
  - Added automated formatting and lint checks for benchmark changes.
  - Benchmark comparisons now run for relevant library and workflow changes.


.github/workflows/ci.yml
```diff
@@ -85,6 +85,63 @@ jobs:
...
+      - name: Checkout pull request
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Install pull request dependencies
+        run: pnpm install
+      - name: Compare pull request benchmarks
+        run: pnpm exec vitest bench --run representative --compare=/tmp/valibot-baseline.json
+        working-directory: bench
```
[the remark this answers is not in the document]
yslpn: Expected and will be fixed after merge

.github/workflows/ci.yml
```diff
@@ -85,6 +85,63 @@ jobs:
...
+      - name: Checkout pull request base
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+          ref: ${{ github.event.pull_request.base.sha }}
+      - name: Setup environment
+        uses: ./.github/actions/environment
+      - name: Benchmark pull request base
+        run: pnpm exec vitest bench --run representative --outputJson=/tmp/valibot-baseline.json
+        working-directory: bench
```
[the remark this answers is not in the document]
yslpn: Expected and will be fixed after merge

.github/workflows/ci.yml
```diff
@@ -85,6 +85,63 @@ jobs:
         run: pnpm test
         working-directory: library
 
+  bench:
+    name: Run benchmarks
+    needs: install
+    if: github.event_name == 'pull_request'
```
ysknsid25, line 91: Why don't we separate from this to other GitHub Actions Workflow?
In this state, this ci must be run every pull request, but We want to run only target of Valibot. if target is `website`, do we need not to run bench ?

bench/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@valibot/bench",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "bench": "vitest bench --run",
```
ysknsid25, line 6: Valibot originally uses https://github.com/open-circle/schema-benchmarks for benchmarking, so wouldn't using Vitest's `bench` feature skew all the reference points?

The point is that the figures shown in the PR do not strictly match those published in the blog or documentation.

Also, Vitest benchmark is still experimental at the moment, so in that sense, it might be better to avoid it.

So before adding benchmark measurements to the CI, it might be better to first modify https://github.com/open-circle/schema-benchmarks so that it can be used within the CI environment 👀

.github/workflows/ci.yml
```diff
@@ -85,6 +85,63 @@ jobs:
         run: pnpm test
         working-directory: library
 
+  bench:
```
ysknsid25, line 88: Shouldn't auto-generated content be displayed in PR comments or descriptions?
Actions logs are not permanent. They disappear after the Actions log retention period (default 90 days) passes, making it impossible to track comparisons like "how does this compare to three months ago."

.github/workflows/ci.yml
```diff
@@ -85,6 +85,63 @@ jobs:
...
+    steps:
+      - name: Checkout pull request base
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+          ref: ${{ github.event.pull_request.base.sha }}
+      - name: Setup environment
+        uses: ./.github/actions/environment
+      - name: Benchmark pull request base
+        run: pnpm exec vitest bench --run representative --outputJson=/tmp/valibot-baseline.json
```
ysknsid25, line 102: If the purpose of this PR is to measure benchmarks and detect performance degradation, shouldn't the CI fail if performance actually degrades?
(Right now, it doesn't fail unless each workflow throws an exception, does it?)

bench/src/benchmarks/focused/object.bench.ts
ysknsid25, line 1: (Sorry, I don't have any concrete ideas myself, but...)

I wonder if this is the right way to set up the benchmark?

From the perspective of the PR's purpose, we need to be able to detect cases where a change introduces an implementation that could degrade performance. However, if—for instance—an `ISBN` action were added, wouldn't we fail to notice any resulting performance degradation or improvement unless we also updated this benchmark code to account for it?

ysknsid25 (commented):
I like this idea and challenge but I'm care some point.
I don't have answear however I'm writing things come up with just now.

.github/workflows/ci.yml
yslpn: It will run only on changes to the library.

bench/package.json
yslpn: > so wouldn't using Vitest's `bench` feature skew all the reference points?

What do you mean?

> 
> The point is that the figures shown in the PR do not strictly match those published in the blog or documentation.

Why should they match?

> Also, Vitest benchmark is still experimental at the moment, so in that sense, it might be better to avoid it.

No problem, this doesn't affect the runtime or the published library. There are no risks. Furthermore, in version 5, the Vitest API will be fully stabilized, with minor breaking changes described in the migration guide. https://vitest.dev/blog/vitest-5.html

> So before adding benchmark measurements to the CI, it might be better to first modify https://github.com/open-circle/schema-benchmarks so that it can be used within the CI environment 👀

We discussed this with @EskiMojo14 and @fabian-hiller on Discord: `schema-benchmarks` can only measure published package versions. Furthermore, its goal is to benchmark various libraries that have different capabilities and are not interchangeable. Developers need a way to easily track function performance during development, even if other libraries lack similar validation features.

A key objective is also to make it easier to evaluate performance locally when testing new implementations and alternative approaches.

.github/workflows/ci.yml
yslpn: This is the initial version; we first need to see and test how it works in the PR, and so on, before confusing users with any unnecessary comments.

.github/workflows/ci.yml
yslpn: The current goal is to provide developers with a convenient, pre-configured tool for local development.

Additionally, during code review, it is possible to view a report if there are concerns about a potential drop in performance.

I wouldn't want to implement a gate right from the start that could block users. I don't want to configure the CI to fail at this stage without proper testing; we can add that in the future.

bench/src/benchmarks/focused/object.bench.ts
yslpn: For now, we have a basic reference case, and that suffices. New methods can be added later—for instance, during optimization—but they shouldn't be automatically included in CI checks. A reviewer or developer can run a specific benchmark locally or share the results.

In principle, we could run all the benchmarks, but that should be done infrequently—such as when releasing a new version.

We can consider this later; it’s not a blocker.

bench/package.json
ysknsid25: > so wouldn't using Vitest's bench feature skew all the reference points?
> The point is that the figures shown in the PR do not strictly match those published in the blog or documentation.

if we measure Valibot benchmarks using a different method than before, the benchmark figures won't strictly match those presented in the blog post below, and I am concerned this might give users the impression that we are benchmarking in an ambiguous manner.

https://valibot.dev/blog/why-migrate-to-valibot/#:~:text=This%20is%20one%20of%20Valibot%27s%20biggest%20strengths.%20On%20Schema%20Benchmarks%2C%20Valibot%20currently%20needs%20only%201.91%20kB%20gzipped%20whereas%20other%20schema%20libraries%20such%20as%20Zod%20v4%20require%2016.57%20kB.%20On%20the%20corresponding%20initialization%20benchmark%2C%20the%20same%20schema%20is%20initialized%20in%2054%20%CE%BCs%20and%20therefore%2016x%20faster%20than%20Zod%20v4

> A key objective is also to make it easier to evaluate performance locally when testing new implementations and alternative approaches.

I see. So, if we were to add a method for local verification, is there any reason why adding a way to verify using `schema-benchmarks` wouldn't work? (Instead of using Vitest bench.)

.github/workflows/ci.yml
ysknsid25: I have confirmed that `bench.yml` has been created 👌

.github/workflows/ci.yml
ysknsid25: got it 👍

In that case, how about setting up `bench` and `bench_eslint` as separate CI jobs targeting `/bench`? There are two reasons for this:

1. It makes it clear where to make changes if we ever decide to discontinue this CI (you would simply delete the files).
2. Since `/bench` is unlikely to undergo frequent changes in future PRs, there is no need to run these jobs unnecessarily (especially given that linting and formatting there have absolutely no impact on production code).

bench/package.json
yslpn: > Could we add a local verification mode to `schema-benchmarks` instead of using Vitest's `bench`?

Technically, yes, but it would require changing `schema-benchmarks` in two important ways.

First, `schema-benchmarks` is currently designed around a shared cross-library suite and published package versions. Local verification would require a separate mode for loading and benchmarking the unpublished source from the current PR.

Second, the shared suite intentionally focuses on scenarios that are comparable across libraries. Our internal suite also needs to cover Valibot-specific validations, transformations, and implementation alternatives that other libraries may not support. Adding those cases to the shared suite would either make them unavailable for other libraries or move `schema-benchmarks` away from its original purpose.

So I am not saying that a local `schema-benchmarks` mode is impossible. I am saying that it would be a separate feature with a different purpose, and it would not replace the need for an internal Valibot regression suite.

`schema-benchmarks` should remain the source for public cross-library comparisons, while this private suite is for testing unpublished Valibot code and tracking regressions during development. The goal of this PR is to provide that internal feedback loop without changing the public benchmark methodology.

.github/workflows/ci.yml
yslpn: Move it

bench/src/benchmarks/focused/object.bench.ts
ysknsid25: I see. This is just out of pure curiosity, but do you have a concrete exit strategy?
For instance, what are the conditions for scrapping this change, and what are the conditions for introducing the next one?
My main concern is simply that I’d hate for it to just linger on in a half-finished state.

PR #1611, opened 2026-09-03
fix(to-json-schema): keep the stricter bound when actions repeat

rajanpanth:
### What

Repeated bound actions overwrite the corresponding JSON Schema keyword instead of keeping the stricter bound, so the produced schema accepts values the valibot schema rejects, with no error raised.

Executed differential (valibot `safeParse` vs ajv on the converted schema):

```
v.pipe(v.number(), v.minValue(5), v.minValue(3))  ->  {minimum: 3}
input 4: valibot=false, ajv=true

v.pipe(v.string(), v.minLength(3), v.nonEmpty())  ->  {"type":"string","minLength":1}
input "ab": valibot=false, ajv=true
```

### How

The `safe_integer` case in `convertAction.ts` already implements keep-the-stricter-bound when a keyword exists; this applies the same semantics to `min_value`, `max_value`, `gt_value`, `lt_value`, `min_length`, `max_length`, `length`, `non_empty`, `entries`, `min_entries` and `max_entries` via two small `mergeLowerBound`/`mergeUpperBound` helpers (and refactors `safe_integer` onto them).

### Tests

7 unit tests plus 2 end-to-end tests. With only the source reverted, 9 fail with e.g. `expected { type: 'string', minLength: 1 } to strictly equal { type: 'string', minLength: 3 }`. With the fix, all 237 package tests pass. `prettier`, `eslint` and `tsc` clean on changed files.

Related: #1606 / #1595 (issue #1594) address numeric keywords landing on non-number types in the same switch; this change is orthogonal (their diffs add breaks, this one merges assignments) and applies cleanly alongside.

## Summary by CodeRabbit

- **Bug Fixes**
  - Generated JSON Schemas now preserve the strictest applicable validation limits when multiple rules overlap.
  - Numeric, string-length, array-length, and object-entry constraints merge more consistently, including exclusive bounds.
  - Invalid or non-finite limits are handled safely, preserving valid finite values when available.
  - Invalid numeric constraints no longer alter schemas in warning or ignore modes.
  - Combined integer and numeric constraints retain appropriate safe-integer limits.
  - Action conversion now exposes examples as an array.
  - Literal validation and lazy-schema reference IDs are now handled more reliably.


rajanpanth:
Rebased onto current main (the conflict was my branch still carrying the pre-merge commit of #1607) and addressed both reviews.

**CodeRabbit, non-finite existing bounds:** confirmed and fixed in 24f84a43. `typeof current === 'number'` accepted `NaN`, and since both `NaN > 3` and `3 > NaN` are false the bound was retained, so `v.pipe(v.number(), v.metadata({ minimum: NaN }), v.minValue(3))` produced `{"type":"number","minimum":null}` after serialization, silently dropping the `minValue(3)`. Both helpers now gate on `Number.isFinite`. Regression test added; it fails with the old `typeof` check and passes now, suite at 238.

**Copilot, numeric keywords written after a type error in warn/ignore mode:** this is real, but it predates this PR and is not something I changed. On current main the same cases already do:

```ts
case 'min_value': {
  if (jsonSchema.type !== 'number' && jsonSchema.type !== 'integer') {
    errors = addError(errors, `The "min_value" action is not supported on type "${jsonSchema.type}".`);
  }
  jsonSchema.minimum = valibotAction.requirement as number;   // unconditional
```

My diff swaps that assignment for `mergeLowerBound` and leaves the control flow untouched, so the behavior in warn and ignore mode is unchanged either way. That gap is tracked as #1594 with two open PRs (#1606 and #1595) proposing the guard, so fixing it here would collide with whichever of those lands. Happy to add the guard in this PR instead if you would rather have it here, just say the word and I will rebase onto whichever approach you prefer.

rajanpanth:
Thanks. P2 is already addressed in 24f84a43 as you noted.

On P1, adding a `break` after the type error in `gt_value`, `lt_value`, `min_value` and `max_value`: I am deliberately leaving that out of this PR, for the same reason I gave Copilot above.

It is pre-existing rather than introduced here. On current `main` those cases already write the keyword unconditionally:

```ts
case 'min_value': {
  if (jsonSchema.type !== 'number' && jsonSchema.type !== 'integer') {
    errors = addError(errors, `The "min_value" action is not supported on type "${jsonSchema.type}".`);
  }
  jsonSchema.minimum = valibotAction.requirement as number;   // unconditional
```

This PR swaps that assignment for `mergeLowerBound` and leaves the control flow untouched, so warn and ignore mode behave exactly as before either way.

More importantly, the same "record an error then continue" shape is pinned by tests elsewhere in the codebase. In `convertSchema`, `convertSchema({}, v.literal(123n), { errorMode: 'warn' }, ...)` is asserted to return `{ const: 123n }`, so the unsupported value is expected to still be written. Changing that contract in `convertAction` alone would make the two files inconsistent.

It is also already tracked as #1594 with two open PRs, #1606 and #1595, so a fix here would collide with whichever lands.

If maintainers would rather have it in this PR, say so and I will add it and update the affected tests; I just do not think this PR is the right place to decide the error-mode contract.

fabian-hiller:
I made a few follow-up changes while keeping the original bound-merging behavior:

- Simplified the helpers to `getLowerBound(current, value)` and `getUpperBound(current, value)`. They return the chosen bound, and each call site assigns it to the JSON Schema property. This removes the keyword unions and dynamic property access, makes the mutation explicit, and follows the nearby `getDefinitionRef` naming pattern.
- Used a type guard before comparing bounds, removing the helpers' type assertions while preserving replacement of non-finite existing bounds.
- Expanded the tests to cover both action orders, equal and zero bounds, exclusive bounds, contradictory exact string/array lengths and object entry counts, `safeInteger` interactions, and existing `NaN`/positive-infinity/negative-infinity bounds.
- Added the changelog entry.

Validation: 244 tests pass on this branch. I also combined the changes with current main in a temporary snapshot: all 264 tests, TypeScript, Deno, and the build pass. ESLint and Prettier pass in the active checkout. Across 705 bound combinations, all 4,201 comparisons between Valibot validation and Ajv validation of the generated schemas agree. All eight end-to-end bound-regression tests fail when the merging implementation is removed.

The simpler helpers also performed well in the local Node 24 benchmark: numeric examples were essentially unchanged, while overlapping string bounds and a 200-property schema were about 11% faster than the previous helper implementation.

The outstanding unsupported-type review comments are covered by the already-merged #1595. I verified that its four early exits and regression tests remain intact when this PR is combined with main. Main has not been merged into this branch; these changes are incremental commits on top of the contributor's work.

fabian-hiller:
@codex review

PR #1612, opened 2026-09-03
fix(to-json-schema): report non-finite literal, picklist and enum values as not JSON compatible

rajanpanth:
### What

`toJsonSchema(v.literal(Infinity))` raises no error and yields `const: Infinity`, which JSON-serializes to `{"const":null}`. The emitted schema then accepts `null` while the valibot schema rejects it. Executed differential:

```
serialized schema on input null:  ajv=true   valibot=false
v.picklist([1, Infinity])  ->  {"enum":[1,null]}
v.enum({A: NaN})           ->  {"enum":[null]}
```

The `value` action already rejects these correctly (`The requirement of the "value" action is not JSON compatible.`) because `isJsonConstValue` checks `Number.isFinite`; the schema-side literal, picklist and enum cases only check `typeof`. All three are reachable with clean TypeScript types (a plain `number`), and the README marks literal and picklist as "Only JSON compatible values are supported".

### How

The `literal` case now goes through `isJsonConstValue`, the `picklist` and `enum` predicates gain `Number.isFinite`, and `enum` raises the same not-JSON-compatible error as `picklist`. The README enum row gets the same note.

### Tests

Three added, one per schema kind. With only the source reverted they fail with `expected [Function] to throw an error`; restored, all 231 package tests pass. `prettier`, `eslint` and `tsc` clean on changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * JSON Schema conversion now rejects non-finite values such as `NaN`, `Infinity`, and `-Infinity` in literal, enum, and picklist schemas.
  * Enum and picklist conversion now accepts only strings and finite numbers, rejecting unsupported values such as booleans and bigints.
  * In warning and ignore modes, unsupported values are skipped rather than included in generated schemas.

* **Documentation**
  * Clarified that enum conversion supports only JSON-compatible values.


rajanpanth:
Thanks, two findings and I took one of them.

**P3, duplicated option predicate: fixed in 95419e14.** You were right that it reimplemented logic that already exists. I used the `isJsonEnumValues` helper, which is `values.every(isJsonConstValue)` and is already what `convertAction` uses for the `values` and `notValues` requirements, so the `enum` and `picklist` cases are now consistent with the `literal` and `value` paths.

One behavior consequence worth flagging rather than burying: `isJsonConstValue` accepts booleans, while the previous `picklist` predicate did not. So `v.picklist([true, false])` now converts to `enum: [true, false]` instead of raising "not JSON compatible". Booleans are valid JSON and valid in an `enum`, so I believe that is a fix rather than a regression, and no existing test pinned the old behavior. Happy to keep the boolean rejection if you would rather that change be made separately.

**P1, exiting the case in warn and ignore mode: not taken, because the current behavior is deliberate and pinned by tests.** I tried it, and it fails the existing suite:

```
convertSchema({}, v.literal(123n), { errorMode: 'warn' }, createContext())
  expected to equal { const: 123n }
```

That test asserts that in `warn` mode the unsupported value **is** still written to the schema, and the `picklist` equivalent asserts the same for `enum: [123n, 456n]`. So "leave the schema unchanged" would be a deliberate behavior change to the error-mode contract across the whole file, not a fix to this PR. It is also the same systemic pattern as `convertAction`, tracked in #1594 with two open PRs.

I did confirm your underlying point is real: in `warn` and `ignore` mode `toJsonSchema(v.literal(Infinity))` yields `{"const": null}` after serialization. If maintainers want the error modes to stop emitting invalid values, I am happy to do it, but it deserves its own PR and a decision about updating those tests, rather than being folded in here.

Suite is 231 passing, prettier and tsc clean.

rajanpanth:
P2 is right, and it also proves my earlier boolean reasoning wrong, so I am correcting both. Fixed in 08b66921.

I had argued that accepting booleans was a fix because booleans are valid JSON. Two things I checked afterwards show that was the wrong call for these schema kinds:

1. Valibot itself does not allow boolean options here. `PicklistOptions` is `MaybeReadonly<(string | number | bigint)[]>`, and a TypeScript enum can only hold string and number members. A boolean option can only arrive through a cast.
2. As you point out, the type inference below only ever emits `string` and `number`, so a boolean option produced `enum: [true, 'a']` with `type: ['string', 'number']`, a schema that excludes its own enum value.

Rather than adding a boolean type branch for a case valibot's types forbid, the option check is now a file-local `isValidEnumOption` predicate accepting exactly strings and finite numbers. That keeps the single shared check from the duplication cleanup (the two cases still share one predicate) while restoring the boolean rejection, and a `@ts-expect-error` test now pins it so it cannot silently flip again: with the predicate widened to accept booleans, the new test fails with `expected [Function] to throw an error`.

Suite is 232 passing, prettier and tsc clean.

fabian-hiller:
I pushed a follow-up in d6d4243bf to simplify the implementation and make the error modes consistent.

- Combined the `enum` and `picklist` branches and inlined their shared string/finite-number check. Both now need the same conversion logic, so a separate helper and duplicate branches are unnecessary. `isJsonConstValue` remains appropriate for literals, where booleans are also supported.
- Stop converting an unsupported `literal`, `enum`, or `picklist` after recording the error. Previously, `warn` and `ignore` still emitted unsupported values: for example, an enum containing `Infinity` serialized that value as `null`, while bigint values caused JSON serialization to throw. These modes now omit the entire unsupported restriction, preserving existing metadata and allowing `overrideSchema` to handle the error. They do not filter individual invalid options out of the list.
- Kept the existing default rejection of bigint literals and picklists. Valibot's bigint support is unchanged; this concerns JSON Schema conversion. The new default rejection of non-finite enum/picklist options remains the main fix in this PR. The corresponding literal validation is already on main from #1573, so I removed the duplicate literal test.
- Expanded coverage across all three targets and error modes, including negative infinity, mixed valid/invalid options, finite numeric boundaries, metadata preservation, overrides, nested JSON serialization, and a later `values` action. Updated the changelog to describe both behavior changes.

Validation: all 315 package tests pass, along with TypeScript, ESLint, and formatting checks. Earlier Deno, build, and differential checks also passed; the final inline refactor was followed by another test and type-check run. Benchmarks against the original PR implementation showed comparable performance.

fabian-hiller:
@codex review

PR #1613, opened 2026-09-03
fix(to-json-schema): merge repeated value, values and notValue restrictions

rajanpanth:
### What

Repeated `const`, `enum` and `not` restrictions silently replace earlier ones in `convertAction`, so the produced schema accepts values the valibot schema rejects. Executed differential (valibot `safeParse` vs ajv on the converted schema):

```
v.pipe(v.number(), v.value(5), v.value(6))        ->  {const: 6}
input 6: valibot=false, ajv=true

v.pipe(v.number(), v.values([1,2]), v.values([2,3]))  ->  {enum: [2,3]}
input 3: valibot=false, ajv=true

v.pipe(v.number(), v.notValue(1), v.notValue(2))  ->  {not: {const: 2}}
input 1: valibot=false, ajv=true
```

### How

- `values` intersects with an existing `enum` (the valibot runtime semantics of chaining them).
- Repeated `not` restrictions merge into `not: { anyOf: [...] }` (not A and not B).
- A `value` action that conflicts with a different existing `const` raises the same "not supported in combination" error the converter already uses for repeated regex actions; an identical repeated value stays silent so currently-working schemas keep converting.

### Tests

Five added. With only the source reverted they fail with e.g. `expected { type: 'number', enum: [ 2, 3 ] } to strictly equal { type: 'number', enum: [ 2 ] }`; restored, all 235 package tests pass. Re-running the differential fuzzer after the fix, the three divergences above are gone. `prettier`, `eslint` and `tsc` clean on changed files.

Note: this touches the same switch as #1611 (bound merging); the two apply cleanly to main independently, and whichever lands second is a trivial rebase.

## Summary by CodeRabbit

- **Bug Fixes**
  - JSON Schema generation now correctly intersects multiple allowed-value restrictions and removes duplicates.
  - Multiple excluded-value restrictions are combined without losing existing rules or adding empty exclusions.
  - Conflicting constant-value restrictions are reported according to the configured error mode.
  - OpenAPI 3.0 schemas merge compatible value restrictions more accurately.
  - Disjoint or empty allowed-value restrictions now produce an explicitly impossible schema instead of an invalid empty enum.
  - Existing schema keywords and boolean exclusions are preserved when restrictions are combined.


rajanpanth:
Good catch, and worse than described. Fixed in 3675f8bf.

I checked against ajv rather than assuming it was a style issue, and an empty `enum` is refused outright in **both** strict and non-strict mode:

```
{type:'number', enum: []}   strict=true   THREW: schema is invalid: data/enum must NOT have fewer than 1 item
{type:'number', enum: []}   strict=false  THREW: schema is invalid: data/enum must NOT have fewer than 1 item
{type:'number', not: {}}    strict=true   compiled, validate(1) = false
{type:'number', not: {}}    strict=false  compiled, validate(1) = false
```

So the converted schema could not be compiled at all, not merely warned about.

`mergeEnumRestriction` now drops the `enum` and merges a `not: {}` when the intersection is empty, which reuses the existing `mergeNotRestriction` path so a later `notValue` still combines correctly. Verified the result compiles under ajv and rejects every value, which matches the valibot semantics that nothing satisfies both restrictions.

Regression test added. With only the source change reverted it fails with `expected { type: 'number', enum: [] } to strictly equal { type: 'number', not: {} }`. Suite is at 236 passing, prettier and tsc clean.

fabian-hiller:
I added a follow-up commit to keep the implementation simple and cover a few missing cases:

- Renamed the enum helper to `intersectEnum`, since it keeps only values allowed by both restrictions. It computes the result before updating the schema, uses set membership for efficient lookups, and preserves the existing value order.
- Changed the exclusion helper to `getNotRestriction`, which returns the combined restriction so assignments stay explicit at the call sites. Existing `not` schemas remain intact, including boolean schemas and schemas with additional keywords.
- Handled empty and duplicate lists. `values([])` now produces `not: {}` because no input can satisfy it; `notValues([])` adds no restriction. Duplicate enum values are removed without modifying the original arrays. These cases previously produced invalid enum arrays.
- Expanded tests for OpenAPI combinations, contradictory restrictions, later actions preserving impossible schemas, error modes, and array preservation. Updated the changelog to cover the fixes.

Validation: all 297 package tests pass, along with ESLint, formatting, and TypeScript checks. Differential checks against Ajv also passed for the string-only, number-only, and boolean-only cases tested.

We deliberately deferred mixed-type value matching to keep this change focused. Valibot can match `1` with `'1'`, while the generated restrictions distinguish them; intersecting those mixed-type restrictions can reject inputs that Valibot accepts. Correctly representing that behavior requires a broader conversion change, so it remains a known limitation of this PR.

fabian-hiller:
@codex review

PR #1621, opened 2026-09-06
fix(intersect): skip inherited prototype props in _merge

Willfric:
the `for...in` loop in `_merge` iterates inherited enumerable properties from `value2`'s prototype chain, not just own properties. this causes prototype properties to leak into the merged output.

the same class of bug was already fixed in the object schema (#1523/#1536). adding an `hasOwnProperty` guard on `value2` at the top of the loop fixes it.

before:
```ts
const proto = { inheritedProp: "leaked" }
const b = Object.create(proto)
b.own = "value"
// _merge({ a: 1 }, b) => { a: 1, own: "value", inheritedProp: "leaked" }
```

after:
```ts
// _merge({ a: 1 }, b) => { a: 1, own: "value" }
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved object merging for objects with custom `constructor` properties.
  - Prevented prototype pollution when processing inputs containing special keys such as `__proto__` and `prototype`.
  - Improved handling of inherited properties, shared references, getters, and special-key conflicts during schema intersections.
  - Object and asynchronous object intersections now correctly accept declared `prototype` entries.

### DIFF
--- a/library/CHANGELOG.md
+++ b/library/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to the library will be documented in this file.
 - Fix `intersect` schema to merge matching `NaN` values and invalid dates (pull request #1573)
 - Fix `cache` and `cacheAsync` methods to clone the issues of a cached dataset, preventing parent schemas from adding their path item to the same issue on every cache hit (pull request #1620)
 - Fix `strictObject`, `looseObject`, `objectWithRest` and their async variants to correctly handle unknown input keys that collide with `Object.prototype` members (pull request #1523)
+- Fix `intersect` and `intersectAsync` schemas to ignore inherited properties when merging objects and preserve own properties without invoking inherited setters or changing the output prototype (pull request #1621)
 
 ## v1.4.2 (June 28, 2026)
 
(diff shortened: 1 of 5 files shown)


fabian-hiller:
Updated `_merge` to ignore inherited properties while preserving legitimate own keys instead of silently dropping them. `Object.keys` limits iteration to own enumerable keys; a computed-property spread creates an own `__proto__` data property before assignment, preventing the inherited setter from changing the output prototype. Ordinary keys keep direct assignment, reducing bundle size and avoiding the per-property `Object.defineProperty` overhead. Added regression tests and a changelog entry; all 4,621 library tests, type checks, lint, and formatting pass.

fabian-hiller:
Switched `_merge` to `{ ...value1, ...value2 }`, followed by recursive merging of shared keys. This is smaller and simpler, preserves own `__proto__` data, and avoids inherited setters/non-writable properties without `Object.defineProperty` on every write. For objects with only a few shared keys, benchmarks against the earlier single-spread version were roughly unchanged on Node and 5–10% faster on Bun. The tradeoffs are slower heavily overlapping merges and reading shared getters twice.

Restored the previous constructor-based check because the exact-prototype check broke the existing custom-prototype regression test; constructor handling remains a separate follow-up. Updated regression tests and restored the changelog entry. All 4,684 library tests, type checks, lint, formatting, and build pass locally.

PR #1622, opened 2026-09-07
chore: Upgrade to TypeScript 6

ysknsid25:
FYI: https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/
> TypeScript 6.0 acts as the bridge between TypeScript 5.9 and 7.0

Since TypeScript 6 serves as a bridge version to version 7, we will first upgrade to version 6 and run operations on it. After that, if no issues are found, we will upgrade to version 7. I will also handle the version upgrade process to version 7.

# Actions Required for the Migration to TypeScript 6

## typescript-eslint, tsdown
We needed to upgrade these libraries to versions compatible with TypeScript 6.

## packages/i18n node
In TS 6, the default value for types changed from reading everything under `node_modules/@types` to `[]`.
`packages/i18n` has `@types/node@^24.10.1` in its devDependency, which was automatically loaded up to TS 5.9. In TS 6, it is no longer loaded, leading to the issue where `import fs from 'node:fs'` in `scripts/build-*.ts` could no longer be resolved. We have explicitly specified `["node"]`.

## moduleResolution
Since `node` has been deprecated, we are using `bundler`, the same as formisch.
https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/

## Fixes to pipe.ts
The new rule `unified-signatures` in typescript-eslint 8.68 detected an issue and caused `pnpm lint` to fail. We have disabled it only for this part.

## partial.test-d.ts, partialAsync.test-d.ts, required.test-d.ts, requiredAsync.test-d.ts, record.test-d.ts, recordAsync.test-d.ts
TS 6 changed its internal identity judgment logic, and now distinguishes between `k?: T` and `k?: T | undefined`. We have reflected the updates to adapt to this change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected type-level expectations so optional properties and record values explicitly allow `undefined` across synchronous and asynchronous schema APIs.

- **Chores**
  - Updated TypeScript and related development tooling versions.
  - Modernized compiler configuration with bundler-based module resolution and Node type declarations across applicable packages.


yslpn:
> TS 6 changed its internal identity judgment logic, and now distinguishes between k?: T and k?: T | undefined. We have reflected the updates to adapt to this change.

@ysknsid25 Hi Thanks for working on this!

However, this distinction was introduced in TypeScript 4.4 when the `exactOptionalPropertyTypes` option is enabled. I couldn't find any mention of this in the migration guide: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#breaking-changes-and-deprecations-in-typescript-60

Can you pinpoint the exact cause of the errors?

ysknsid25:
@yslpn 

# Root cause of the `TS2554` errors in `*.test-d.ts` under TypeScript 6

You're right on both counts, and my earlier explanation was imprecise. `exactOptionalPropertyTypes`
has drawn the `{ k?: T }` vs. `{ k?: T | undefined }` distinction since TypeScript 4.4, and there is
nothing in the 6.0 breaking-changes list that covers this. Here is what I was able to establish.

## What the errors actually are

The 26 failures all look like this:

```
src/methods/partial/partial.test-d.ts(46,45): error TS2554: Expected 1 arguments, but got 0.
```

Despite the error code, this is not an arity problem — it is how `expect-type` reports a failed
assertion. From `expect-type@1.3.0`:

```ts
toEqualTypeOf: <Expected extends ...>(
  ...MISMATCH: MismatchArgs<StrictEqualUsingTSInternalIdenticalToOperator<Actual, Expected>, true>
) => true
```

`MismatchArgs<Result, true>` is an empty tuple when `Result` is `true` and a one-element tuple when it
is `false`. So a type mismatch makes the parameter required, and the call site reports
"Expected 1 arguments, but got 0". Every one of the 26 errors is a genuine `toEqualTypeOf` mismatch.

The equality check itself is the well-known trick that surfaces TypeScript's **internal type-identity
relation**:

```ts
type StrictEqualUsingTSInternalIdenticalToOperator<L, R> =
  (<T>() => T extends (L & T) | T ? true : false) extends
  (<T>() => T extends (R & T) | T ? true : false) ? ... : false
```

## Minimal reproduction

No valibot code involved. Same `tsconfig.json` for both runs, only the compiler binary differs:

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "strict": true,
    "exactOptionalPropertyTypes": true,
    "target": "ES2020",
    "lib": ["ESNext"],
    "module": "ESNext",
    "moduleResolution": "bundler",
    "noEmit": true
  },
  "include": ["pin.ts"]
}
```

```ts
// pin.ts
type Identical<L, R> =
  (<T>() => T extends (L & T) | T ? true : false) extends
  (<T>() => T extends (R & T) | T ? true : false) ? true : false;
type Assign<L, R> = [L] extends [R] ? true : false;

type P     = Partial<Record<'foo' | 'bar', string>>;
type WithU = { foo?: string | undefined; bar?: string | undefined };
type NoU   = { foo?: string; bar?: string };

declare const i1: Identical<P, WithU>;
declare const i2: Identical<P, NoU>;
declare const i3: Identical<WithU, NoU>;
declare const a1: Assign<WithU, NoU>;
declare const a2: Assign<NoU, WithU>;

// Assigning to `0` forces the compiler to print each resolved type in the error message.
export const _i1: 0 = i1;
export const _i2: 0 = i2;
export const _i3: 0 = i3;
export const _a1: 0 = a1;
export const _a2: 0 = a2;
```

Results:

| Expression | TS 5.9.3 | TS 6.0.3 |
| --- | --- | --- |
| `Identical<Partial<Record<'foo' \| 'bar', string>>, { k?: string \| undefined }>` | `true` | **`false`** |
| `Identical<Partial<Record<'foo' \| 'bar', string>>, { k?: string }>` | `true` | `true` |
| `Identical<{ k?: string \| undefined }, { k?: string }>` | **`true`** | **`false`** |
| `Assign<{ k?: string \| undefined }, { k?: string }>` | `false` | `false` |
| `Assign<{ k?: string }, { k?: string \| undefined }>` | `true` | `true` |

Three things follow:

1. **`Partial` did not change.** Row 2 is `true` on both versions, so
   `Partial<Record<'foo' | 'bar', string>>` is `{ foo?: string; bar?: string }` — without
   `| undefined` — on 5.9 and on 6.0 alike.
2. **Assignability did not change.** Rows 4 and 5 are identical across versions, in both directions.
   The `exactOptionalPropertyTypes` distinction was already honored by the assignability relation.
3. **Only the identity relation changed.** Row 3 flips: TS 5.9 considered `{ k?: T | undefined }` and
   `{ k?: T }` identical, TS 6 does not.

## Conclusion

TypeScript's **type-identity relation** now accounts for the `exactOptionalPropertyTypes` distinction
between an optional property whose type includes `undefined` and one whose type does not. Until 5.9
there was an inconsistency — assignability drew the distinction, identity did not. In 6.0 the two
agree.

That also explains why it is absent from the migration guide. It is not a language-level change and
is unobservable from ordinary code; it only becomes visible through the
`(<T>() => ...) extends (<T>() => ...)` trick that libraries such as `expect-type` and `type-fest`
use to reach the internal relation. `toEqualTypeOf` is built on exactly that, which is why it is the
only thing in the suite that reacted. Assignability-based assertions such as `toExtend` are unaffected.

I did not bisect the nightlies to identify the specific PR, so the statement above describes the
observable change in behavior rather than a cited commit.

## Why this is not a regression in valibot

The inferred types are byte-for-byte identical on both compilers. Checked against the library source:

| | `InferInput<Schema2>` |
| --- | --- |
| TS 5.9.3 | `{ foo?: string \| undefined; bar?: string \| undefined; }` |
| TS 6.0.3 | `{ foo?: string \| undefined; bar?: string \| undefine

yslpn (dismissed):

PR #1626, opened 2026-09-09
fix(intersect): support own constructor properties

fabian-hiller:
Objects with data such as `{ constructor: 'foo' }` currently fail intersection merging because the own property shadows `Object`. This change falls back to the prototype's constructor when the existing check fails, allowing these fields to be preserved and recursively merged while retaining the normal fast path and existing custom-prototype support.

Follow-up to #1621, originally contributed by @ItzXynx, addressing the remaining [constructor review feedback](https://github.com/open-circle/valibot/pull/1621#discussion_r3962208007). This avoids the compatibility regression from requiring the immediate prototype to equal `Object.prototype`.

Validation: all 4,701 library tests, type checks, lint, formatting, and build pass. Nine regression tests fail before the fix and pass afterward; coverage includes both sync and async intersections, nested/shared constructor data, and custom prototypes. Ordinary parsing benchmarks remain within about 2% of the current implementation on Node 24.15 and Bun 1.3.13. The parsing fixture adds 100 minified bytes / 12 gzip bytes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed synchronous and asynchronous schema intersections to accept valid objects with their own `constructor` property.
  * Improved object merging for objects with shadowed `constructor` properties.
  * Preserved validation errors for unsupported class instances and non-plain objects.

* **Documentation**
  * Added an unreleased changelog entry describing the intersection schema fix.


Willfric:
yeah the fallback to `getPrototypeOf` is the right call, cleaner than what I had. lgtm

PR #1629, opened 2026-09-09
fix(forward): stop forwarding only on undefined, not all falsy values

Willfric:
`!pathValue` stops on `0`, `false`, `''` in the middle of a path, cutting it short. also throws `TypeError` when an intermediate value is `null` since `null[key]` crashes.

changed to `pathValue === undefined || pathValue === null` in both `forward` and `forwardAsync`. also added null-safe read so `null` input yields `undefined` instead of throwing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of nested paths when values are `null` or `undefined`, preventing errors during traversal.
  * Preserved traversal for valid falsy values such as `false`, `0`, and empty strings.
  * Ensured synchronous and asynchronous forwarding produce consistent results when encountering nullish or falsy intermediate values.

* **Tests**
  * Added coverage for null path values and traversable falsy values to help prevent regressions.

### DIFF
--- a/library/src/methods/forward/forward.test.ts
+++ b/library/src/methods/forward/forward.test.ts
@@ -201,6 +201,84 @@ describe('forward', () => {
     >);
   });
 
+
+  test('should stop forwarding if path input is null', () => {
+    const input = { nested: null };
+    type Input = { nested: null };
+    const requirement = () => false;
+    expect(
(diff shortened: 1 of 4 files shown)


Willfric:
@coderabbitai review

Willfric:
@coderabbitai full review

PR #1631, opened 2026-09-09
perf: optimize string count validations

yslpn:
# Optimize String Count Validations

## Summary

Improve the performance of byte, grapheme, word, and code point validation actions while preserving exact validation errors.

## Changes

- `minBytes` and `notBytes`: use `input.length` as a lower bound for UTF-8 byte length and skip `TextEncoder` when the result is already guaranteed.
- `minGraphemes` and `notGraphemes`: stop `_getGraphemeCount` after reaching the required threshold; invalid inputs still receive the exact count.
- `minWords` and `notWords`: apply the same threshold-based early exit to `_getWordCount`.
- `minCodePoints` and `notCodePoints`: count code points only until the validation result is known.
- `maxGraphemes` and `maxCodePoints`: accept inputs immediately when their UTF-16 length cannot exceed the configured maximum.
- Add unit tests for limited Unicode counting.

## Local Benchmark Results

The following results were measured locally against a baseline. The benchmark files and baseline are not included in this PR.

The focused workloads are faster in every case:

| Validation group | Improvement |
| ---------------- | ----------: |
| Bytes            |        2.6× |
| Graphemes        |    4.7–168× |
| Words            |        1.1× |
| Code points      |    1.5–2.1× |

The largest gain comes from the `maxGraphemes` fast path, which avoids `Intl.Segmenter` entirely for inputs below the configured limit.

## Summary by CodeRabbit

- **Performance**
  - Improved string-length validation efficiency by skipping unnecessary byte, code point, grapheme, and word counting when results can be determined from the input length.
  - Added early stopping when counting reaches the configured validation limit.

- **Bug Fixes**
  - Preserved existing validation outcomes while reducing processing for short or over-limit inputs.

- **Tests**
  - Added coverage for limited counting, including emoji, graphemes, words, zero limits, and shorter inputs.

### DIFF
--- a/library/src/utils/_getCodePointCount/_getCodePointCount.ts
+++ b/library/src/utils/_getCodePointCount/_getCodePointCount.ts
@@ -2,13 +2,30 @@
  * Returns the code point count of the input.
  *
  * @param input The input to be measured.
+ * @param limit The optional count limit.
  *
- * @returns The code point count.
+ * @returns The code point count, or the count at which the limit was reached.
  *
  * @internal
  */
 // @__NO_SIDE_EFFECTS__
-export function _getCodePointCount(input: string): number {
+export function _getCodePointCount(input: string, limit?: number): number {
+  if (limit !== undefined) {
+    if (limit <= 0) {
+      return 0;
+    }
+    let count = 0;
+    for (let i = 0; i < input.length; ) {
+      count++;
+      if (count >= limit) {
+        return count;
+      }
+      // codePointAt never returns undefined because i is always in bounds
+      i += input.codePointAt(i)! > 0xffff ? 2 : 1;
+    }
+    return count;
+  }
+
   let count = input.length;
   // A surrogate pair cannot start at the last code unit, so no iteration needs
   // to start there
--- a/library/src/utils/_getGraphemeCount/_getGraphemeCount.ts
+++ b/library/src/utils/_getGraphemeCount/_getGraphemeCount.ts
@@ -4,13 +4,17 @@ let segmenter: Intl.Segmenter;
  * Returns the grapheme count of the input.
  *
  * @param input The input to be measured.
+ * @param limit The optional count limit.
  *
- * @returns The grapheme count.
+ * @returns The grapheme count, or the count at which the limit was reached.
  *
  * @internal
  */
 // @__NO_SIDE_EFFECTS__
-export function _getGraphemeCount(input: string): number {
+export function _getGraphemeCount(input: string, limit?: number): number {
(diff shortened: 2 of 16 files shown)


library/src/utils/_getCodePointCount/_getCodePointCount.ts
```diff
@@ -2,13 +2,30 @@
...
+ * @param limit The optional count limit.
  *
- * @returns The code point count.
+ * @returns The code point count, or the count at which the limit was reached.
  *
  * @internal
  */
 // @__NO_SIDE_EFFECTS__
-export function _getCodePointCount(input: string): number {
+export function _getCodePointCount(input: string, limit?: number): number {
```
fabian-hiller: How about we make `limit` required? This would reduce bundle size. We could maybe remove some of the code of the function.

library/src/utils/_getGraphemeCount/_getGraphemeCount.ts
```diff
@@ -4,13 +4,17 @@ let segmenter: Intl.Segmenter;
...
+ * @param limit The optional count limit.
  *
- * @returns The grapheme count.
+ * @returns The grapheme count, or the count at which the limit was reached.
  *
  * @internal
  */
 // @__NO_SIDE_EFFECTS__
-export function _getGraphemeCount(input: string): number {
+export function _getGraphemeCount(input: string, limit?: number): number {
```
fabian-hiller: Same here. Maybe we can then remove the early if check in all the actions.

library/src/utils/_getWordCount/_getWordCount.ts
```diff
@@ -5,16 +5,21 @@ let store: Map<string, Intl.Segmenter>;
...
+ * @returns The word count, or the count at which the limit was reached.
  *
  * @internal
  */
 // @__NO_SIDE_EFFECTS__
 export function _getWordCount(
   locales: Intl.LocalesArgument,
-  input: string
+  input: string,
+  limit?: number
```
fabian-hiller: Same here

fabian-hiller (commented):
It is ok the break the internal functions that start with `_`.

yslpn:
@fabian-hiller 
> How about we make limit required? This would reduce bundle size. We could maybe remove some of the code of the function.

I changed my mind on this one and kept `limit` optional with an `Infinity` default. I removed as much code as I could (the `limit <= 0` guards are gone and `_getCodePointCount` is a single loop now), and the default keeps `codePoints`, `graphemes` and `words` unchanged, so the diff stays smaller.

Bundle size is the same either way. Happy to switch back to required if you prefer.

PR #1638, opened 2026-09-12
feat(to-json-schema): support code point string constraints

yslpn:
# Add code point string constraints to JSON Schema conversion

## Summary

- Add conversion support for `codePoints`, `minCodePoints`, `maxCodePoints`, and `notCodePoints`.
- Keep the existing string conversion for `length`, `minLength`, and `maxLength` for compatibility, while documenting it as deprecated and warning when `errorMode: 'warn'` is enabled.
- Preserve conversion of these legacy actions for arrays.
- Add test coverage and update the package README.

## Why `length` is not suitable for strings

For strings, Valibot uses JavaScript `value.length`, which counts UTF-16 code units. JSON Schema `minLength` and `maxLength` count Unicode code points. The same constraint can therefore validate differently after conversion.

For example:

~~~ts
const value = '😀';

value.length; // 2 UTF-16 code units
[...value].length; // 1 Unicode code point
~~~

`v.pipe(v.string(), v.length(1))` rejects `😀`, while its converted JSON Schema with `minLength: 1` and `maxLength: 1` accepts it. The `codePoints` actions use the same counting unit as JSON Schema and avoid this mismatch.

Combining marks and ZWJ sequences are another caveat: `é` consists of 2 code points, and `👩‍💻` of 3, even though each renders as one user-perceived character. `codePoints` aligns with JSON Schema, but validating user-perceived characters requires grapheme-cluster segmentation.

## Validation

- `pnpm -C packages/to-json-schema test`
- `pnpm -C packages/to-json-schema lint`
- `pnpm -C packages/to-json-schema format.check`

## Sources

1. [OpenAPI Specification 3.1.1, Schema Object](https://spec.openapis.org/oas/v3.1.1.html#schema-object) — the Schema Object is a superset of JSON Schema Draft 2020-12, and keyword definitions follow JSON Schema unless OpenAPI explicitly adds semantics.
2. [JSON Schema Validation Draft 2020-12, §6.3](https://json-schema.org/draft/2020-12/json-schema-validation#section-6.3) — the length used by minLength and maxLength is the number of characters defined by RFC 8259.
3. [RFC 8259, §1](https://www.rfc-editor.org/rfc/rfc8259#section-1) and [§7](https://www.rfc-editor.org/rfc/rfc8259#section-7) — JSON strings are sequences of Unicode characters, and non-BMP characters can be represented by escaped UTF-16 surrogate pairs.
4. [JSON Schema Test Suite: maxLength](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft2020-12/maxLength.json) — the official suite includes a valid maxLength: 2 case with two emoji represented by two surrogate pairs, demonstrating that JSON Schema counts them as two characters rather than four UTF-16 code units.

## Summary by CodeRabbit

* **New Features**
  * Added JSON Schema conversion support for code-point constraints, including minimum, maximum, exact, and excluded counts.
  * Preserved stricter existing bounds and combined exclusion restrictions during conversion.

* **Bug Fixes**
  * Improved handling of invalid code-point requirements and unsupported schema types across error modes.

* **Documentation**
  * Documented code-point actions and clarified the deprecation of string-based length conversions, while retaining array support.
  * Added warnings for deprecated string length actions when configured to warn.

### DIFF
--- a/packages/to-json-schema/README.md
+++ b/packages/to-json-schema/README.md
@@ -58,6 +58,7 @@ This package is particularly popular for:
 | --------------- | ------ | ------------------------------------------------------ |
 | `base64`        | ✅     |                                                        |
 | `bic`           | ✅     |                                                        |
+| `codePoints`    | ✅     |                                                        |
 | `cuid2`         | ✅     |                                                        |
 | `decimal`       | ✅     |                                                        |
 | `description`   | ✅     |                                                        |
@@ -86,21 +87,24 @@ This package is particularly popular for:
 | `isrc`          | ✅     |                                                        |
 | `jwsCompact`    | ✅     |                                                        |
 | `ksuid`         | ✅     |                                                        |
-| `length`        | ⚠️     | Only in combination with `string` and `array` schema   |
+| `length`        | ⚠️     | Deprecated for strings; use `codePoints`; arrays are supported |
 | `ltValue`       | ⚠️     | Only in combination with `number` and `integer` schema |
 | `mac`           | ✅     |                                                        |
 | `mac48`         | ✅     |                                                        |
 | `mac64`         | ✅     |                                                        |
+| `maxCodePoints` | ✅     |                                                        |
 | `maxEntries`    | ✅     |                                                        |
-| `maxLength`     | ⚠️     | Only in combination with `string` and `array` schema   |
+| `maxLength`     | ⚠️     | Deprecated for strings; use `maxCodePoints`; arrays are supported |
 | `maxValue`      | ⚠️     | Only in combination with `number` schema               |
 | `metadata`      | ✅     | Additional properties are added without validation     |
+| `minCodePoints` | ✅     |                                                        |
 | `minEntries`    | ✅     |                                                        |
-| `minLength`     | ⚠️     | Only in combination with `string` and `array` schema   |
+| `minLength`     | ⚠️     | Deprecated for strings; use `minCodePoints`; arrays are supported |
 | `minValue`      | ⚠️     | Only in combination with `number` schema               |
 | `multipleOf`    | ✅     |                                                        |
 | `nanoid`        | ✅     |                                                        |
 | `nonEmpty`      | ✅     |                                                        |
+| `notCodePoints` | ✅     |                                                        |
 | `notValue`      | ⚠️     | Only JSON compatible values are supported              |
 | `notValues`     | ⚠️     | Only JSON compatible values are supported              |
 | `octal`         | ✅     |                                                        |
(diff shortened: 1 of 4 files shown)


packages/to-json-schema/src/converters/convertAction/convertAction.ts
```diff
@@ -499,6 +551,12 @@ export function convertAction(
             errors,
             `The "${valibotAction.type}" action is not supported on type "${jsonSchema.type}".`
           );
+        } else {
+          if (config?.errorMode === 'warn') {
```
fabian-hiller, line 556: Maybe we should show this whenever `errorMode` is not `"ignore"`. We could also not log anything as it would keep the bundle size smaller and this warning is only useful in edge cases.

packages/to-json-schema/README.md
```diff
@@ -86,21 +87,24 @@ This package is particularly popular for:
 | `isrc`          | ✅     |                                                        |
 | `jwsCompact`    | ✅     |                                                        |
 | `ksuid`         | ✅     |                                                        |
-| `length`        | ⚠️     | Only in combination with `string` and `array` schema   |
+| `length`        | ⚠️     | Deprecated for strings; use `codePoints`; arrays are supported |
```
fabian-hiller: Can you fix the formatting of the table? Prettier can do this for you. Otherwise, I can also do it.

fabian-hiller (commented):
I totally get this PR but I am worried if we remove it in v2 many users will complain. It seems unlikely that users change their schemas from length to code points. 😐 And coding agents very likely also use length. Not sure what is best here. I am just worried we remove the happy path. We could also consider to just document it and/or keep a short log warning.

packages/to-json-schema/README.md
yslpn: Fixed

packages/to-json-schema/src/converters/convertAction/convertAction.ts
yslpn: For now, I kept the warning opt-in via `errorMode: 'warn'`, so existing conversions remain quiet by default while users can still detect the semantic mismatch.

I think we can revisit this later and either show it for every mode except `ignore` or remove the warning entirely if bundle size is more important.

yslpn:
> I totally get this PR but I am worried if we remove it in v2 many users will complain. It seems unlikely that users change their schemas from length to code points. 😐 And coding agents very likely also use length. Not sure what is best here. I am just worried we remove the happy path. We could also consider to just document it and/or keep a short log warning.

I think that the converter should be strict in the long term, because `length` is not semantically equivalent to JSON Schema string length.

At the same time, removing the existing conversion could make migration unnecessarily painful. I think we can defer the decision of whether to remove or retain it in v2 and address it as part of a broader migration plan.

PR #1645, opened 2026-09-15
fix(i18n): add missing async action translations

MaxFreedomPollard:
`@valibot/i18n` never translates issues from `partialCheckAsync`, `rawCheckAsync` or `rawTransformAsync`. A pipeline that uses one of them reports the English default message while the sync twin in the same locale is translated.

```ts
import '@valibot/i18n/de';
import * as v from 'valibot';

const Sync = v.pipe(
  v.object({ password: v.string(), confirm: v.string() }),
  v.partialCheck([['password'], ['confirm']], (i) => i.password === i.confirm)
);

const Async = v.pipeAsync(
  v.objectAsync({ password: v.string(), confirm: v.string() }),
  v.partialCheckAsync([['password'], ['confirm']], async (i) => i.password === i.confirm)
);

const input = { password: 'a', confirm: 'b' };

v.safeParse(Sync, input, { lang: 'de' }).issues?.[0].message;
// 'Ungültige Eingabe: Object erhalten'

(await v.safeParseAsync(Async, input, { lang: 'de' })).issues?.[0].message;
// 'Invalid input: Received Object'
```

This hits anyone doing cross-field validation against a database or an API, which is the main reason to reach for `partialCheckAsync` in the first place. The form ends up with a mix of translated and untranslated messages.

### Cause

`_addIssue` resolves a specific message by the action's own `reference` function, in `library/src/utils/_addIssue/_addIssue.ts`: `getSpecificMessage(context.reference, issue.lang)`. An async action's reference is the async function itself, so `library/src/actions/partialCheck/partialCheckAsync.ts` sets `reference: partialCheckAsync`, not `partialCheck`, and `rawCheck/rawCheckAsync.ts` and `rawTransform/rawTransformAsync.ts` in the same folder layout do the same. The registrations come from `Language['specific']` in `packages/i18n/src/types.ts`, which listed `partialCheck`, `rawCheck` and `rawTransform` with no `Async` siblings, so `setSpecificMessage` was never called for the three async references and the lookup always missed. `check`/`checkAsync` and `checkItems`/`checkItemsAsync` were already listed as separate keys, so this was the remaining gap rather than a design choice.

Comparing the 105 actions under `library/src/actions` that call `_addIssue` against the 102 keys in `Language['specific']` leaves exactly `partialCheckAsync`, `rawCheckAsync` and `rawTransformAsync`, and nothing in the other direction.

### Fix

The three keys are added to `Language['specific']` and to all 34 locale files, each reusing the string already written for its sync twin, exactly as `checkAsync` and `checkItemsAsync` already do. No wording is invented and no existing message changes. `kr.ts` is untouched because it spreads `ko`. The generated `packages/i18n/package.json` and `packages/i18n/jsr.json` export maps gain the matching submodule entries in the position the build scripts emit them, which is the order of `Object.entries(language.specific)`, so each new entry sits immediately after its sync twin in all 34 emitted languages and both files stay byte identical to their generated form. `pnpm-lock.yaml`, the build scripts and the library itself are not touched.

### Tests

`packages/i18n` has no test runner and no job in `.github/workflows/ci.yml` (which covers `library`, `packages/to-json-schema`, `website` and the codemod), so nothing automated would have caught this and the guard here is the `Language` interface: dropping any one of the new keys from a single locale file is a `tsc --noEmit` error naming that file, which is what proves all 34 were updated. I also ran the repro above both ways against the locale data. Before the change the three async messages come back as `Invalid input: Received Object` and the sync one is translated. After it, all four return `Ungültige Eingabe: Object erhalten`.

### Verification

`tsc --noEmit` over `packages/i18n/src` passes (the `scripts` half of that package's lint needs `@types/node`, which the scoped install does not pull, and nothing under `scripts` changed) and the diff is 1231 insertions with zero deletions across 38 files. `pnpm test` in `library`: 530 files, 4688 tests passed, no type errors. `prettier --check` on the changed files passes.

## Summary by CodeRabbit

- **New Features**
  - Added localized validation messages for asynchronous checks and transformations across 34 supported languages.
  - Added package exports for the asynchronous validation helpers, including both modern module and CommonJS usage.
- **Documentation**
  - Updated the unreleased changelog to document the new localized messages.


yslpn (approved):
Yes, it seems we've lost that. Thank you!

fabian-hiller:
Thank you!

PR #1649, opened 2026-09-17
hotfix(ksuid): ksuid allows IDs that should be rejected

ysknsid25:
Similar modification: https://github.com/open-circle/valibot/pull/1498

[Following this](https://github.com/segmentio/ksuid), KSUID is:

- Binary KSUIDs are 20-bytes
- 27 characters, encoded in alphanumeric base62

In other words, there is a discrepancy between the range of possible values ​​and the range that can be represented by 27 characters.

- Value Range: 2^160 = 1461501637330902918203684832716283019655932542976 pattern
- 27 characters combination: 62^27 = 2480707824361381849652170924082266893544595652608 pattern

**Value Range < Characters pattern**, so Among the combinations of 27 characters, there are some that are unacceptable.

Specifically, numbers greater than `2^160 - 1` = `1461501637330902918203684832716283019655932542975` = `aWgEPTl1tmebfsQzFP4bxwgy80V` must not be allowed to pass.

To determine this, simply compare the magnitude of the numbers digit by digit. Essentially, it works the same way as comparing 120 and 121 to see which is larger.

- 1st digit: Both are 1, so they are the same.
- 2nd digit: Both are 2, so they are the same.
- 3rd digit: Comparing 0 and 1 shows that 121 is larger.

In Base62, the order is `0–9` < `A–Z` < `a–z`.

So in `aWgEPTl1tmebfsQzFP4bxwgy80V` case,

- 1st digit: can accept `0-9A-Z`
- 2nd digit: can accept `0-9A-V`
- ...

We just need to prepare a regular expression for 27 digits.

## Summary by CodeRabbit

- **Bug Fixes**
  - KSUID validation now rejects 27-character values exceeding the maximum representable KSUID.
  - Valid KSUIDs, including minimum and maximum boundary values, continue to be accepted correctly.

- **Tests**
  - Expanded coverage for valid KSUIDs, per-digit limits, maximum values, overflowing values, and smallest overflowing cases.
  - Added regression checks to verify invalid KSUIDs remain rejected.


yslpn (dismissed):
LGTM, thanks

fabian-hiller:
Thank you! Did you hand-crafted this regex?

ysknsid25:
No, while I devised the solution and the implementation strategy using regular expressions—as outlined in the PR—I left the actual coding to ClaudeCode. I decided that using AI was the better approach because the task involved somewhat repetitive coding and required handling large numerical values.

yslpn:
@ysknsid25 This isn't a blocker for the current PR. However, for version 2.0, it might be worth considering a switch to a proper validation algorithm. This would eliminate the need to maintain a complex regular expression.

ysknsid25:
@yslpn 
For a method that does not use regular expressions, I am noting here that you can write it like `str <= 'aWgEPTl1tmebfsQzFP4bxwgy80V'`. This is because JavaScript allows string comparisons in the form of "AB" < "aB".

Alternatively, we can verify [this](https://github.com/segmentio/ksuid/blob/d33724947fcfba7949906c2b1821e96a1c8d06e7/base62.go#L102-L175) using division. In other words, there are several methods available.

At first glance, this regular expression may appear complex, but it is actually simple once you understand the logic behind the matching process.

